PR #2, opened 2019-10-25
Remove dist folder and binary file. Ignore dist in gitignore

kinow:
Hi,

Not sure if the `dist/` folder is being intentionally kept with a binary (2.8MB) with the sources?

It is probably in PYPI as well, so I think it doesn't have to be part of the git repository? Or, if intended, this could be uploaded to the GitHub release instead too?

This PR removes the binary, and also marks it as ignored in `.gitignore`.

Thanks for `cf-python` :+1: 
Bruno


davidhassell:
Hi Bruno,

The good news is the the dist/ directory is _not_ in the PyPi tar ball.

I do want to keep `cf-python-1.5.4.post6.tar.gz` as it was the last release before a major API change, and was developed in different repository that may not exist for ever.

I looked at attaching it to a GitHub release, but it appears that I can only attach the binary to a snapshot of the current repo, which is not really how I want to store it. 

I'm going to think about it  for a bit ... more suggestions welcome!

All the best,
David

sadielbartholomew:
Hi @kinow (nice to converse with you on GH again!),

Sorry for the delay. Replying on behalf of @davidhassell who's on leave, having had time to reconsider he thinks it is sufficient to have the binary up on PyPI, & I agree, so we are happy to merge. You've also initiated a solution towards #21, which is great. Thanks for both contributions.

kinow:
Always great to work with you @sadielbartholomew !

PR #5, opened 2019-10-25
Simplify comparisons

kinow:
This PR replaces `var == True` by a simple `var` test, and also uses `is None` instead of `== None` (see [ref](http://jaredgrubb.blogspot.com/2009/04/python-is-none-vs-none.html))

### DIFF
--- a/cf/bounds.py
+++ b/cf/bounds.py
@@ -60,7 +60,7 @@ the `nc_set_dimension`, `nc_get_dimension`, `nc_del_dimension` and
             return True
 
         if nbounds == 4 and data.ndim == 3:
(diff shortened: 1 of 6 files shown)


davidhassell:
Hi,

I won't merge this, not because I disagree with it (on the contrary), but because it highlighted a slight change in the functionality that was needed in `cf.functions._section`, and it was easy to implement all of the other `is None` changes by hand, as there weren't many of them.

Many thanks

kinow:
Thanks @davidhassell ! Some tests were failing in my environment (before the change too), so I couldn't confirm what exactly would be impacted by this.

PR #7, opened 2019-10-31
Minor changes to read.py

kinow:
Hi,

Starting reading the code from the `read` function in the `README.md` example. This PR changes a few things in the code that should not affect the behaviour, but simplify the code or protect against possible issues.


cf/read_write/read.py
```diff
@@ -540,8 +540,6 @@ def read(files, external=None, verbose=False, warnings=False,
     if aggregate and len(field_list) > 1:
         if verbose:
             org_len = len(field_list) # pragma: no cover
-            
-        if verbose:
```
kinow: Same as above :point_up:

cf/read_write/read.py
```diff
@@ -598,7 +596,7 @@ def _plural(n): # pragma: no cover
     return 's' if n !=1 else '' # pragma: no cover
 
 
-def _read_a_file(filename, aggregate=True, aggregate_options={},
+def _read_a_file(filename, aggregate=True, aggregate_options=None,
```
kinow: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

cf/read_write/read.py
```diff
@@ -630,6 +628,8 @@ def _read_a_file(filename, aggregate=True, aggregate_options={},
             The fields in the file.
 
     '''
+    if aggregate_options is None:
+        aggregate_options = {}
```
kinow: Part of the removal of the default mutable argument above.

cf/read_write/read.py
```diff
@@ -678,10 +678,8 @@ def _read_a_file(filename, aggregate=True, aggregate_options={},
     # ----------------------------------------------------------------
     # Still here? Read the file into fields.
     # ----------------------------------------------------------------
-    cdl = False
     if ftype == 'CDL':
         # Create a temporary netCDF file from input CDL
-        cdl = True
```
kinow: `cdl` is never used.

PR #19, opened 2019-11-25
Docs (tutorial): add context for referenced functions

sadielbartholomew:
(As per #17, I will add a commit to build changes into the HTML files, once I can work out why I am getting errors trying to build the documentation.)

In the ['Inspection' section of the documentation tutorial](https://ncas-cms.github.io/cf-python/tutorial.html?highlight=lbproc#inspection) there is reference to the built-ins ``repr()`` & ``str()``, but in the subsequent code examples those built-ins are not called directly, rather called under the hood by calling, & by calling ``print()`` on, respectively, the field construct (or dataset, etc.).

To someone without relevant knowledge of Python to know that ``print()`` show the ``str()`` etc. I think it could be unclear as to how the code snippets & the functions referred to relate, so I suggest adding something similar to that proposed here to put that into context.

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -251,7 +251,8 @@ levels of detail.
 Minimal detail
 ^^^^^^^^^^^^^^
 
-The built-in `repr` function returns a short, one-line description:
+The built-in `repr` function, invoked on a variable by calling that variable
+alone at the interpreter prompt, returns a short, one-line description:
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


davidhassell:
Thanks for these changes - they do make things clearer.

It's probably best to build the docs (when you fix the path problem ...) with something like `make html dev`, as any changes to the html created by running `make html .` will, if committed, appear in the latest docs at https://ncas-cms.github.io/cf-python/, which would then include documentation for features that are not in the latest release (3.0.5).

PR #20, opened 2019-11-25
Code: consolidate some logic

sadielbartholomew:
I have spotted some instances of redundant logic or duplication which can be reduced, so here are some suggestions to simplify those cases.

I've checked that the test battery results are not affected by this change.


davidhassell:
Less duplication and easier to read - thank you.

Issue #1, opened 2019-10-07
Add support for reading CDL text files


davidhassell:

davidhassell:
All ready in for the next version (v3.0.2).

Try it with [test_file.zip](https://github.com/NCAS-CMS/cf-python/files/3696898/test_file.zip)


```
$ unzip test_file.zip
Archive:  test_file.zip
  inflating: test_file.nc          
$ ncdump test_file.nc > test_file.cdl
$ ncdump -h test_file.nc > test_file_h.cdl
$ ncdump -c test_file.nc > test_file_c.cdl
```

```
In [10]: import cf                                                                                                  

In [11]: f = cf.read('test_file.nc')[0]                                                                                                    

In [12]: g = cf.read('test_file.cdl')[0]                                                                                                   

In [13]: f.equals(g)                                                                                                                         
Out[13]: True

In [14]: cf.read('test_file_h.cdl')[0]                                                                                                     
Out[14]: <Field: eastward_wind(atmosphere_hybrid_height_coordinate(1), grid_latitude(10), grid_longitude(9)) m s-1>

In [15]: cf.read('test_file_c.cdl')[0]                                                                                                     
Out[15]: <Field: eastward_wind(atmosphere_hybrid_height_coordinate(1), grid_latitude(10), grid_longitude(9)) m s-1>
```

PR #46, opened 2020-04-07
Test #1 for GitHub Action test suite run workflow

sadielbartholomew:
Commit pushes to master don't seem to be triggering the Action as they should. But does it work for PRs?

[Hoping opening this will trigger an Action which will at least get to the 'Install dependencies' stage before failing on lack of UDUNITS. Not expecting to merge PR but to close it after seeing if anything happens in Actions]

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -1,0 +1,60 @@
+# A GitHub Action to run the cf-python test suite upon push or PR to master.
+name: Run test suite
+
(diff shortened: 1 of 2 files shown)


sadielbartholomew:
> [Hoping opening this will trigger an Action which will at least get to the 'Install dependencies' stage before failing on lack of UDUNITS. ... ]

Bingo! Why the Action was not triggered for a commit push to master remains a mystery, however...

sadielbartholomew:
Test complete.

PR #48, opened 2020-04-07
Test #2 for GitHub Action test suite run workflow

sadielbartholomew:
Commit pushes to master don't seem to be triggering the Action as they should, again, after that seemed to be working again. As in #46, a PR should work to trigger the Action workflow.

[Hoping opening this will trigger an Action which will fingers crossed will install the dependencies successfully and move on to running the test suite. Not expecting to merge PR but to close it after seeing how this goes in Actions.]


sadielbartholomew:
I've put an Issue into the repo which contains the downstream action to setup conda which I am trying to use, as that seems to be unavailable ATM, hence the workflow fails. I'll try again tomorrow...

PR #82, opened 2020-06-05
Logging w/ consolidation of verbosity kwargs & attrs

### LINKED ISSUE #37: Add logging (replace '_debug' & more) pre-parallelisation
More compound/intensive processes such as partitioning & subspace processing have logic dotted about to print relevant info if ``_debug`` is set for debugging purposes. It would be better to replace these with dedicated logging (& then extend the logging over time to cover more logic & include more detail for highly-verbose cases).

I think we should add basic logging *before* we start optimisation work in earnest because configurable well-placed logging calls will allow us to see clearly the state & location of the code as it runs, & then we e.g. only need to apply a custom handler to keep track of each process set-up by `mpi4py` if we go with that as a/the solution.

Logging would be good for various other reasons, both towards development & user support:

* we can log info at various levels instead of just a 'debug' level, & get relevant output for those cases;
* more control: we can save logs if appropriate & set where the logs are;
* in general, 'debug' implies something negative (bugs) whereas logs imply something positive (care), so if users see the latter they will have more confidence in `cf-python` I think.

My opinion is that Python's ``logging`` module would be the best tool to use, as it is simple & there can be seamless interfacing with logged info from key imported modules (e.g. ESMPy, NumPy).

sadielbartholomew:
Closes #37. Sister PR to NCAS-CMS/cfdm#34, with largely the same purpose & implementation, & using much of the corresponding functionality there (by import where possible) but with the following key differences:

* ``cf-python`` applies ``info`` & ``_debug`` keywords as well as just ``verbose`` to enable filtering on some level of verbosity. These keywords have all been converted to ``verbose`` for consistency, & appropriate level have been given to any call being filtered as such, e.g:
    ```python
    if _debug:
        print("my message")
    ```
    has been changed to the following, so it only emerges with debug logging configured locally via ``verbose`` keyword (see next point), or if not globally:
    ```python
    logger.debug("my message")
    ```
    where ``info`` levels have been mapped appropriately, as discussed, namely:

    | Min. verbosity | (Old) `info` level | New `verbose` level |
    |----------------|--------------------|---------------------|
    | warning        | 0                  | 1                   |
    | info           | 1                  | 2                   |
    | detail         | 2                  | 3                   |
    | debug          | 3                  | -1                  |

* unlike in ``cfdm`` there are cases where the verbosity is specified not via a function keywords arguments, but via an instance attribute e.g. ``self.verbose`` (or, pre-conversion, ``self.info``). To manage these I have created another decorator which is analagous to the existing ``_manage_log_level_via_verbosity`` but adapts the log level for runs of decorated methods according to the value of the ``verbose`` as an attribute instead as a keyword argument.

### Work still TODO for this PR

Currently a WIP with functionality largely there but docs & testing pending:

- [x] apply deprecation messages on ``info`` & ``_debug`` where appropriate;
- [x] adapt test verbose specifications;
- [x] add docs section;
- [x] adapt docstrings;
- [x] double check format conversions from multi-argument print statements;
- [x] add explicit ``assetLog``-based tests;
- [x] manual testing in anger (esp. check interfacing with ``cfdm`` logging);
- [x] add new function to API ref;
- [x] update changelog;
- [x] address DH feedback on docstring formatting.


cf/aggregate.py
```diff
@@ -1340,16 +1338,38 @@ def aggregate(fields,
...
-            default *info* is 0 and no information is displayed. If
-            *info* is 1 or more then display information on which
-            fields are unaggregatable, and why. If *info* is 2 or more
-            then also display the structural signatures of the fields
-            and, when there is more than one field construct with the
-            same structural signature, their canonical first and last
-            coordinate values. If *info* is 3 or more then display the
-            field construct's complete aggregation metadata.
+        verbose: `int` or `None`, optional
+            If an integer from `0` to `3`, corresponding to increasing
```
davidhassell, line 1342: Hi - minor point - elsewhere we use double backticks on Python instances (``1``, ``'foobar'``, etc) and only use single backticks on classes/functions/etc that can link to docs (`None`, `int`, etc.)

sadielbartholomew: Thanks, noted (for this PR but also for future work, as consistency is important!). I'll amend the new docstring captions accordingly. I did copy them directly from cfdm, so I should amend the cfdm ones similarly, once I have this PR ready to go.

sadielbartholomew: Addressed in f9a4ce3dac8086c02b9ac8db8ef17f759b43aff8, cheers.

sadielbartholomew:
I will address the two outstanding tasks while doing the release & awaiting the tests to complete for that.

Opening now to see how bad the merge conflicts are via the UI. Then I will see how the CI tests fare. I am seeing a sub-test error locally (`test_Datetime_Data (__main__.DatetimeTest) ... ERROR`) but I suspect it is because I need to update my branch so it includes everything that has since gone into `master`.

sadielbartholomew:
Conflicts resolved, now closing & re-opening to force the GH Actions checks to start.

sadielbartholomew:
Conflicts resolved, now closing & re-opening to force the GH Actions checks to start.

sadielbartholomew:
Trying the CI checks again, & I have added an install of `pycodestyle` so the corresponding, & new, style test can run via Actions.

sadielbartholomew:
This PR has now become muddied with some pre-release commits, notably PEP8 style fix lines & UM read tweaks to try to get some of the test failures to pass. I should really have done these post-PR but I did not think ahead.

Running the tests once more, as I was trying to pinpoint which test was giving a full debug log dump & think I have sorted that. It was due to updates that had gone into master & needed amending for the logging logic after incorporation here.

sadielbartholomew:
Final check before merge. I expect the test_PP_* tests to fail, as there is clearly an issue with the GH Actions environment/setup in this respect. Those tests pass locally so it is fine.

sadielbartholomew:
Just the three ``test_PP_*`` tests erroring as expected, so now good to merge.

Issue #70, opened 2020-04-30
Consolidate functions that get & set global settings


sadielbartholomew:
We have agreed that instead of having *multiple individual* functions that allow one to change or inspect from the return output the (previous or current) value of *a certain* global setting, there should be a *single* function that will enable the getting & setting of *all such* variables. (The existing individual functions can then be deprecated after a sufficient amount of time such that this would be an eventual API change.)

The existing functions that would be consolidated as such are those with *capitalised* names [listed as 'cf functions' in the documentation](https://ncas-cms.github.io/cf-python/function.html), e.g. ``cf.``:

* ``ATOL`` & ``RTOL``;
* ``SET_PERFORMANCE``;
* ``RELAXED_IDENTITIES``.

and should include, as a new global setting, the lowest-severity logging level when logging is implemented (see #37).

It should still be workable on a single-setting basis, which is simple enough as users can specify the setting & desired value by keyword argument, & an obvious return structure would be a dictionary such that they can use indexing as standard to pick out the item they are interested in.

sadielbartholomew:
I am including de-capitialisation of the names of the functions for getting & setting the various global settings, as suggested in https://github.com/NCAS-CMS/cfdm/pull/34#issuecomment-626741031, as part of the Issue (I am in the process of resolving it).

PR #93, opened 2020-07-06
Lower-case names & aliases for global settings functions

sadielbartholomew:
Towards #70: renaming the global settings functions to the lower-case equivalents as the first step because changes towards that are necessarily spread across the codebase, whereas the new consolidation function will live in a single module & be tested in one other.

Merging via PR to ensure the CI tests pass. This will rely on recent groups updates to `cfdm` having been accounted for in `cf` so we may need to wait until that point for the tests to have a chance of running cleanly (locally I have been using an older branch before the groups changes went in).


sadielbartholomew:
Hmm, getting the right `cfdm` setup so the tests run with the appropriate state here is tricky. We need the renamed lower-case global functions that are used from `cfdm` to be available (NCAS-CMS/cfdm@b12ae033f61a7ec9c79402a6191d6c80828246c9), but need a state without the groups work which was added as that requires the ``netcdf_flattener`` which needs manual install steps on the GH Actions workflow. And in fact they have been added in a non-linear fashion so that these changes are mixed together, so it is not as simple as needing to test against a specific commit (or one within a small window of) the `cfdm` `master` branch in GH Actions, which would be a pain but could be done.

So I think the best plan in context is to ensure the tests run & pass locally, where it won't be disruptive to get the specific `cfdm` state required, simply running off one of my `cfdm` development branches for the global settings functions renaming where I had not rebased onto master to incorporate the groups work. Starting the local test suite configured as such now...

sadielbartholomew:
Local test suite, as described above, completed. I just get some PEP8 errors, mostly introduced in commits from this PR:

```pytb
test_Field_concatenate (test_Field.FieldTest) ... ok
test_DomainAxis__repr__str (test_DomainAxis.DomainAxisTest) ... ok
test_DomainAxis (test_DomainAxis.DomainAxisTest) ... ok

FAIL: test_pep8_compliance (test_style.styleTest)
  File "/home/sadie/cf-python/cf/test/test_style.py", line 49, in test_pep8_compliance
    'Detected {!s} PEP8 errors or warnings:'.format(pep8_issues)
AssertionError: 5 != 0 : Detected 5 PEP8 errors or warnings:

Ran 264 tests in 2619.773s

FAILED (failures=1)
```
and running that test fixture:
```bash
test_pep8_compliance (__main__.styleTest) ... ../functions.py:70:1: W293 blank line contains whitespace
../functions.py:74:1: E302 expected 2 blank lines, found 1
../functions.py:297:1: E302 expected 2 blank lines, found 1
../functions.py:317:26: E222 multiple spaces after operator
../read_write/um/umread.py:2197:17: E129 visually indented line with same indent as next logical line
FAIL

```

so I will push a final commit to fix those. Then this is good to merge as far as is possible to tell without lots of Actions workflow twiddling.

sadielbartholomew:
Merging despite CI failures as justified above.

Issue #85, opened 2020-06-15
Determine interaction of namesake constants from cfdm


sadielbartholomew:
There are some functions tied to constants that are named identically in ``cfdm`` & ``cf``, e.g. `ATOL`, `RTOL` & `LOG_LEVEL`. Most if not all have the same role in each of the standalone libraries. But we noted that we are not sure how these interact in cf-python. We should check how they interact and that the behaviour is as expected (which may involve deciding how they should interact, if at all).

As an example, ``cf.LOG_LEVEL`` has been setup, initially at least, to only (directly) change the logging of modules in ``cf``, but we may want to influence the log messages emerging in ``cf`` from the ``cfdm`` module, e.g. by having any setting of ``cf.LOG_LEVEL`` also simultaneously set ``cfdm.LOG_LEVEL``. In fact, this may not be required because the Python logging module used in both cases may internally integrate the logging by the two libraries, such that changing the severity level on the root logger of ``cf`` may via change it effectively for the root logger of ``cfdm`` (in which case it the ``cfdm`` loggers would probably change in behaviour without a corresponding change in the ``cfdm.LOG_LEVEL`` as coded which would be a problem). But until we investigate I do not know.

This Issue may require no changes, but we should examine the interaction as otherwise this unknown could be or become a source of bugs.

A related question is whether we think it at all viable that users would need to or want to import *both* libraries:

```python
import cf
import cfdm

<mixture of API calls to cf & to cfdm>
...
``` 
as in that case we will need to be thorough to ensure setting of any ``cf`` global does not influence the equivalent ``cfdm`` global and behaviour on ``cfdm`` calls in non-intuitive or undocumented ways. And if importing both would be a bad idea, we may wish to document that users should import either ``cf`` or ``cfdm``, not both, depending on their needs.

sadielbartholomew:
Work on #70 in #95 has been illuminating on this front. It seems all the functions tied to constants that are named identically in `cf` and `cfdm` behave as they should, with the exception of `atol` &  `rtol` (& their aliased old upper-case old-name equivalents).

Namely, because these are (unlike the others) assigned to the equivalent function from `cfdm`:

https://github.com/NCAS-CMS/cf-python/blob/94c86149024af8bff9a44c3e668a5015a31d56d8/cf/functions.py#L285-L290

[...]

PR #95, opened 2020-07-08
New function cf.configuration with unit test

### LINKED ISSUE #70: Consolidate functions that get & set global settings

sadielbartholomew:
Final step towards & therefore closes #70, after work to convert all functions to get & set individual global constants to lower-case was completed in #93. This is equivalent to NCAS-CMS/cfdm#61 but is more complicated as there are many more global constant settings to include for `cf` (13 vs. 3 for `cfdm`).

Opening as a WIP because in implementing this, particularly the unit test, I noticed that `cf.atol` & `cf.rtol` are detached from the corresponding constants in `cf.CONSTANTS`, instead getting & setting the `cfdm` equivalent (see #85 where I will detail it further). I would like to fix this first before completing this PR as it causes the unit test to fail (for valid reason). Best wait until the underlying issue is addressed to merge with a passing unit test (without bodging it temporarily).

Once that is discussed & resolved, I still want to extend the unit test for ``cf.configuration`` added here slightly, before removing the WIP status.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -20,6 +20,13 @@ version 3.6.0
 * New method: `cf.DomainAxis.nc_set_dimension_groups`
 * New method: `cf.DomainAxis.nc_clear_dimension_groups`
 * New keyword parameter to `cf.write`: ``group``
+* New function: `cf.configuration`
+* Renamed to lower-case (but otherwise identical) names all functions which
+  get and/or set global constants: `cf.atol`, `cf.rtol`, `cf.log_level`,
+  `cf.chunksize`, `cf.collapse_parallel_mode`, `cf.free_memory`,
+  `cf.free_memory_factor`, `cf.fm_threshold`, `cf.of_fraction`,
+  `cf.regrid_logging`, `cf.set_performance`, `cf.tempdir`, `cf.total_memory`,
+  `cf.relaxed_identities`. The upper-case names remain functional as aliases.
 * Changed dependency: ``1.8.6<=cfdm<1.9.0``
 * Changed dependency: ``cfunits>=3.2.8``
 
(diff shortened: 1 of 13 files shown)


sadielbartholomew:
The CI is failing at the dependency stage, because the imminent version of `cfdm` is set as required, but that is not worth amending since the tests would also fail for reasons required to the latest `netcdf_flattener` not being packaged in a way that makes it practicable to install on the GH Actions job environment.

I ran the test suite locally and the only failure was for PEP8 compliance which I will address in a final commit:

```
...
test_Field_domain_axis_position (test_Field.FieldTest) ... ok
test_Field__mul__ (test_Field.FieldTest) ... ok

FAIL: test_pep8_compliance (test_style.styleTest)
  File "/home/sadie/cf-python/cf/test/test_style.py", line 49, in test_pep8_compliance
    'Detected {!s} PEP8 errors or warnings:'.format(pep8_issues)
AssertionError: 66 != 0 : Detected 66 PEP8 errors or warnings:

Ran 274 tests in 724.515s

FAILED (failures=1)
```

sadielbartholomew:
Local test suite now passes. Merging despite CI failures as justified above.

PR #108, opened 2020-08-20
Docstring rewriter

davidhassell:
Allow reuse of docstrings in different classes, and of docstring snippets.


cf/test/test_read_write.py
```diff
@@ -462,26 +445,15 @@ def test_read_write_string(self):
...
         for string0 in (True, False):
             for fmt0 in ('NETCDF4',
-                         'NETCDF3_CLASSIC',
-                         'NETCDF4_CLASSIC',
-                         'NETCDF3_64BIT',
-                         'NETCDF3_64BIT_OFFSET',
-                         'NETCDF3_64BIT_DATA'):
-                # print ('fmt0=', fmt0)
-                f0 = cf.read(self.string_filename)
+                         'NETCDF3_CLASSIC'):
```
sadielbartholomew, line 451: Is the removal of the loops to test the other formats ('NETCDF4_CLASSIC' etc.) intentional here? Raising this in case you removed them from the sequence for development only, such that we should put them back as they were.

Ditto for the equivalent removal just below in the change set.

cf/test/test_docstring.py
```diff
@@ -0,0 +1,182 @@
...
+import unittest
+
+import cf
+
+
+class DocstringTest(unittest.TestCase):
+    def setUp(self):
+        self.package = 'cf'
+        self.repr = 'CF '
+        self.subclasses_of_Container = (
```
sadielbartholomew, line 11: I am thinking that instead of listing these out directly (and in the equivalent `cfdm` test), we could generate such subclasses lists via the `__subclasses__()` method, e.g. `cf.mixin.properties.Container.__subclasses__()`, perhaps requiring some recursive logic on that.

I'm not sure if you thought of that and decided against it. If not, I guess it would be more maintainable in the sense that we would not need to update it with any new classes if ever added, and of course no classes will get missed out, but it is less explicit and maybe explicitness is more important for a test. I'll leave that here as a thought, given there are pros and cons.

cf/field.py
```diff
@@ -626,15 +626,15 @@ def analyse_items(self, relaxed_identities=None):
      'aux_coords': {'N-d': {'aux0': <CF AuxiliaryCoordinate: latitude(110, 106) degrees_N>,
                             'aux1': <CF AuxiliaryCoordinate: longitude(110, 106) degrees_E>},
                     'dim0': {'1-d': {},
-                             'N-d': {}},
+                             'N-d': {},},
```
sadielbartholomew, line 629: Would you recommend we now do this, or similar, to prevent natural ``}}`` cases in docstrings, then, say if writing a new one? I only see the test to detect unprocessed `{{ ... }}` expressions fail with both opening and closing double braces , but I suppose this is to be safe in case somehow the former could get added beforehand?

sadielbartholomew (commented):
Generally this is looking great. I have a few minor comments & questions, for now (some as sub-comments on the code in question).

1. In case you are unaware, the branch has conflicts that need resolving. Then we can see the CI test results, which will be useful.

    Locally, the test suite passes for me except for the `test_style` sub-test `test_pep8_compliance` test PEP8 `test_pep8_compliance` test fails due to `../read_write/um/umread.py`.which fails due to several error-code cases. Would be good to fix that up before we run on GH Actions in case that failure cancels parallel jobs.

2. I've noticed that ``{{ ... }}`` aren't processed, and aren't detected as not being processed by `test_docstring`, if they in the docstrings of any (leading & trailing) dunderscore methods. E.g. adding one to `cf.Data.__contains__`:

    ```diff
    diff --git a/cf/data/data.py b/cf/data/data.py
    index 4a79b2bdc..14ac2f7cb 100644
    --- a/cf/data/data.py
    +++ b/cf/data/data.py
    @@ -948,20 +948,24 @@ place.
             self._set_Array(compressed_array, copy=False)

         def __contains__(self, value):
             '''Membership test operator ``in``

         x.__contains__(y) <==> y in x

         Returns True if the value is contained anywhere in the data
         array. The value may be a `cf.Data` object.

    +    {{verbose: `int` or `str` or `None`, optional}}
    +
    +    {{ make test fail }}
    +
         **Examples:**

         >>> d = Data([[0.0, 1,  2], [3, 4, 5]], 'm')
         >>> 4 in d
    ```

    The test still passes. Is that deliberate? If not, is it a problem?

cf/test/test_read_write.py
davidhassell: This was intentional, as the other omitted formats collapse into the `NETCDF3_CLASSIC` case - there is no code bypassed here, fortunately.

davidhassell:
@sadielbartholomew re. your first comment: 

1) the pep8 stuff is now fixed (https://github.com/NCAS-CMS/cf-python/pull/108/commits/1aef9dbda667d0de8722558dd9bb13e7c905c875, good point)

2) That should indeed have failed. The latest commit (https://github.com/NCAS-CMS/cf-python/pull/108/commits/b2c5b88cda2e6c6a7e1d51e0c210bba11dd86362) should fix this

cf/test/test_read_write.py
sadielbartholomew: Ah, fair. All good in that case. Thanks.

cf/test/test_docstring.py
sadielbartholomew: Agreed, best merge before doing any follow-on work.

sadielbartholomew: Happy to try to sort that out tonight myself for both cf-python & cfdm.

sadielbartholomew:
Going to close & re-open to trigger the GH Actions jobs to run (need to re-look at our triggering at some point to find a better way to do this!).

sadielbartholomew:
Oh, of course the GH Actions jobs are failing on requiring our `master` development version of cfdm, for its recently-added docstring re-writing capabilities, rather than the latest released version. I've made a note to set some option in the Actions workflow whereby we can run off the dev cfdm instead of a released version. For now & in this case, I'll just go by the local test suite result, so ignore those failures.

sadielbartholomew (approved):
Fantastic. GH Actions fail on dependencies as explained in my previous comment, but I just got a full test suite pass locally & it took ~1/3 of time it used to take, without unintended consequences, given the changes to the testing here.

Docstring substitution works really well, and I can't find any further issues with it. Feedback addressed such that double-brace patterns left over in dunderscore functions are now raised as failures by the test.

cf/field.py
davidhassell: @sadielbartholomew I think this is now not necessary - as we have dropped the test for `}}` in the docstrings. It seemed unfair to test for it, as it is valid python (unlike `{{`)

davidhassell:
Thanks, @sadielbartholomew

PR #118, opened 2020-09-16
Test update to GitHub Actions (run against cfdm master)

sadielbartholomew:


sadielbartholomew:
Okay, first try has failed because it seems the relative directory structure of the job host VM isn't as I thought it was. The next commit I push in a moment tweaks the specified directories to change into before running `pip install -e .` to install development versions of `cfdm/` and then `cf-python/`.

sadielbartholomew:
Opening & closing to re-trigger.

sadielbartholomew:
The run on cfdm master is working for all jobs except Python 3.5 ones where we are hitting a separate issue whereby netCDF4 isn't being installed successfully, I suspect this is due to it or one of its dependencies having dropped support for 3.5 since there is a deprecated warning given so 3.5 may soon be unsupported. Investigating, to confirm or otherwise, now.

sadielbartholomew:
Yes, `numpy` now requires Python version 3.6 or greater (see https://github.com/numpy/numpy/blob/f1f1436b853f3855d75667ab4055cce06960fda9/setup.py#L464) hence the failure of the Python 3.5 jobs on the new 'install development...' step. So those failures are unrelated to the change to using cfdm master.

I will remove 3.5 from the run matrix for now, but @davidhassell we should discuss whether we should drop support for deprecated 3.5 ourselves (FYI it [reaches "end of life"](https://www.python.org/downloads/release/python-3510/#:~:text=Python%203.5%20will%20reach%20its,release%20of%20the%203.5%20series.) this month).

sadielbartholomew:
I think this should be the last time I need to re-trigger, since the Python 3.X jobs for X != 5 all reached the 'run test suite' step for the previous try, so should pass without 3.5 jobs failing to cancel them like before. Let's see...

sadielbartholomew:
Hmm, one spurious test failure that I can't recreate locally:

```
FAIL: test_Data_exp (test_Data.DataTest)
  File "/home/runner/work/cf-python/cf-python/main/cf/test/test_Data.py", line 2830, in test_Data_exp
    self.assertTrue((d.array == c).all())

Ran 286 tests in 775.721s

FAILED (failures=1)
```

Trying once more in case it is a non-deterministic test.

sadielbartholomew:
Okay the same test as above has failed again on the Python 3.8 job, so I will investigate that separately.

The run against cfdm master aspect is finalised, since the jobs run to completion even if they fail, so this is not good to merge.

PR #120, opened 2020-09-23
Creation commands

davidhassell:
Move creation_commands methods from cf-python to cfdm, as mentioned in https://github.com/NCAS-CMS/cfdm/issues/53

Companion PR to https://github.com/NCAS-CMS/cfdm/pull/82


sadielbartholomew:
Re-triggering the tests here (via open-close) now that NCAS-CMS/cfdm#82 has been merged such that the creation commands have become available to cf-python from that, where they were removed in this PR, such the tests should pass unless further work needs to be done to get the right end-to-end behaviour.

sadielbartholomew:
Interestingly there are a couple of CI failures (plus one test failure probably unrelated since we've seen it before, which I will look into), both on `test_docstring`. Locally they also fail when I run them separately (`python test_docstring.py`), but they pass if run as part of `python run_tests`.

(Next week at normal working hours) it would be good to see if you see the same locally, David. It is certainly possible I am doing something silly, but I may not be...

Otherwise I will leave this to you. Giv

sadielbartholomew:
Oh, it looks like you have already solved the pair of test failures in NCAS-CMS/cfdm#83. :+1:

davidhassell:
Ah yes - sorry I forgot to mention that one. I think we can merge, now.

PR #121, opened 2020-09-29
Test Actions update (generate & upload coverage reports)

sadielbartholomew:
Test and, if/when working, merge, an update to the GH Actions `run_test_suite` workflow to upload a(XML format) coverage report for one representative job to the Codecov coverage service. See individual commit messages for the sub-steps required to achieve this.

Equivalent to a combination of the coverage-related PRs in cfdm, NCAS-CMS/cfdm#52, NCAS-CMS/cfdm#79 and NCAS-CMS/cfdm#84, but for cf-python.

The outcome I am looking for is for a Codecov check to run once the relevant job reaches that step in the workflow. The Codecov integration should then post a comment with coverage stats and a link to the full report. It won't have a "pull request base" until this initial report at least has been submitted. (Note the tests themselves may fail due to one known issue, which is not relevant to this PR specifically, though will be fixed elsewhere.)


sadielbartholomew:
> The outcome I am looking for is for a Codecov check to run once the relevant job reaches that step in the workflow.
...
(Note the tests themselves may fail due to one known issue, which is not relevant to this PR specifically, though will be fixed elsewhere.)

Ah, I did not anticipate that the test failing on one job would cancel the rest and hence the coverage report generation & upload step would not be reached (which is by-design and generally good but means my original test plan needs rethinking). The best way forward is for me to solve the test failures first. Doing that now...

sadielbartholomew:
Re-triggering via open-close.

sadielbartholomew:
Re-triggering via open-close.

sadielbartholomew:
It looks like the test suite now passes, but unfortunately I had missed amending the path to the coverage report when copying & pasting the coverage-uploading step from the working cfdm workflow. So the coverage upload step failed. Retriggering again to re-try. It looks like it should work this time.

sadielbartholomew:
Tests pass for all jobs and the reports are uploaded to Codecov, but the Codecov bot doesn't seem to have commented with a report for some reason. Possibly it is the merge conflicts that have just cropped up (it seems some of the same fixes were made in parallel). Addressing the conflicts and then I will retry once more to check the conflicts were resolved okay and to try to get a Codecov bot comment, but it is not essential. Possibly something I put in the YAML config has disabled it, anyway.

sadielbartholomew:
Interesting, it looks like there is a MacOS platform-specific failure in the test suite. I will investigate that separately.

No Codecov bot comment as I hoped (may need to tweak the new YAML config), but the coverage is [definitely being uploaded](https://codecov.io/gh/NCAS-CMS/cf-python/commits) now as was the aim of this PR, so I consider this good to merge.

Issue #115, opened 2020-09-15
Aggregation parameters: mismatch between actual & doc'd


sadielbartholomew:
There are two discrepancies between the documented parameters of the `cf.aggregate` function and the actual parameters it accepts, namely:
* `cf.aggregates` takes the undocumented parameters, with defaults, `ignore=None` & `exclude=False`;
  https://github.com/NCAS-CMS/cf-python/blob/db0a9f82394f9c68f701bf3c09ea3185f02a39bc/cf/aggregate.py#L1293-L1294
* the [corresponding API reference describes](https://ncas-cms.github.io/cf-python/function/cf.aggregate.html?highlight=ignore) undefined parameters `exist_ignore` and `equal_ignore` (with no specific defaults indicated in the summary text):

    ```pycon
    >>> a = cf.example_field(0)
    >>> b = cf.aggregate(a, equal_ignore=False)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/sadie/cfdm/cfdm/decorators.py", line 186, in verbose_override_wrapper
        return method_with_verbose_kwarg(self, *args, **kwargs)
    TypeError: aggregate() got an unexpected keyword argument 'equal_ignore'
    >>> b = cf.aggregate(a, exist_ignore=False)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/sadie/cfdm/cfdm/decorators.py", line 186, in verbose_override_wrapper
        return method_with_verbose_kwarg(self, *args, **kwargs)
    TypeError: aggregate() got an unexpected keyword argument 'exist_ignore'
    ```

(I double-checked and found those are the only mismatches.)

It seems most likely given the equal number of parameters missing in each case that the pair of undocumented parameters were intended as the pair of documented but undefined parameters, but both having been given the wrong name in the API reference. I just want to check that this is indeed the case, and whether the mapping is (as there is not much in the logic to indicate potential matches & from the dev history it seems the mismatch was there [as far back as at least 2015](https://bitbucket.org/cfpython/cf-python/src/660265561a63e269a44c853b167f9cbff63e10e5/cf/aggregate.py?at=master)):

* `ignore` -> `equal_ignore`;
* `exclude` -> `exist_ignore`;

or vice versa. Arrows indicate the direction of change, as well, since to avoid an API change it seems preferable to re-document the mapping than to make an equivalent re-name of the parameters, do you agree @davidhassell ?

[...]

PR #123, opened 2020-09-29
property clarification

davidhassell:
See #115

### DIFF
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -1437,10 +1437,6 @@ def aggregate(fields,
             `~cf.Field.long_name`), with the same values. See the
             *concatenate* parameter.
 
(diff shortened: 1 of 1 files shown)


sadielbartholomew (approved):
This looks and works fine. I think the new single argument and description of it is much clearer than the two  ones previously intended.

I recently added the stub for a test to check the various aggregate arguments work:

https://github.com/NCAS-CMS/cf-python/blob/777e83eaa3daef2f6768e65915918fb72efe2b91/cf/test/test_aggregate.py#L130-L133

so it would be good to add a quick test of this new `ignore`  argument, though I can pop in a commit to do that afterwards if you prefer to merge this as-is.

davidhassell:
Thanks. Sounds like it might be easier to merge this and put the test in separately.

PR #124, opened 2020-09-29
Shields for quick links to testing & coverage results

sadielbartholomew:
Add a few new shields to the README to make accessible from a single click the:

* latest coverage reports and visualisations of those; &
* the latest Actions CI job runs and results.

I am merging this via PR rather than committing directly by `push` as usual for a single self-contained change because a coverage report has not yet been picked up for the `master` branch despite some of my commits having uploaded reports. I think it is because the Codecov logic was at that point on a branch under my userspace rather than in the `cfdm` `master` itself. Until there is a coverage report as a base for the `master` branch the new coverage shield will read 'coverage | unknown' which is not what we want.


sadielbartholomew:
I thought doing another PR could generate the coverage report via the `master` branch but again it has just gone off my own forked branch. Will have to temporarily enable triggering upon commit pushes to get the report on `master` (see 47e53261f19606368a5c39db2144ab791de257a0). I will have to see how others manage to get reports from `master` with a PR from a forked branch, as for cf-python we don't want to trigger with pushes generally, to save on Actions resource minutes.

Can merge this commit, anyhow.

sadielbartholomew:
Yep, all is good with the README shields after enabling trigger by push separately to this PR:

shields-scr

PR #137, opened 2020-10-20
Workflow to create and test source dist for all PRs

### LINKED ISSUE #136: Separate Action workflow to test on new source distribution
In the lead up to release 3.7.0 we observed several sub-tests throw errors when run against a freshly-created source distribution (i.e. via `python setup.py sdist` then `./test_release <vn>`), even though they passed when the test suite ran against the git repo codebase (i.e. via `python run_tests.py` in the `cf/test/` dir).

In order to spot any such errors as they arise instead of only at release time, we can have an Actions workflow job which will run the source distribution test (just for one platform and Python version I think, no need to run it for each matrix combo).

This work will also help towards #134.

sadielbartholomew:
Closes #136 by adding a new workflow to run a pair of jobs, one for each of the latest OSs (Ubuntu & Mac OS), both with Python 3.7 as a representative version, where each job:

1. sets up the required cf-python environment by installing dependencies etc.;
2. creates the source dist from the state of the checked-out codebase;
3. (and ultimately) tests the source dist via `./test_release <vn>`.

I am putting this in via PR instead of pushing the commit because I want to check the new two jobs run and triggering is set only for PR, not for pushes (there is no need to test the sdist that often as the results shouldn't vary *too* much from the standard test suite results, if at all). I don't expect the jobs to succeed, however (as there are errors still under investigation which are likely to fail at least one job and then the rest will be cancelled automatically).

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -140,5 +140,5 @@ jobs:
     # End with a message indicating the suite has completed its run
     - name: Notify about a completed run
       run: |
-        echo The cf-python test-suite has completed and you may now
+        echo The cf-python test suite has completed and you may now
         echo inspect the results.
--- a/.github/workflows/test-source-dist.yml
+++ b/.github/workflows/test-source-dist.yml
@@ -1,0 +1,109 @@
+# A GitHub Action to test a source distribution created from the repo state.
+name: Test source distribution
+
+# Triggers the workflow on PR events for the master branch (only)
+on:
+  pull_request:
+    # 'reopened' enables manual retrigger via close & re-open. Disable for all
+    # edits to manage limited resource (PRs often change before merge-ready).
+    types: [opened, reopened, ready_for_review]
+    branches:
+      - master
+
+jobs:
+  source-dist-test:
+    # Set-up the build matrix.
+    # Note: only use one Python version, but it is easier to update in future
+    # by setting it here and refering to it as ${{ matrix.python-version }}.
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: [3.7]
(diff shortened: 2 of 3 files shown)


sadielbartholomew:
Eight jobs have just started running, including the two new ones to test the sdist, so all is good with respect to the aim of this PR. I will just wait to see the job outputs before merging, as they will help towards diagnosing some errors and hanging of tests we are seeing and trying to fix.

sadielbartholomew:
I forgot I need to change directory to get to the cf-python repo root dir given we check out both the cf-python and cfdm repos into the workspace for these jobs. So the the `test_release` was not found.

I've corrected this now and will close-open to re-trigger.

sadielbartholomew:
Another fix added: the slightly contrived command I used to get the version required to test on the right sdist required cf-python to be installed, but I forgot I had removed the `pip install` command to install it so that the cf

sadielbartholomew:
The jobs succeeded very quickly. It turns out the script didn't have the exit status of the test run, and was failing early due to not having `cfunits` & possibly other libraries installed. I forgot that some dependencies aren't installed from the conda channels such that `pip install -e .` was also getting some dependencies.

I've set the script exit code and added a line to install the dependencies without installing cf-python itself. The jobs should now run and report appropriately (with failure status expected). Final re-trigger hopefully...

sadielbartholomew:
All is good to merge with the current Actions check results: though most jobs are hanging and getting cancelled after the maximum job run time the job logs show that the pair of sdist test jobs are creating the sdist and running the tests on it.

I am looking into the cause of the hanging jobs and will fix that separately. I'll try to do that locally where we also see hanging on the sdist (but not standard codebase) testing as hanging jobs are rather wasteful on limited Actions resource minutes.

PR #143, opened 2020-11-05
Tests: diagnose cause of indefinite hanging or seg fault

sadielbartholomew:
Somewhere between the previous release and the one before that, i.e. 3.6.0 - 3.7.0, something changed and caused the test suite to consistently hang or occasionally even fail by seg fault both locally and on Actions (on some jobs, in turn cancelling the other jobs), especially when run for the source dist rather than on the cloned/forked local repo.

I think I've isolated one test method that is causing issue, so this is a PR to check the outcome of Actions with that test excluded, and to house a thread on thoughts about what might be going wrong.


davidhassell:
Interesting! A benign looking test, put one which does involve interaction with files on disk. I look forward to seeing if it all goes through ...

sadielbartholomew:
Retriggering by open/close...

sadielbartholomew:
So far I saw that the commenting-out of 'test_Field_close` did not solve anything, despite that sub-test seemingly causing issue locally, with almost all of the tests still hanging. Next I removed the whole `test_Field` module since that at least seems to be causing problem in some way. Then all but one module hung, where I am not seeing any pattern in the tests that hang or seg fault.

I will now try removing (obviously later in the PR I will revert those removals after isolating any issues, or close altogether) `test_CoordinateReference` as the one test where something hung in the that one job from the last Actions run.

sadielbartholomew:
Naturally when I randomise to see what is happening, it stops happening... :expressionless:

It's definitely not the randomisation in itself as that has been working without issue across many releases. But an issue cropping up only for a certain ordering of the tests would make sense, explaining the non-determinism and lack of pattern in the job failures.

Let me re-add `test_Field` to see if all jobs still pass in a strict alphabetical order.

sadielbartholomew:
I have just worked out that `test_Field_close` being a culprit is a red herring as a re-build of the `umread` library fixed that from consistently seg-faulting locally. At least something in `test_Field` is not happy, though.

sadielbartholomew:
I've just cancelled all of the most recent run of jobs since they are all clearly hanging based on time spent running (~1hr20, with ~20 expected).

Okay, I think I am quite convinced now that at the least that `test_Field` is leading to some issues causing hanging or seg faults, but not confident that there aren't other test modules also leading to some non-deterministic issues. I'll close this for now whilst I try and narrow the issues down further locally, but may open this PR again for further Actions testing.

sadielbartholomew:
I've just cancelled all of the most recent run of jobs since they are all clearly hanging based on time spent running (~1hr20, with ~20 expected).

Okay, I think I am quite convinced now that at the least that `test_Field` is leading to some issues causing hanging or seg faults, but not confident that there aren't other test modules also leading to some non-deterministic issues. I'll close this for now whilst I try and narrow the issues down further locally, but may open this PR again for further Actions testing.

Issue #150, opened 2020-11-10
How does cf.aggregate work?


lguo-uk:
Hi,

This is a recurrent issue. I still can not figure out how to make `cf.aggregate` work.

I have two files: `ta_Amon_MIROC6_ssp119_r1i1p1f1_gn_201501-202412.nc` and `ta_Amon_MIROC6_ssp119_r1i1p1f1_gn_202501-203412.nc`. They are CMIP6 air temperature on pressure levels. Their metadata are identical expect time coordinate (see attachments). 

However, when read in using `ta=cf.read("ta_Amon_MIROC6_ssp119_r1i1p1f1_gn_*.nc",select="air_temperature")`, I've got a cf.FieldList instead of a cf.Field:
```
[<CF Field: air_temperature(time(120), air_pressure(19), latitude(128), longitude(256)) K>,
 <CF Field: air_temperature(time(120), air_pressure(19), latitude(128), longitude(256)) K>]
```

Then, I tried to aggregate them using cf.aggregate(). I've tried the following commands:
```
cf.aggregate(ta,relaxed_identities=True)
cf.aggregate(ta,allow_no_identity =True)
cf.aggregate(ta,field_identity="long_name")
cf.aggregate(ta,ncvar_identities=True)
```
, but nothing changed.

I also tried to debug this issue with argument `verbose`. However, no matter what value that I set to `verbose`, there was not any extra information shown.

Any idea?

Many thanks,
Liang

# Attachment 1: ncdump -c information
```
netcdf ta_Amon_MIROC6_ssp119_r1i1p1f1_gn_201501-202412 {
dimensions:
	time = UNLIMITED ; // (120 currently)
	plev = 19 ;
	lat = 128 ;
	lon = 256 ;
	bnds = 2 ;
variables:
	double time(time) ;
		time:bounds = "time_bnds" ;
		time:units = "days since 1850-1-1" ;
		time:calendar = "gregorian" ;
		time:axis = "T" ;
		time:long_name = "time" ;
		time:standard_name = "time" ;
	double time_bnds(time, bnds) ;
	double plev(plev) ;
		plev:units = "Pa" ;
		plev:axis = "Z" ;
		plev:positive = "down" ;
		plev:long_name = "pressure" ;
		plev:standard_name = "air_pressure" ;
	double lat(lat) ;
		lat:bounds = "lat_bnds" ;
		lat:units = "degrees_north" ;
		lat:axis = "Y" ;
[...]

PR #151, opened 2020-11-12
Re-establish configurable verbosity for cf.aggregate

sadielbartholomew:
Fix bug alluded to in #150 (though it is not the core of the issue raised there, so it does not resolve it), namely that the verbosity keyword is ineffectual:

> I also tried to debug this issue with argument verbose. However, no matter what value that I set to verbose, there was not any extra information shown.

This turned out to be because a decorator was not applied to a central method for logging output during aggregation, `print_info`. A one-line change to add the decorator fixes the problem.

To ensure it did fix it and for robustness I have added a unit tests for the verbosity on `cf.aggregate` specifically (though the verbosity decorators are themselves unit tested, `cf.aggregate` in particular warrants a specific test that the verbosity is managed correctly since it involves a mix of the two decorators, `_manage_log_level_via_verbosity` and `_manage_log_level_via_verbose_attr`, and there could be bugs emerging from combined use).

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,12 @@
+version 3.?.?
+-------------
+----
+
+**2020-??-??**
+
+* Fixed bug that prevented the verbosity from changing to any value specified
+  as a ``verbose`` keyword parameter to `cf.aggregate` (only).
+
 version 3.7.0
 -------------
 ----
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -1049,6 +1049,7 @@ class _Meta:
 
         return anc_identity
 
+    @_manage_log_level_via_verbose_attr
     def print_info(self, signature=True):
         '''TODO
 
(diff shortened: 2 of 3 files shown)


sadielbartholomew:
(Cancelling the Actions jobs are they are hanging and we are still investigating the cause, so we will most likely waste resources as many of them run until they are cancelled after 3 hours.)

sadielbartholomew:
@davidhassell this is a one-line fix to add a decorator and in the self-contained, so should be innocuous and not negatively influence anything else, but because I can't run the full test suite without hanging on my laptop you might wish to. I'll leave this unmerged for now in case you do. I have run `test_aggregate`, `test_FieldList` and `test_Field` (except `test_Field_close` with seg faults so I am commenting it out for the time being) which have all passed. Thanks.

davidhassell:
I've fetched your branch and tested it with the test and on the command line. All looks good, so I'll merge.

Thanks.

PR #148, opened 2020-11-09
Extend indexed assignment and binary operations to bounds, if present.

davidhassell:
Fixes #146

PR #155, opened 2020-11-19
Allow global constants to be controlled by a context manager

### LINKED ISSUE #85: Determine interaction of namesake constants from cfdm

### LINKED ISSUE #154: Allow global constants to be controlled by a context manager
This is so we can stop doing things like:

```python
>>> old = cf.log_level('DEBUG')
>>> <execute some code>
>>> cf.log_level(old)
```
and start doing things like:

```python
>>> with cf.log_level('DEBUG'):
...     <execute some code>
...
>>>
```

The core implementation is inherited from cfdm (https://github.com/NCAS-CMS/cfdm/issues/100), which needs to be extended for the extra constants in extra cf-python

davidhassell:
Fixes #154 
Fixes #85

Hi @sadielbartholomew 

Note that I removed the unsettable constants from `cf.configuration` because they upset using the output to reset the global config. I.e. 

```python
>>> old = cf.configuration()
>>> cf.confiuration(old)
```

did not work. We should review this another day, at the same time as thinking about code re-use between cfdm and cf-python in this area.

I also made cf-python copies of ATOL and RTOL to address https://github.com/NCAS-CMS/cfdm/pull/103#issue-523871611

Thanks.

### DIFF
--- a/cf/functions.py
+++ b/cf/functions.py
@@ -54,7 +54,9 @@ import cfdm
 import cfunits
 
 from .          import __version__, __file__
-from .constants import CONSTANTS, _file_to_fh, _stash2standard_name
+from .constants import (CONSTANTS,
+                        _file_to_fh,
+                        _stash2standard_name)
 
 from . import mpi_on
 from . import mpi_size
@@ -69,6 +71,16 @@ def _close_proc_meminfo():
 
 
 atexit.register(_close_proc_meminfo)
+
+
+# --------------------------------------------------------------------
+# Inherit classes from cfdm
+# --------------------------------------------------------------------
+Constant = cfdm.Constant
+Configuration = cfdm.Configuration
+
+Constant.__doc__ = Constant.__doc__.replace('cfdm.', 'cf.')
+Configuration.__doc__ = Configuration.__doc__.replace('cfdm.', 'cf.')
 
 
 class DeprecationError(Exception):
@@ -234,15 +246,15 @@ def configuration(
 
     :Parameters:
 
-        atol: `float`, optional
+        atol: `float` or `Constant`, optional
             The new value of absolute tolerance. The default is to not
             change the current value.
 
-        rtol: `float`, optional
+        rtol: `float` or `Constant`, optional
             The new value of relative tolerance. The default is to not
             change the current value.
 
-        tempdir: `str`, optional
+        tempdir: `str` or `Constant`, optional
             The new directory for temporary files. Tilde expansion (an
             initial component of ``~`` or ``~user`` is replaced by
             that *user*'s home directory) and environment variable
@@ -252,27 +264,27 @@ def configuration(
 
             The default is to not change the directory.
 
-        of_fraction: `float`, optional
+        of_fraction: `float` or `Constant`, optional
             The new fraction (between 0.0 and 1.0). The default is to
             not change the current behaviour.
 
-        chunksize: `float`, optional
+        chunksize: `float` or `Constant`, optional
             The new chunksize in bytes. The default is to not change
             the current behaviour.
 
-        collapse_parallel_mode: `int`, optional
+        collapse_parallel_mode: `int` or `Constant`, optional
             The new value (0, 1 or 2).
 
-        free_memory_factor: `float`
+        free_memory_factor: `float` or `Constant`, optional
             The new value of the fraction of memory kept free as a
             temporary workspace. The default is to not change the
             current behaviour.
 
-        log_level: `str` or `int`, optional
+        log_level: `str` or `int` or `Constant`, optional
             The new value of the minimal log severity level. This can
             be specified either as a string equal (ignoring case) to
-            the named set of log levels or identifier 'DISABLE', or an
-            integer code corresponding to each of these, namely:
+            the named set of log levels or identifier ``'DISABLE'``,
+            or an integer code corresponding to each of these, namely:
 
             * ``'DISABLE'`` (``0``);
             * ``'WARNING'`` (``1``);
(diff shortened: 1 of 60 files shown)


davidhassell:
@sadielbartholomew I have just seen that the `.rst` have been corrupted (I suspect by a newer version of sphinx that installed itself against my wishes) - I shall fix them.

davidhassell:
This PR now also requires NCAS-CMS/cfdm/pull/104

sadielbartholomew:
> Note that I removed the unsettable constants from cf.configuration because they upset using the output to reset the global config

That's okay, in fact for the best I would say, as values that can't be set aren't strictly 'configuration'. That and they were a bit of a pain to handle as a special case, logic-wise!

I believe all PR dependencies are now merged, so I'll start reviewing.

sadielbartholomew:
I think this PR will close #85 too, by the way.

davidhassell:
I'll merge #148 into this PR, so that `cf.bounds_combination_mode` can get the "class" and "with" treatment

sadielbartholomew:
> so that cf.bounds_combination_mode can get the "class" and "with" treatment (or is that a bad idea?)

No, I think that is a good idea. If a user does something silly with the newly-acquired flexibility then that is their own fault...

sadielbartholomew:
I guess this is ready for review now? If not, please let me know when it is ready.

davidhassell:
All good - please do review, thanks.

sadielbartholomew (changes requested):
The functionality otherwise looks good (so far, not quite finished testing that aspect), but I've spotted that the context manager settings persist if it exits with an exception, for example note how the `rtol` and `chunksize` are permanently changed when it errors (cleanly, as standard) with a bad log level:

```pycon
>>> import cf
>>> orig = cf.configuration()
>>> print(orig['log_level'], orig['rtol'], orig['chunksize'])
WARNING 2.220446049250313e-16 82873466.88000001
>>> with cf.configuration(log_level=4, rtol=5e-10, chunksize=5e10):
...     print(cf.configuration)
... 
ValueError: 4 is not a valid ValidLogLevels

During handling of the above exception, another exception occurred:

  File "/home/sadie/cf-python/cf/functions.py", line 364, in configuration
    return _configuration(
  File "/home/sadie/cf-python/cf/functions.py", line 439, in _configuration
    reset_mapping[setting_alias](new_value)  # ...run corresponding func
  File "/home/sadie/cfdm/cfdm/functions.py", line 926, in __new__
    cls._CONSTANTS[cls._name] = cls._parse(cls, arg)
  File "/home/sadie/cfdm/cfdm/functions.py", line 1242, in _parse
    elif cls._is_valid_log_level_int(arg):
  File "/home/sadie/cfdm/cfdm/functions.py", line 227, in _is_valid_log_level_int
    ValidLogLevels(int_log_level)
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/enum.py", line 309, in __call__
    return cls.__new__(cls, value)
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/enum.py", line 600, in __new__
    raise exc
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/enum.py", line 584, in __new__
    result = cls._missing_(value)
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/enum.py", line 613, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 4 is not a valid ValidLogLevels
>>> new = cf.configuration()
>>> print(new['log_level'], new['rtol'], new['chunksize'])
WARNING 5e-10 50000000000.0
```

Possibly this is the case for the logic in `cfdm` (I'll check in a moment) but unfortunately if it is, I missed it there.

So it seems like we need a tweak to the PR (and possibly the cfdm merged code) to handle resetting of any temporary changes to values by the context manager upon tear down with an exception, generally. A possible way to do that would be with a `try`/`except`/`finally` block. I actually used one for a similar case of resetting in any closure scenario with the `manage_log_level_via_verbos{ity, e_attr}` decorators, e.g. here in `cf-python` for the latter:

https://github.com/NCAS-CMS/cf-python/blob/2f252374e23fc0f996f694ececefc7a46f64b400/cf/decorators.py#L142-L162

davidhassell:
Good spot, and thanks for the tips. I have put in a fix for cfdm (https://github.com/NCAS-CMS/cfdm/pull/107), which I shall follow up here, too.

sadielbartholomew:
Thanks @davidhassell, continuing with a (re-)review inclusive of the new fix commit now and I'll finish it tonight.

cf/functions.py
```diff
@@ -434,8 +430,22 @@ def _configuration(_Configuration, **kwargs):
...
+
+    old_values = {}
+
+    try:
+        # Run the corresponding func for all input kwargs
+        for setting_alias, new_value in kwargs.items():
+            reset_mapping[setting_alias](new_value)
+            setting = setting_alias.replace('new_', '', 1)
+            old_values[setting_alias] = old[setting]
+    except ValueError:
```
sadielbartholomew, line 442: I was wondering if there might be cases missed where exceptions other than `ValueError` are raised, especially here in cf-python with many more constants where some error could be encountered in trying to set them, but it does seem that at the moment it is only possible to run into a `ValueError` (sometimes raised in that form after being caught as e.g. an `AttributeError` or a `TypeError`) so that is all fine.

I'm mainly noting this since, in future, if we add exception handling for the constants they should ultimately raise only `ValueError`, else they should be added to this `except` catch.

sadielbartholomew (approved):
Feedback regarding the issue with persistence on tear down has been addressed.

The context manager works and is well-tested. The documentation has been updated appropriately. Great stuff. Happy to merge.

PR #159, opened 2020-11-27
Refactor and test for cf.environment

sadielbartholomew:
Some amendments relating to `cf.environment` which don't modify its behaviour in any way, namely:

* remove a defunct keyword argument from the signature;
* refactor the code which had quite a bit of duplication and wasn't as readable as it could be;
* add a basic test.

### DIFF
--- a/cf/functions.py
+++ b/cf/functions.py
@@ -4,6 +4,7 @@ import os
 import platform
 import re
 import resource
+import sys
 import ctypes.util
 # import cPickle
 import netCDF4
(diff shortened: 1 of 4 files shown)


sadielbartholomew:
I've added a commit to address a few minor issues which were not related to this PR as raised by one of the Actions jobs runs of the test suite.

However, I won't re-trigger the Actions jobs as some of them are hanging ATM, as is under our investigation, so will run to full time and waste Actions resource. The relevant tests did pass on that one job I saw and locally are passing. So I have as much confidence as I can in the circumstances and therefore can merge this.

PR #179, opened 2021-01-20
Enable faulthandler for all test modules

sadielbartholomew:
This PR enables `faulthandler` for each test module, so that we can catch important information about any segmentation faults or timeouts going forward. Note that while [`pytest` enables `faulthandler` by default](https://docs.pytest.org/en/stable/usage.html#fault-handler), tests run under the `unittest` framework do not seem to have it enabled, as demonstrated by the lack of information captured in all previous and recent segmentation faults arising locally and on the CI, hence the need to add lines to explicitly import and enable it

Hopefully we will see very few seg faults once any current issue(s) leading to them are resolved, but it will always be useful to catch detail in the rare case they do occur, hence the permanent addition of these lines.

Adding these changes via a PR rather than a direct commit push to run the CI in order to capture some post-seg-fault traceback that I can use as evidence for a further issue on current seg fault problems across the test suite.

### DIFF
--- a/cf/test/create_test_files.py
+++ b/cf/test/create_test_files.py
@@ -3,6 +3,10 @@ import os
 import unittest
 
 import numpy
+
+import faulthandler
+faulthandler.enable()  # to debug seg faults and timeouts
+
 import netCDF4
 
 import cf
--- a/cf/test/run_tests.py
+++ b/cf/test/run_tests.py
@@ -3,6 +3,9 @@ import os
 from random import choice, shuffle
 import unittest
 
+import faulthandler
+faulthandler.enable()  # to debug seg faults and timeouts
+
 import cf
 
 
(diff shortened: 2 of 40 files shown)


sadielbartholomew:
It appears there are some genuine failures to resolve first, adding a new commit...

sadielbartholomew:
Re-triggering via open-close...

sadielbartholomew:
It looks like two tests have passed but the remainder will timeout, I'll check back in in a few hours to see the results.

sadielbartholomew:
Normally I would cancel after CI tests run for more than ~1.5 hour to preserve limited (free) resource so I will do that now as all jobs that haven't competed have clearly hung (expected to take ~20 minutes). I was wanting to see the timeout output and thought they'd time out after 3 hours as in the past but they've been going for 3.5+ hours so it looks like the limit is higher now. I certainly don't want to waste 6 hours worth of Actions VM time so I'll cancel now.

sadielbartholomew:
Good to merge since the changes here are minimal and designed to help to diagnose the mysterious timeouts which arose here on Actions...

PR #181, opened 2021-02-02
Bug fixes to __getitem__ and __setitem__

davidhassell:
Minor fixes to `__getitem__` and `__setitem__`

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -1117,15 +1117,24 @@ place.
             return self.copy()
 
         d = self
+        
+        auxiliary_mask = ()
+        try:
+            arg = indices[0]
+        except (IndexError, TypeError):
+            pass
+        else:
+            if isinstance(arg, str) and arg == 'mask':
+                auxiliary_mask = indices[1]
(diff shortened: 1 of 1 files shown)


sadielbartholomew (approved):
This looks good and I think I understand the motivation behind the core changes (you were deliberately terse to allow me to work it out, right!? :grinning: All I have in my notes from our meeting is that you were about to put up a PR to "improving the indexing a bit" so sorry if more was mentioned.).

The `test_Data` tests all pass when I run this also using the `cfdm` branch `upstream/dask_compatibility` and your `setitem` dask branch, so all seems fine. And outside of the seg faulting that I can't avoid and the minority of tests they obscure, nothing appears to have been adversely affected that the test suite can pick up on at least.

Thanks, I shall take good case of the `dask` branch henceforward...

davidhassell:
Thank, Sadie - all you say is right. The `__setitem__` unit test is very basic at this stage, more of a place holder, but I envisage that the units tests will grow over time. Most units test will not need changing much, though, as we're retaining the API (for now ...)

PR #186, opened 2021-02-12
Back in black: format code and set up pre-commit hooks

sadielbartholomew:
Equivalent to sibling PR NCAS-CMS/cfunits#21; see the description there which is precisely applicable here.

Opening as a draft for now, though will have completed the PR ready to merge later today, so we can discuss the appropriate way to address some (mainly) 'F821 undefined name' errors raised by `flake8` as it is not clear the intention of the code in these cases.


sadielbartholomew:
Hi @davidhassell, I'll quickly go in to address some conflicts but afterwards if you are able to check out this branch and run `flake8 cf/read_write/` and look at the non-trivial cases I left out from 4d083a7, to suggest how best to address those cases, that would be very useful. It is good to get a second opinion there in case removing the references to the undefined names is not the right thing to do in context, leading to (another form of) bug.

sadielbartholomew:
Okay, apologies for the delay, the merge conflicts are now resolved (had a spot of bother with development installs because the `requirements.txt` was out-of-sync with `cf/__init__.py` for the `cfdm` version restrictions after #185, also I ran the test suite locally a few times to ensure I hadn't broken something in the resolution).

sadielbartholomew:
Soldiering on, will leave some markers to indicate areas where a second opinion would be wise for you @davidhassell. Thanks.

sadielbartholomew:
The CI is failing on `cfdm` version which has not been updated. We can sort out the version constraints  and check the CI passes after I have done the run through with `docformatter` in moment, so I'll merge this now.

Issue #98, opened 2020-07-17
Linting on docstrings


sadielbartholomew:
It would be good to extend linting of the codebase in the ``test_style`` test module to check that the docstrings specifically comply with Python conventions. Besides consistency it will be especially useful to ensure the correct formatting of the auto-generated API reference.

There are various options for the linting but as a standalone docstring checker, [`pydocstyle`](https://github.com/PyCQA/pydocstyle) seems to give a practical amount of, & scope in, feedback, e.g:

```console
$ pydocstyle cf/domain.py 
cf/domain.py:1 at module level:
        D100: Missing docstring in public module
cf/domain.py:5 in public class `Domain`:
        D204: 1 blank line required after class docstring (found 0)
cf/domain.py:5 in public class `Domain`:
        D300: Use """triple double quotes""" (found '''-quotes)
cf/domain.py:19 in public method `__repr__`:
        D207: Docstring is under-indented
cf/domain.py:19 in public method `__repr__`:
        D300: Use """triple double quotes""" (found '''-quotes)
cf/domain.py:19 in public method `__repr__`:
        D401: First line should be in imperative mood (perhaps 'Call', not 'Called')
```

sadielbartholomew:
Forgot to close this: it was closed by #187.

Issue #182, opened 2021-02-03
Replace LAMA with Dask to parallelise cf.Data


sadielbartholomew:
**A master issue to register steps and progress towards migrating the current approach to domain-distributed array representation, Large Amounts of Massive Arrays (LAMA), with an approach based on the efficient task parallelism provided by the [Dask](https://docs.dask.org/en/latest/) library.**

The core aims, to be implemented in the separate branch [`dask`](https://github.com/NCAS-CMS/cf-python/tree/dask) for merging into `master` at the end of the work ready for a `v4.0.0` major release, are to:
- replace all LAMA logic with Dask-based logic,
- in doing so, parallelise all existing `cf.Data` operations,
- reproducing exactly the current API and functionality  so that there are no user-facing differences other than better performance (though after the work is done we agreed we might want to make API amendments).

*This comment is intended as a working document to flesh out with detail and new steps as they arise during the progress of the work. Please update and amend it as appropriate, or comment in the thread.*

### Step-by-step plan

**Relevant PRs will be linked here**

*[To be fleshed out shortly]*

### Open questions, points of discussion, etc.

See the comments thread for this Issue, else external locations as linked from the thread for meatier discussion points.

***

### Progress breakdown in terms of methods and attributes: OLD

:exclamation:  NOTE: superseded by #295 with new, easier-to-parse, table with notes and groupings, etc. :exclamation: 

The ~300 functional/user-facing `cf.Data` methods and attributes are outlined below in a pretty-printed dict where the value is a Bool to indicate if they have been "Daskified" on the `dask` branch so far. Copy and filter using Python if desired:

<details>
  <summary>Click to expand [updated by @davidhassell  2021-02-22, @sadielbartholomew 19.08.21 (upon work restart)]...</summary>
  
```python
complete = {
    'HDF_chunks': False,
        'Units': True,  # by DH
    '_HDF_chunks': False,
    '_Units': False,
    '_YMDhms': False,
    '__abs__': False,
    '__add__': False,
    '__and__': False,
    '__array__': False,
    '__bool__': False,
        '__contains__': True,  # by DH
    '__data__': False,
    '__deepcopy__': False,
    '__div__': False,
    '__doc_template__': False,
    '__docstring_package_depth__': False,
    '__docstring_substitutions__': False,
    '__eq__': False,
    '__float__': False,
    '__floordiv__': False,
    '__ge__': False,
        '__getitem__': True,  # by DH
    '__gt__': False,
[...]

PR #187, opened 2021-02-12
Docformatter hooks and docstring standardisation

sadielbartholomew:
Similar to NCAS-CMS/cfunits#22, but only to auto-format to, and set up a pre-commit hook for, `docformatter`, i.e. without (additionally) the formatting to abide by check pre-commit for `pydocstyle` compliance to Google-style docstrings.

Adapting the codebase to abide by `pydocstyle` will be postponed for cf-python because it is a manual process that would have to be done manually (and very likely incrementally), and we want to get started right away on the Dask-migration related work (#182) which requires a separate branch (`dask`) to keep in sync.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,6 +24,22 @@ repos:
       - id: black
         language_version: python3
 
+  # Also format docstrings in a consistent way, on top of 'black' requirements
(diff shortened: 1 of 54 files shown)


sadielbartholomew:
Cancelling th3 CI workflows because they will fail for ssame reaosn as

sadielbartholomew:
Closes #98.

Issue #180, opened 2021-01-22
Non-deterministic segmentation faults throughout test suite


sadielbartholomew:
Some months ago seg faults appeared en masse both locally (for multiple testers) and on GitHub Actions, hitting various modules in the test suite though occurring for specific individual modules sporadically. These faults have persisted. Though I hoped and tried somewhat to discreetly investigate and fix the source (we don't have any external contributors ATM so it didn't seem urgent to broadcast), it is proving quite difficult to pinpoint, so an Issue is overdue to register this.

Details about the observed seg faults are provided below. I intend this to become an evidence log of sorts to hopefully guide us to getting to the root of the problem.

ESMValGroup/ESMValCore#644 is possibly relevant because it indicates similar symptoms in ESMValCore. I had a chat with some of the ESMValGroup devs today to see if we can help each other in these potentially-linked investigations, so this Issue is also to assist them with comparisons.

### General details

1. We have not seen, or heard of anyone else seeing, any seg faults during actual cf-python usage, so they only seem to occur when running some or all of the tests;
2. The seg faulting occurs both on Actions and locally for both developers who have tried it on their machines.

### Affected test modules and methods

These are test modules which we have observed to seg fault at least once, though in most cases they do not *always* seg fault for a given environment (OS, conda and pip libraries etc.) and Python version. (I've been running `for filename in test_*.py; do python $filename; done` to run as many test methods as possible without a single seg fault stopping the experiment:

* **`test_Field`**: `test_Field_close`
* **`test_pp`**: ? [specific method(s) unknown]
* **`test_gathering`**: ?
* **`test_CoordinateReference`**: ?
* **`test_dsg`**: ?
* **`test_groups`**: ?
* **`test_read_write`**: `test_read_write_format`
 

### Example seg fault traceback

(Captured using `faulthandler` which I recently enabled for all of the test modules.)

```console
$ python test_groups.py 
Run date: 2021-01-06 18:21:22.812031
Platform: Linux-4.15.0-54-generic-x86_64-with-glibc2.10 
HDF5 library: 1.10.6 
netcdf library: 4.7.4 
Python: 3.8.5 /home/sadie/anaconda3/envs/cf-env/bin/python
netCDF4: 1.5.4 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/netCDF4/__init__.py
[...]

PR #195, opened 2021-03-18
Fix docstring example syntax errors and test Python 3.9

### LINKED ISSUE #132: Verify support for Python 3.9 (once it is more established)
As per the description in NCAS-CMS/cfunits#15, though in this case I have not yet tried to add 3.9 jobs to the workflow to run the test suite (if setting up the environment for cfunits fails at the moment, it certainly will do the same for cf-python).

In short: **in late 2020 or early 2021** once 3.9 is more established and probably supported by at least some dependencies, we should check whether our dependencies allow us to support 3.9, and document (perhaps package) whether or not 3.9 is supported.

sadielbartholomew:
Closes #132.

Now the linting pre-commit checks are passing after #194, I want to check if Python 3.9 passes the CI to confirm the resolution of #132. It passes for me locally as far as I can tell around the seg faults which are still occurring and under investigation (see #180), though the seg faults may also appear on the Actions jobs since we have seen them emerge there before.

Since the Actions jobs for the workflow to run the test suite are only set to be triggered by PR, I am putting in a trivial PR to trigger and inspect the jobs. These trivial commits fix  all of the syntax errors raised by doctest on the codebase (I have set up a CLI option `-d` so we can check these via `py run_tests.py -d` towards #4, though until they are all passing I won't add them to the test suite).

### DIFF
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -2871,7 +2871,7 @@ def _ok_coordinate_arrays(meta, axis, overlap, contiguous, verbose=None):
 
     **Examples:**
 
-    >>> if not _ok_coordinate_arrays(meta, 'latitude', True, False)
+    >>> if not _ok_coordinate_arrays(meta, 'latitude', True, False):
     ...     print("Don't aggregate")
 
     """
--- a/cf/cellmethod.py
+++ b/cf/cellmethod.py
@@ -404,7 +404,7 @@ class CellMethod(cfdm.CellMethod):
 
         >>> c
         <CF CellMethod: lat: lon: mean>
-        >>> c.intervals = ['0.2 degree_N', cf.Data(0.1 'degree_E')]
+        >>> c.intervals = ['0.2 degree_N', cf.Data(0.1, 'degree_E')]
         >>> c
         <CF CellMethod: lat: lon: mean (interval: 0.1 degree_N interval: 0.2 degree_E)>
 
--- a/cf/coordinatereference.py
+++ b/cf/coordinatereference.py
@@ -618,7 +618,7 @@ class CoordinateReference(cfdm.CoordinateReference):
         >>> r.coordinates
         {'atmosphere_hybrid_height_coordinate'}
         >>> r.change_coord_identitiers({
-        ...     'atmosphere_hybrid_height_coordinate', 'dim1', 'ncvar:ak': 'aux0'
+        ...     'atmosphere_hybrid_height_coordinate', 'dim1', 'ncvar:ak', 'aux0'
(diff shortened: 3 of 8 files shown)


sadielbartholomew:
Results are in... both the standard codebase-based (with 3.9) and the source dist (with 3.7) test jobs have failed (else been cancelled) due to one test method error emerging due to some spurious partitioning-based issue:

```pytb
ERROR: test_Field_indices (test_Field.FieldTest)
  File "/home/runner/work/cf-python/cf-python/main/cf/test/test_Field.py", line 1490, in test_Field_indices
    indices = f.indices(grid_longitude=cf.wi(50, 130))
  File "/home/runner/work/cf-python/cf-python/main/cf/field.py", line 13194, in indices
    b = self.flip(axis).anchor(axis, anchor1, dry_run=True)[
  File "/home/runner/work/cf-python/cf-python/main/cf/decorators.py", line 64, in precede_with_kwarg_deprecation_check
  File "/home/runner/work/cf-python/cf-python/cfdm/cfdm/decorators.py", line 49, in inplace_wrapper
  File "/home/runner/work/cf-python/cf-python/main/cf/field.py", line 15527, in anchor
    if dim.increasing:
  File "/home/runner/work/cf-python/cf-python/main/cf/dimensioncoordinate.py", line 259, in increasing
    return self.direction()
  File "/home/runner/work/cf-python/cf-python/main/cf/dimensioncoordinate.py", line 359, in direction
    _direction = self._infer_direction()
  File "/home/runner/work/cf-python/cf-python/main/cf/dimensioncoordinate.py", line 129, in _infer_direction
    data = data[0:2].array
  File "/home/runner/work/cf-python/cf-python/main/cf/data/data.py", line 7921, in array
    p_array = partition.array
  File "/home/runner/work/cf-python/cf-python/main/cf/data/partition.py", line 1153, in array
    p_data = p_data.reshape(p_shape)
ValueError: cannot reshape array of size 9 into shape (1,)

Ran 304 tests in 1246.062s

FAILED (errors=1)
```

@davidhassell should we be concerned about that error now, or assume it is some unknown facet of the partitioning that has been there for a long time as I believe we agreed last time?

The great news is no seg faults are happening (at least for the jobs that ran to completion and for the tests that ran before the cancellations otherwise)! Which is great as even with `faulthandler` it's difficult to diagnose and debug those.

Since the 3.9 test is passing beside the above test method in a way that is consistent with the jobs with the other minor Python versions, I am satisfied that 3.9 is running fine on the CI as well as locally. Hence:

* close #132;
* test completed so this PR is now good to merge.

Issue #197, opened 2021-03-24
NumPy MaskError with read-in of certain CDL files


sadielbartholomew:
With the current `cf` and `cfdm` master branches installed, as well as on the current release set, the reading of certain CDL files fails due to a `MaskedError` which is raised by numpy to indicate an issue with conversion of masked elements. Examples are given below.

## Details

```pycon
>>> import cf
>>> f = cf.read('contiguous.nc.cdl', fmt='CDL')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sadie/cfdm/cfdm/decorators.py", line 181, in verbose_override_wrapper
    return method_with_verbose_kwarg(*args, **kwargs)
  File "/home/sadie/cf-python/cf/read_write/read.py", line 667, in read
    fields = _read_a_file(
  File "/home/sadie/cfdm/cfdm/decorators.py", line 181, in verbose_override_wrapper
    return method_with_verbose_kwarg(*args, **kwargs)
  File "/home/sadie/cf-python/cf/read_write/read.py", line 941, in _read_a_file
    fields = netcdf.read(
  File "/home/sadie/cfdm/cfdm/decorators.py", line 181, in verbose_override_wrapper
    return method_with_verbose_kwarg(*args, **kwargs)
  File "/home/sadie/cfdm/cfdm/read_write/netcdf/netcdfread.py", line 1270, in read
    self._parse_ragged_contiguous_compression(
  File "/home/sadie/cfdm/cfdm/read_write/netcdf/netcdfread.py", line 1856, in _parse_ragged_contiguous_compression
    element_dimension = self._set_ragged_contiguous_parameters(
  File "/home/sadie/cfdm/cfdm/read_write/netcdf/netcdfread.py", line 2351, in _set_ragged_contiguous_parameters
    element_dimension_size = int(
  File "/home/sadie/cf-python/cf/data/data.py", line 1467, in __int__
    return int(self.datum())
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/numpy/ma/core.py", line 4383, in __int__
    raise MaskError('Cannot convert masked element to a Python int.')
```

As a test I also converted all of the other sample netCDF datasets from our tutorial to CDL:

```console
$ for filename in *.nc; do ncdump -h $filename > ${filename}.cdl ; done
```

[...]

PR #199, opened 2021-03-26
Short-term fix for errors in reading of metadata-only CDL

sadielbartholomew:
See #197 and #NCAS-CMS/cfdm#128 for details.

Note in response to the CI failing on coverage I've realised that even for this short-term fix I should add some more tests to cover metadata-only CDL, so I need to create and commit those before this can be merged.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -9,6 +9,9 @@ version 3.9.0
   zero size (https://github.com/NCAS-CMS/cfdm/issues/113)
 * Fixes for changes in behaviour in cftime==1.4.0
   (https://github.com/NCAS-CMS/cf-python/issues/184)
+* Better error message in the case of a `numpy.ma.core.MaskError` occurring
+  upon reading of CDL files with only header or coordinate information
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
The new commit adds testing to cover the cases of relevance. Opening and closing to re-trigger the CI tests...

sadielbartholomew:
Regarding the CI results, the first test job to run to completion failed, cancelling the others, but only due to one sporadic test failure we have seen often before (namely a "ValueError: cannot reshape" on `test_Field_indices (test_Field.FieldTest)`).

Since it's very likely at least one other test-running job will fail in the same way, I won't re-run the set, but can have reasonable confidence the changeset is sound from that one job outcome. We'll be running the full set of CI jobs again very soon for the release, anyhow.

As for the source dist testing job, it seems there may some issue in the `test_release` script, else in the workflow in question, but I'll look into that separately. Otherwise linting passes.

Overall, I'm satisfied I can merge this despite some CI failures.

PR #210, opened 2021-05-11
Performance improvements

### LINKED ISSUE #201: Improved performance of cf.read "select" parameter for PP/UM files
When selecting fields with the "select" parameter of `cf.read` with PP/UM file inputs, the currently selection occurs after all of the input files have been processed. This can be made much more efficient by avoiding processing the non-selected fields.

E.g.

```python
import re
import cf
f = cf.read('~/aaadpa#pg000003385c1+.pp', select=["stash_code=3237", "stash_code=1"])
g = cf.read('~/aaadpa#pg000003385c1+.pp', select=re.compile('^stash_code=\d\d\d\d$'))
```

### LINKED ISSUE #202: Change constructs access API to facilitate performance improvements
#### Current situation

Access to named constructs is currently either by the "filter_by_*" `Constructs` methods, or by Field attributes:

```python
>>> f = cf.example_field(0)
>>> f.dimension_coordinates
<Constructs: dimension_coordinate(3)>
>>> f.constructs.filter_by_type('dimension_coordinate')
<Constructs: dimension_coordinate(3)>
```

The reason for the "dimension_coordinates" alias being an attribute rather than method is that it was expected that it would often be followed by more options:

```python
>>> f.dimension_coordinates('time')
<Constructs: dimension_coordinate(1)>
>>> f.dimension_coordinates.filter_by_property(units='degrees_east')
<Constructs: dimension_coordinate(1)>
```

This is fine, but comes with the slight confusion that the result of the attribute is a callable `Constructs` instance, which obfuscates the help:

```python
>>> help(f.dimension_coordinates)
Help on Constructs in module cf.constructs object:

class Constructs(cfdm.constructs.Constructs)
 |  Constructs(auxiliary_coordinate=None, dimension_coordinate=None, domain_ancillary=None, field_ancillary=None, cell_measure=None, coordinate_reference=None, domain_axis=None, cell_method=None, source=None, copy=True, _use_data=True, _view=False, _ignore=())
 |
 |  A container for metadata constructs.
 ...  
```

#### Performance

It turns out that the code for accessing constructs is very slow, and must be improved.

A key part of this improvement relies on being able to work with python dictionaries rather than dictionary-like `Constructs` objects.

To facilitate this, whilst retaining the intuitive nature of the API the least intrusive change is to make what were attributes properties: i.e. `f.dimension_coordinate`s becomes `f.dimension_coordinates()`. This then allows keyword parameters that can change the behaviour when speed is an issue.
Code breaking

This change will break any code that uses bare construct access attributes:

```python
>>> # These won't work any more
>>> f.auxiliary_coordinates  
>>> f.coordinate_references
>>> f.coordinates  
>>> f.cell_measures
>>> f.dimension_coordinates
>>> f.domain_ancillaries
>>> f.domain_axes
>>> f.cell_methods
>>> f.field_ancillaries
```

```python
>>> # These will work as before
>>> f.auxiliary_coordinates()
>>> f.coordinate_references()
>>> f.coordinates()
>>> f.cell_measures()
>>> f.dimension_coordinates()
>>> f.domain_ancillaries()
>>> f.domain_axes()
>>> f.cell_methods()
>>> f.field_ancillaries()
```
```python
>>> # These will also work as before
>>> f.auxiliary_coordinates(x)
>>> f.coordinate_references(x)
>>> f.coordinates(x)
>>> f.cell_measures(x)
>>> f.dimension_coordinates(x)
>>> f.domain_ancillaries(x)
>>> f.domain_axes(x)
>>> f.cell_methods(x)
>>> f.field_ancillaries(x)
```
This is the only backwards incompatible change to the API. All other changes will not break existing code.

With the new API, reading a file is, in one reproducible test, ~3 times faster:
```python
>>> import cf, timeit
>>> f = cf.example_field(1)
>>> cf.write(f, 'tmp.nc')
>>> cf.__version__
3.8.0
>>> sum(timeit.repeat("cf.read('tmp.nc')", globals=globals(), repeat=100, number=1))/100
0.22618969043996912
```
```python
>>> import cf, timeit
>>> cf.__version__
3.9.0b1
>>> sum(timeit.repeat("cf.read('tmp.nc')", globals=globals(), repeat=100, number=1))/100
0.07341453492001165

```
Note that much of the improvement comes from unrelated changes (such as removing unnecessary `__repr__` calls and unnecessary deep copies).

See https://github.com/NCAS-CMS/cfdm/issues/130 for further details. In particular, the cfdm version of the above timing test gives a ~10 times speed up - the reasons for the lesser improvement in cf-python will need investigating

_Edit: The less speed-up is understandable and due to extra checks being carried out in, e.g., `set_construct`. This is not to say that things can't be sped up more, but that is for another issue_ 

PR to follow ...

### LINKED ISSUE #203: Add "item" keyword to construct selection methods
Add an `item` keyword to `construct`, `dimension_coordinate`, and similar methods that will return both a unique construct and its identifier in an efficient manner:

```python
>>> import cf
>>> f = cf.example_field(1)
>>> f.construct('latitude', item=True)
('auxiliarycoordinate0',` <CF AuxiliaryCoordinate: latitude(10, 9) degrees_N>)
```

### LINKED ISSUE #204: Change constructs access API for consistency
This enhancements imports changes from cfdm v1.8.9.0: https://github.com/NCAS-CMS/cfdm/issues/132 by making  the `*identity`, `**filter_kwargs` API available to construct selection methods:

```python
>>> import cf
>>> f = cf.example_field(1)
>>> f.domain_axes(filter_by_size=(cf.gt(1),))
<CF Constructs: domain_axis(2)>
>>> f.dimension_coordinates('X', 'Y')
<CF Constructs: dimension_coordinate(2)>
>>> f.coordinates('X', 'Y')
<CF Constructs: auxiliary_coordinate(2), dimension_coordinate(2)>
>>> f.coordinates('X', 'Y', filter_by_naxes=(1,))
<CF Constructs: dimension_coordinate(2)>

```

davidhassell:
Fixes #201
Fixes #202
Fixes #203
Fixes #204


davidhassell:
Hi @sadielbartholomew,

This is another rambling old PR (106 files changed, this time), but most changes are either style, linting, or the implementation of the new API throughout the library.

This PR basically brings in the cfdm changes from https://github.com/NCAS-CMS/cfdm/pull/138; adds some PP related optimisations; and looks a bit ahead to CF-1.9 (with  `fielddomainlist.py`)


When reviewing, you might search for 3.9.0 in the code, or I might recommend starting with:

    cf/core/constructs.py
    cf/constructs.py
    cf/mixin/container.py
    cf/mixin/*.py
    cf/mixin/fielddomain.py
    cf/mixin/fielddomainlist.py
    cf/field.py
    cf/fieldlist.py
    cf/read_write/um/umread.py [*]

[*] Interacts with `field.py` for some short-circuiting optimisations when setting metadata constructs on a field.


You might get away with just that, but I'm sure that you could also treat these files as spring board to other things.

Thanks.

sadielbartholomew:
Thanks for another helpful review guide @davidhassell. Once my append mode work is out for review I'll be straight onto this...

cf/mixin/coordinate.py
```diff
@@ -390,18 +412,220 @@ def autoperiod(self, inplace=False):
...
+
+        return super().identities(generator=generator, **kwargs)
+
+    def _ctypes_iter(self, ctypes):
+        """Generator for returning the coordinate type letter."""
+        for c in ctypes:
+            if getattr(self, c):
+                # This coordinate construct is of this type
+                yield c
+                return
```
sadielbartholomew, line 631: Is the `return` at the right indentation level here? Just to check since it logically I would have thought it should go one level down.

cf/mixin/fielddomain.py
```diff
@@ -0,0 +1,2958 @@
...
+        for c_key in ref.coordinates():
+            axes.extend(self.get_data_axes(c_key))
+
+        for da_key in ref.coordinate_conversion.domain_ancillaries().values():
+            axes.extend(self.get_data_axes(da_key))
+
+        return set(axes)
+
+    def _conform_coordinate_references(self, key, coordref=None):
+        """Where possible replace the content of coordiante reference
```
sadielbartholomew, line 73:
```suggestion
        """Where possible replace the content of coordinate reference
```

cf/mixin/fielddomain.py
```diff
@@ -0,0 +1,2958 @@
...
+                key0=identifier0,  # key0,
+                key1=identifier1,  # key1,
+                rtol=rtol,
+                atol=atol,
+                s=s,
+                t=t,
+                verbose=verbose,
+                axis_map=axis_map,
+            ):
+                # add traceback TODO
```
sadielbartholomew, line 255: There's a few TODOs for this function that might not have been addressed, in particular this one.

cf/mixin/fielddomain.py
```diff
@@ -0,0 +1,2958 @@
...
+                See the **kwargs** parameters of `indices` for
+                details.
+
+        :Returns:
+
+            `dict`
+                 The dictionary has two keys: ``'indices'`` and
+                 ``'mask'``.
+
+                 The ``'indices'`` key stores a dictionary in keyed by
```
sadielbartholomew, line 294: What is meant by "in keyed" (also used in the next paragraph)?

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+    def select_by_construct(self, *identities, OR=False, **conditions):
+        """Select elements by their metadata constructs.
+
+        To find the inverse of the selection, use a list comprehension
+        with the !match_by_construct` method of the construct
+        elements. For example, to select all constructs that do *not*
+        have a "latitude" metadata construct:
+
+           >>> gl = cf.{{class}}(
+           ...     f for f in fl if not f.match_by_construct('latitude')
```
sadielbartholomew, line 16: That looks like a generator expression rather than a list comp. as-is ~~so we should probably add the brackets around it unless the former is in fact required in which case a change of terminology might be suitable?~~ I just tried out the case given above and it errors with `TypeError: cannot create 'generator' instances` so it looks like the brackets are indeed required (tested as working):

```suggestion
           ...     [f for f in fl if not f.match_by_construct('latitude')]
```

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+class FieldDomainList:
+    """TODO."""
+
+    def select_by_construct(self, *identities, OR=False, **conditions):
+        """Select elements by their metadata constructs.
+
+        To find the inverse of the selection, use a list comprehension
+        with the !match_by_construct` method of the construct
```
sadielbartholomew, line 11:
```suggestion
        with the `match_by_construct` method of the construct
```

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+                dimensions, the collapse method, and any cell method
+                construct qualifiers. If N cell methods are described
+                in this way then they will collectively identify the N
+                most recently applied cell method operations. For
+                example, ``'T: maximum within years T: mean over
+                years'`` will be compared with the most two most
+                recently applied cell method operations.
+
+                *Parameter example:*
+                  `'latitude'``
```
sadielbartholomew, line 58: This is a duplicate to the one a few down so can be removed(and the formatting needs fixing, so cutting this one rather than the latter is quickest).

sadielbartholomew, line 58: This is a duplicate to the one a few down so can be removed(and the formatting needs fixing, so cutting this one rather than the latter is quickest).

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+    def select_by_ncvar(self, *ncvars):
+        """Select list elements by netCDF variable name.
+
+        To find the inverse of the selection, use a list comprehension
+        with the `!match_by_ncvar` method of the construct
+        elements. For example, to select all constructs which do *not*
+        have a netCDF name of 'tas':
+
+           >>> gl = cf.{{class}}(
+           ...     f for f in fl if not f.match_by_ncvar('tas')
```
sadielbartholomew, line 178: Ditto to a previous comment:

```suggestion
           ...     [f for f in fl if not f.match_by_ncvar('tas')]
```

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+        .. versionadded:: 3.0.0
+
+        .. seealso:: `select`, `select_by_identity`, `select_by_property`,
+
+        :Parameters:
+
+            ncvars: optional
+                Select constructs from the list. May be one or more:
+
+                  * The netCDF name of a construct.
```
sadielbartholomew, line 190: A single-item list seems a bit confusing to me, as if something has been cut off. It might be clearer to state it in the sentence, something lke:

```suggestion
                Select constructs from the list. May be one or more netCDF names of constructs.
```

cf/mixin/fielddomainlist.py
sadielbartholomew: And it looks like there are some cases of this underlying issue in the code itself too, rather than in the docstrings. See comments below.

sadielbartholomew (commented):
I am mid-review and will finish up on Sunday evening, but here are some initial minor comments.

cf/mixin/fielddomainlist.py
sadielbartholomew: There's a similar case of this too in a subsequent method.

cf/mixin/fielddomainlist.py
sadielbartholomew: Or possibly `` `! `` ...

sadielbartholomew (commented):
I've completed my review and I am very happy with this: all associated issues have been addressed satisfactorily and notably the performance has indeed improved for both netCDF and PP selection reads, as tested with similar code snippet examples as provided in those issues and backed up by shorter times to run a sample of the tests (due to seg faults I can't currently run all of them at once locally). Great work. Users will really appreciate the performance gains :rocket: 

I'v registered some minor comments in-line for your consideration. As a final note I see you've updated the docs here and otherwise all changes look to have been included, but I can't see `FieldDomainList` or `FieldDomain` from a search in the docs generated from this branch, so they might need adding to the class listing unless there's a reason you have left them out?

cf/read_write/read.py
```diff
@@ -4,21 +4,25 @@
 from glob import glob
 from os.path import isdir
 
-from numpy.ma.core import MaskError
```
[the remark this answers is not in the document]
davidhassell: Thanks - sorted.

cf/mixin/coordinate.py
davidhassell: I think this is OK, as we want this method to short circuit on the first coordinate type (T, X, Y or z) that gives a match.

cf/mixin/fielddomain.py
davidhassell: Sorry - in the "in" is rogue. Should be 
```
                 The ``'indices'`` key stores a dictionary keyed by
                 domain axis identifiers, each of which has a value of
                 the index for that domain axis.
```

cf/mixin/fielddomainlist.py
davidhassell: Interesting - works for me:
```python
In [1]: import cf

In [2]: f = cf.example_field(3)

In [3]: cf.FieldList(x for x in [f] if x.match('precipitation_flux'))

Out[3]: [<CF Field: precipitation_flux(cf_role=timeseries_id(4), ncdim%timeseries(9)) kg m-2 day-1>]

In [4]: 
```

cf/mixin/fielddomainlist.py
davidhassell: back-tick bang is right (I struggled to get the markdown right light you did!)

cf/mixin/fielddomainlist.py
davidhassell: Sorted

cf/mixin/fielddomainlist.py
davidhassell: See https://github.com/NCAS-CMS/cf-python/pull/210#discussion_r633289983

cf/mixin/fielddomainlist.py
```diff
@@ -0,0 +1,339 @@
...
+
+                By default (or if the *mode* parameter is ``'and'``) a
+                construct is selected if it matches all of the given
+                properties, but if the *mode* parameter is ``'or'``
+                then a construct will be selected when at least one of
+                its properties matches.
+
+            properties: optional
+                Select the constructs with the given properties. May be
+                one or more of:
```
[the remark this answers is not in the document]
davidhassell: fixed

davidhassell:
> As a final note I see you've updated the docs here and otherwise all changes look to have been included, but I can't see FieldDomainList or FieldDomain from a search in the docs generated from this branch, so they might need adding to the class listing unless there's a reason you have left them out?

These were left out along with all of the other mixin classes. Do you think that the mixins should be in docs? I think I left them out because they don't work as stand alone objects and don't form part of the public API

cf/field.py
```diff
@@ -3239,76 +2904,57 @@ def _regrid_get_latlong(self, name, axes=None):
...
-                    raise ValueError("TODO")
+                    raise ValueError("TODO 222222")
 
-                lon_axes = self.get_data_axes(lon_key)
-                lat_axes = self.get_data_axes(lat_key)
+                lon_axes = data_axes[lon_key]
+                lat_axes = data_axes[lat_key]
                 if lat_axes != lon_axes:
-                    raise ValueError("TODO")
+                    raise ValueError("TODO 3333333")
```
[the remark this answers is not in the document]
davidhassell: Agreed. Let's address these after this release ....

cf/mixin/fielddomain.py
davidhassell: Happy to leave this one till after the release, as the traceback is "merely" informative, rather than functional ...

sadielbartholomew:
> These were left out along with all of the other mixin classes. Do you think that the mixins should be in docs? I think I left them out because they don't work as stand alone objects and don't form part of the public API

Ah, that makes sense. Thanks for the clarification. I see, I hadn't noticed or thought about the mixin classes in relation to the docs before but on reflection it sounds like a good idea to keep them out of the docs for the reasons you state. Though maybe, in future, we could somehow include them in a 'development' section or something along those lines for full transparency for the small number of people it may be useful for.

cf/mixin/coordinate.py
sadielbartholomew: Ah, yes, I've played around with a toy example of an equivalent structure and I think I see now. :+1: Resolving this.

cf/mixin/fielddomain.py
sadielbartholomew: Sure, noted.

cf/mixin/fielddomainlist.py
sadielbartholomew: I did have to look up how to render the markdown there!  FYI, it turns out that if you do double ticks and leave a space between them and the input, you can escape the back-ticks, whereas black-slash escaping doesn't seem to work, e.g.  `'' he'llo ''` but replacing `'` with `` ` ``.

davidhassell:
> Though maybe, in future, we could somehow include them in a 'development' section or something along those lines for full transparency for the small number of people it may be useful for.

Good idea - let's put up an issue for it.

cf/mixin/fielddomainlist.py
sadielbartholomew: Oh, sorry, my bad - it does indeed seem to work :grimacing: . I tested with your example here and also with `[f]` replaced by `cf.FieldList([f, f])`. Looking back at my review notes I was mistakenly calling the method on the `FieldList` rather than the `FieldList` itself.

If it works without the need for list comp. brackets then even better. Nice!

cf/mixin/fielddomainlist.py
sadielbartholomew: Apologies again, it was working fine and I tested the wrong thing to come to the conclusion the brackets were needed.

sadielbartholomew (approved):
Excellent, all feedback has been addressed well, most notably that `cf.example_fields()` works without arguments.

The tests are passing locally (with no seg faulting :tada: ) and with much improvement in speed. I am not sure  why the Actions jobs haven't been triggered and run, but we will test those as part of the release checklist so that can and will be done shortly and therefore it's OK.

Please go ahead and merge!

P.S. When we update the changelog for the imminent releases of for both cf-python and cfdm, I think it would be good to explicitly highlight that performance (generally or specifically) should have improved significantly.

davidhassell:
Thanks for review, @sadielbartholomew, merging now.

Issue #222, opened 2021-05-27
Remembering weights between consecutive regridding operations


ThibHlln:
Given two fields `fld1` and `fld2`, if I use for example `regrids` twice, it seems to take the same amount of time the first and the second time.

```python
fld3 = fld1.regrids(fld2, 'conservative')
fld4 = fld1.regrids(fld2, 'conservative')
```

So I am wondering if `cf` is able to remember the weights from the first time it did the regridding in order to make the second time faster?

sadielbartholomew:
Hi @ThibHlln, thanks for the very good question as usual!

I will double check in the API and code (if I don't comment further assume it is indeed the case) but I suspect that, at the moment, there is not a way to do so. However, caching the weights is certainly a good idea (if, as I assume in practice, you mean a non-trivial case on , not just repeating the exact same regridding operation twice i.e. maybe the metadata differs?) and I think it would not be difficult to implement something that could do that.

We can mark it as a (performance) feature for the next version, as improving performance is one of the core tasks we are working on at the moment and going forward.

davidhassell:
Hi - it certainly was the case that caching the weights was _not_ possible in ESMPy (up to <= 8.0, I think). However, there was a new release a few weeks back, and I must confess that I only skim-read the release notes ... I'll go back and check them more closely for gold. We have engaged with the ESMF team in the past and requested this feature, so they do know that it is desired, but I don't know how hard it is for them to implement.

Thanks,
David

ThibHlln:
Hi Sadie and David,

@sadielbartholomew Yes, you are right, my case was rather trivial. Just to give some context, I need to regrid repeatedly (i.e. iterating over time) from one modelling component resolution to another (specified as `cf.Field`, invariant over time). So, only the data would change between iterations, the metadata would remain unchanged. Sp more something like this:

```python
# first iteration
fld1.set_data(...)
fld3 = fld1.regrids(fld2, 'conservative')
arr3 = fld3.array

# second iteration
fld1.set_data(...)
fld4 = fld1.regrids(fld2, 'conservative')
arr4 = fld4.array

# etc.
```

[...]

Issue #63, opened 2020-04-24
Consolidate & tidy imports in import-heavy modules


sadielbartholomew:
Understandably it is not high on the developer priority list, but from working with certain modules with lots of ``import`` statements, I've noticed certain items aren't grouped together &/or are in places I would not expect them to be relative to other imports.

Looking at ``data.data`` & ``field`` which are particularly import-heavy, it seems that generally (though there are outliers to change to follow suit) the intended order is, with alphabetical ordering as appropriate inside each:

1. General Python in-built modules [though some at position 4 for ``data.data``];
2. SciPy & NumPy;
3. Other CF modules e.g. ``cfdm``;
4. Relative ``cf-python`` imports (all ``..<name>`` then all ``.<name>``);
5. ``mpi4py``-related.

I'm happy to do a small tidy on those modules. More generally, though, this Issue could act as a document describing the desired organisational rules so that we are aware where to put any ``import`` statements we are adding to keep them in a known order as we go.

@davidhassell before I rearrange some out-of-place imports, can I check with you that the above ordering is the one intended & still intended? I also have a few questions. Some items are listed out separately for no obvious reason. Can & should the following sets be combined (via e.g. ``from .data import (...)``) or are they  intentionally like so?:

https://github.com/NCAS-CMS/cf-python/blob/defa44fff6d146da1703bf4133af56afff19a14f/cf/field.py#L73-L77

and:

https://github.com/NCAS-CMS/cf-python/blob/defa44fff6d146da1703bf4133af56afff19a14f/cf/field.py#L45-L57

With something like this, it does seem appropriate to use multiple groups, but instead ones organised by some categories e.g. one for the ``as`` statement, one private methods & one for the rest (all internally alphabetical)?:

https://github.com/NCAS-CMS/cf-python/blob/defa44fff6d146da1703bf4133af56afff19a14f/cf/data/data.py#L104-L114

 Structurally, for modules with sizable lists of imports from a given module, notably ``numpy``(``.ma``), e.g in the ``data.data`` module:

https://github.com/NCAS-CMS/cf-python/blob/defa44fff6d146da1703bf4133af56afff19a14f/cf/data/data.py#L10-L84

would you be opposed to me changing it to the following, which is valid ``import`` syntax & which avoids the duplication of the import statement & makes it a bit easier to check the methods?:

```python
from numpy import (
    arange            as numpy_arange,
    arccos            as numpy_arccos,
    arccosh           as numpy_arccosh,
[...]

PR #223, opened 2021-06-09
Improve performance by allowing a saved `RegridOperator` instance to define the regrid operator

davidhassell:
Fixes: #222


davidhassell:
Hi Sadie - it'd be great if you could cast your eye over this.

The main elements are the change to the `cf.Field.regrid[sc]` APIs to allow a previously created `RegridOperator` (defined in `cf/regrid/regridoperator.py`) instance to define the regridding weights. By default the regridding weights are calculated afresh for every regrid.

I took this opportunity to move the regridding methods from the `Field` object to their own module (`cf.regrid.utils`), but that's just house keeping, so no real change, there.

Whilst I have done some work on the `regrid[sc]` docstrings, I am aware that they are fairly shabby in places. I think that this can be tackled separately - I'll raise a new issue for that.

Thanks,
David

sadielbartholomew:
Thanks David for the reviewing guidance which all seems very clear. I will review later today.

> I took this opportunity to move the regridding methods from the Field object to their own module (cf.regrid.utils)

Excellent idea. I'd also be keen to move out other related aspects if we can find any because the size of the cf `Field` module is enough that e.g. I have had to disable line numbering with emacs as it makes it to slow for practical development (though in fairness with the extensions I use anything above 5000 means it is best to turn them off)!

davidhassell:
Thanks, Sadie. I'm all for reducing the size of `field.py` in that way.

sadielbartholomew:
> I'm all for reducing the size of field.py in that way.

Excellent, I will look out for opportunities to do so.

I have just observed another indicator that the module is larger than it should be, in that GitHub gives up on syntax highlighting after about ~10,000 lines into a file (or a Python one at least)!

cf/regrid/utils.py
```diff
@@ -0,0 +1,1673 @@
...
+        dst_coord_order: `list`, optional
+            A list of lists specifying the ordering of the axes for
+            each 2-d destination coordinate.
+
+    :Returns:
+
+        `None`
+
+    """
+    # NOTE: May be common ground between cartesian and shperical that
```
sadielbartholomew, line 1047: (Not a typo introduced here, just moved I guess, not a functional one, but since I've seen it:)
```suggestion
    # NOTE: May be common ground between cartesian and spherical that
```

cf/test/test_regrid.py
```diff
@@ -167,6 +153,105 @@ def test_Field_regridc(self):
...
+            op = f1.regrids(
+                dst,
+                method="conservative",
+                dst_cyclic=True,
+                return_operator=True,
+            )
+            r = f1.regrids(op)
+            self.assertTrue(f3.equals(r))
+
+            # Regrid global to regional roated pole
```
sadielbartholomew, line 191:
```suggestion
            # Regrid global to regional rotated pole
```

cf/__init__.py
```diff
@@ -81,8 +81,8 @@
 """
 
 __Conventions__ = "CF-1.8"
-__date__ = "2021-05-25"
-__version__ = "3.9.0"
+__date__ = "2021-??-??"
+__version__ = "3.10.0+regrid"
```
sadielbartholomew, line 85: I had to remove the `+regrid` to `pip install - e.` to test, so assuming you will cut that off, all is good!

sadielbartholomew (approved):
Excellent, performance much improved as per the issue at hand and nothing has broken with the update and refactor as far as I can tell. I only have a few typos to point out (and they can be addressed at a future time if you wish).

cf/field.py
```diff
@@ -74,6 +72,37 @@
 
 from . import mixin
 
+from .regrid import (
```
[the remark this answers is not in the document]
davidhassell: Good idea. I'll leave this for now and we'll resolve in #63

davidhassell:
Thanks, Sadie. All points addressed as per your suggestions (apart from the alphabetical ordering of imports which is best left to robots rather than me!)

Issue #230, opened 2021-07-09
cf.aggregate fails when a datum or coordinate conversion parameter has an array value


davidhassell:
At v3.10.0 `cf.aggregate` fails when a datum or coordinate conversion parameter has an array value. 

In this case, the `cf.CorodinateReference.structural_signature()`  cantains a `f.Data` object with more than one value, which leads to problems when the structural signature is used (as part of) d dictionary key: 

```
ValueError: The truth value of Data with more than one element is ambiguous. Use d.any() or d.all()
```

A fix is to replace the `cf.Data` in the structural signature with a python scalar or tuple (of tuples) thereof.

Thanks to @ThibHlln for finding this bug.

PR #231, opened 2021-07-09
Fix for cf.aggregate failure when a datum or coordinate conversion parameter has an array value

davidhassell:
Fixes: #230

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,13 @@
+version 3.??.?
+--------------
+----
+
+**2021-??-??**
+
+* Fix for `cf.aggregate` failures when a datum or coordinate
+  conversion parameter has an array value
+  (https://github.com/NCAS-CMS/cf-python/issues/230)
+
 version 3.10.0
 --------------
 ----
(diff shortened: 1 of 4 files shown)


sadielbartholomew (commented):
This seems to work well though:

* we could do with a corresponding test, ideally bundled with this PR, to isolate the original aggregate failure (as also indicated by the decrease in coverage on the Codecov report);
* given that the solution relies on the recursive function `_totuple`, I wonder if there may be a negative effect on performance (it's difficult to check that without a test to expose the case at hand, though) do you have any ideas about the potential hit in speed and/or the specifics and nicheness of the case that might influence how consequential it could be to users?

davidhassell:
Thanks, Sadie.

I'm not too worried about performance. The recursive nature is "belt and braces" - there is no use case for multidimensional parameter values and none are specified by the conventions - so it shouldn't be a an issue. If it becomes so we can review it then.

Always a good idea to add a test! I'll do that ...

sadielbartholomew:
> The recursive nature is "belt and braces" - there is no use case for multidimensional parameter values and none are specified by the conventions - so it shouldn't be a an issue. If it becomes so we can review it then.

Fair enough, that's all good by me.

> Always a good idea to add a test! I'll do that ...

Excellent, thanks. Please merge away once the test is complete (or tag me to look at the update if you prefer, I am happy to do so).

davidhassell:
Test added, Pretty straight forward, so merging.

Issue #228, opened 2021-07-09
Improve verbose information output from cf.aggregate


davidhassell:
Some of the one-line output messages produced by `verbose` >= 2 in `cf.aggregate` are a little opaque, and can be improved. PR to follow.

PR #232, opened 2021-07-09
Improve verbose information output from cf.aggregate

davidhassell:
Fixes: #228

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,12 @@
+version 3.??.?
+--------------
+----
+
+**2021-??-??**
+
+* Improved verbose information output from `cf.aggregate`
+  (https://github.com/NCAS-CMS/cf-python/issues/228)
+
 version 3.10.0
 --------------
 ----
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -2708,9 +2708,7 @@ def _group_fields(meta, axis):
             # --------------------------------------------------------
             meta[
                 0
-            ].message = (
-                "No corresponding axes have different 1-d coordinate values."
-            )
+            ].message = "Some fields have identical sets of 1-d coordinates."
             return ()
 
         else:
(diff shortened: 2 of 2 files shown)


sadielbartholomew (approved):
The updated messages are clearer than the original ones in my eyes, excellent.

Though I could hack some fields to aggregate to run into the messages that have been changed here, to test, I didn't seem to hit them by ramping up the verbosity "to 11" (i.e. max debug/-1) on the aggregate test, so that suggests we can improve test coverage in future by including cases corresponding to attempted aggregation leading to these messages e.g. of fields where some have identical 1D coordinates, etc.

Also, during reviewing I noticed a few formatting errors in the higher-verbosity logging messages (probably mistakenly introduced by me during the original conversion from `print` statements), so I will put up a PR shortly in soon to fix those. I might also check all tests under debug verbosity to pick up on any obvious further log message issues towards that PR.

davidhassell:
Thanks, Sadie. I'll add try and a unit test for these messages to this PR. Good idea to have a separate PR for the other message you have spotted.

(Can we make `verbose=11` an alias for `verbose=-1`, just for fun? Seriously)

sadielbartholomew:
> (Can we make verbose=11 an alias for verbose=-1, just for fun? Seriously)

It could be a nice easter egg! Though I think we planned to have an even higher parallelisation-logic logging level? In which case that would be the highest of all and should be reserved as 11...

davidhassell:
Hi @sadielbartholomew  - may I merge this one, now? I think improving the test coverage as you suggest is one for another issue.

sadielbartholomew:
> Hi @sadielbartholomew - may I merge this one, now? I think improving the test coverage as you suggest is one for another issue.

There are a few conflicts to resolve, but once that is done, please go ahead and merge (generally, for future, if I give a formal 'approved' review I am happy for the PR to be merged unless there is something outstanding stated in the review comment, and I agree test coverage can be improved separately).

> so I will put up a PR shortly in soon to fix those

I didn't link in the relevant commit, like I should have ideally, but I did do this a while back, so that is sorted too.

davidhassell:
Thanks, Sadie

Issue #229, opened 2021-07-09
cf.aggregate fails when field units are non-valid


davidhassell:
At v3.10.0, if otherwise aggregatable fields have non-valid units (such as `kg -m-2 s-`, `cf.aggregate` fails, ultimately when `cf.Data.concatenate` refuses to aggregate data with non-equivalent units. Since cfunits v3.3.1, non-valid units are no longer equivalent. 

In this case, cf.aggregate needs to return, unaggregated, the fields with bad units, and report that it has done so.

PR #233, opened 2021-07-09
Fix for cf.aggregate failures when field units are non-valid

davidhassell:
Fixes: #229

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,13 @@
+version 3.??.?
+--------------
+----
+
+**2021-??-??**
+
+* Fix bug that causes a failure in `cf.aggregate` when otherwise
+  aggregatable fields have non-valid units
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
All good. As indicated for the desired solution in the corresponding issue:

> cf.aggregate needs to return, unaggregated, the fields with bad units, and report that it has done so

which has been achieved here, as tested by applying `override_units("mangoes", inplace=True)` on a field and observing that it reports `Unaggregatable 'eastward_wind' fields have been output: Non-valid units <Units: mangoes>` upon unsuccessful (without erroring) attempt to aggregate it with itself.

I have made one optional minor suggestion made not pertaining to the functionality, as below.

sadielbartholomew (commented):
Actually, ignore the conclusion from my previous review comment, sorry - I meant to say we could do with a test for the bad units case to check for the right result and log message! I suppose that could be added in a later PR, though.

PR #238, opened 2021-07-26
Apply & add hook for `isort` (systematic import sorting)

sadielbartholomew:
Closes NCAS-CMS/cf-python#63: no functional changes made, just automatic re-organisation of import statments with an associated pre-commit hook to make the changes in-place on attempted commit.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,8 +20,8 @@ jobs:
   # *The config. applied here is from our .pre-commit-config.yaml*
   #
   # Note the pre-commit linting includes all of the other linters configured
-  # to run as pre-commit hooks, namely black, flake8, docformatter and
-  # pydocstyle, so there is no need to set these up individually!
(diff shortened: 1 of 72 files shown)


sadielbartholomew:
OK, from the CI job failure it looks like I missed `scripts/cfa` for 264e12a so I will get that auto-formatted with `isort` in the next commit, and then the linting working should pass. It looks like we are also having some hanging tests still for the test suite run jobs, which I am looking into separately, so as long as one of those jobs passes I am satisfied this is good to merge (given the PR changes aren't functional ones).

sadielbartholomew:
Closing and reopening to re-trigger the CI jobs...

sadielbartholomew:
> It looks like we are also having some hanging tests still for the test suite run jobs, which I am looking into separately, so as long as one of those jobs passes I am satisfied this is good to merge (given the PR changes aren't functional ones).

OK, on that basis this is now good to merge. I will also cancel the hanging tests to save on limited Actions resource.

PR #244, opened 2021-08-13
`test_Data` workflow with skips to monitor Dask migration

sadielbartholomew:
I am about to put up some more interesting PRs to make headway with 'Daskifying' the data-related methods in turn for #182, but before I start doing that in earnest I want to ensure we can track the migration progress in terms of the tests that pass, notably for (self-)review to:

* indicate progress for a given PR by highlighting extra tests that pass due to a given change; and
* ensure tests that do pass due to progress previously made aren't broken by any proposed changes in a new PR.

So in this PR I have:

* added skips to tests in `test_Data`  that can be switched on or off by a new flag `TEST_DASKIFIED_ONLY`;
* added a new workflow specifically specifically for the LAMA-to-dask migration work to trigger Github Actions to run jobs that run the `test_Data` module and not the whole test suite, because the first goal is to get the data modules working, where metadata tests will fail on data issues until that is complete.

### DIFF
--- a/.github/workflows/dask-migration-testing.yml
+++ b/.github/workflows/dask-migration-testing.yml
@@ -1,0 +1,113 @@
+# A GitHub Action to run test_Data.py only for the 'lama-to-dask' branch
+name: Test `cf.Data` during the replacement of LAMA with Dask
+
+on:
+  push:
+    branches:
+      - lama-to-dask
+  pull_request:
+    types: [opened, reopened, ready_for_review]
+    branches:
+      - lama-to-dask
+
+jobs:
+  test-suite-job-0:
+
+    # Set-up the build matrix. We run on different distros and Python versions.
+    strategy:
+      matrix:
+        # Skip older ubuntu-16.04 & macos-10.15 to save usage resource
+        os: [ubuntu-latest, macos-latest]
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    # Run on new and old(er) versions of the distros we support (Linux, Mac OS)
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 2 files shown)


sadielbartholomew:
The new workflow clearly didn't trigger, I try to get it to start on Monday and go from there.

sadielbartholomew:
Using an open-close to see if that will result in a trigger of the new workflow...

sadielbartholomew:
Triggering worked :tada: With the new test skips, I want to get the jobs to pass before I can consider the new workflow to be ready, so let's see how they fare...

sadielbartholomew:
Most of the Ubuntu jobs have passed, largely due to the intentional skips on the `test_Data` they ultimately run:

```
...

Ran 73 tests in 0.221s

OK (skipped=66)

...
```

with the Mac OS jobs generally taking longer and being cancelled by the one failure on the Ubuntu with Python 3.6 job. I have just checked and Dask no longer supports Python 3.6, hence the lone test failure on `test_Data___setitem__` for that version only due to a `NotImplementedError: Item assignment with <class 'tuple'> not supported`. I will remove Python 3.6 from the job matrix in the next commit, and hopefully both the Ubuntu and Mac OS jobs will then all pass.

sadielbartholomew:
Re-triggering via open-close...

sadielbartholomew:
Sorry @davidhassell I initially added you to review and then realised there is nothing much to review here so removed you as a reviewer, but please do read the opening comment so you know about the skipping of some `test_Data` test methods which will be incrementally un-skipped as more progress is made, and the new workflow to run `test_Data` only for PRs to the `lama-to-dask` development branch. for #182

PR #246, opened 2021-08-20
Move deprecated data attrs & methods out into mixin

sadielbartholomew:
Towards reducing the size of the `data.data` module (currently ~15,000 lines), move out the deprecated and newly-deprecated (for the dask migration) attributes and methods of the `Data` class into a new module by creating a dedicated mixin class `DataClassDeprecationsMixin` to house them. (Also, in b18f653, address one minor warning that was being raised by the `test_Data` run.)

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -12779,7 +12779,7 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         i=False,
         _preserve_partitions=False,
     ):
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
@davidhassell I won't request a review on this one as it shouldn't be controversial (except perhaps for the location and name which can easily be changed later), but please note the opening comment for the gist of the change here.

sadielbartholomew:
To provide a breadcrumb trail: linked to #182.

PR #247, opened 2021-08-23
Migrate `Data.transpose` from LAMA to Dask

sadielbartholomew:
Daskify the `Data.transpose` method. This one is fairly trivial (I believe) since there is a Dask intrinsic method to  `transpose`.

Note I had to remove the loop and context management over `chunksize in self.chunk_sizes` for the corresponding test, but otherwise I have not touched that test, which now passes, as indicated by the removal of the `unittest.skipIf` decorator on the test which now passes locally and the CI will indicate if it passes across the job OS and version matrix.


cf/data/data.py
```diff
@@ -12142,21 +12142,14 @@ def transpose(self, axes=None, inplace=False, i=False):
...
-        data_axes = d._axes
-        d._axes = [data_axes[i] for i in iaxes]
-
-        # Permute the shape
-        shape = d._shape
-        d._shape = tuple([shape[i] for i in iaxes])
+        # TODODASK Q) is d._axes still relevant? Or can the following go?
+        # Test passes with it commented out, but it could still be important?
+        # data_axes = d._axes
+        # d._axes = [data_axes[i] for i in iaxes]
```
sadielbartholomew, line 12148: @davidhassell, I am unsure about the role of `_axes` in the Dask scheme you have initiated, so if don't mind it would be good to get some clarification.

For example, with this PR `test_Data_transpose` passes, but I am not sure whether we still need to include the two lines above to permute the `_axes` identifiers. Is `_axes` still important for this or other logic or can those lines now go? Thanks.

davidhassell: Hi Sadie, Still recovering from being on leave (in a good way) so my memory isn't what it could be today ... I _think_ that the `_axes` attribute will be redundant with dask, as it was only there to implement the lazy transpose on a LAMA partition matrix.

That said, I think it's also used elsewhere, e.g. for registering axis cyclicity. I suspect that this might mean that we can't ditch `_axes` altogether, but can from the transpose methods (and possibly other methods).

I'll review its role.

sadielbartholomew: Thanks for the info, David. I hope you enjoyed your leave! The `_axes` question isn't a blocker for this PR or the other ones I am shortly about to put up or indeed work on, so take your time to consider it. For now, assuming you don't realise quickly, I think the best thing is to leave in this `TODODASK` comment as a reminder to review it.

sadielbartholomew:
Of relevance to this PR and to keep an eye on going forward: https://github.com/data-apis/array-api/issues/228

PR #248, opened 2021-08-23
Remove loops over chunksize in `test_Data`

sadielbartholomew:
Towards #182, remove the loop over, and context management of, chunk sizes in `test_Data`.

Chunk size is a LAMA-related aspect and we can assume (at least until the Data module migration is more complete and we can think about how rigorous this test module needs to be and how to improve it) that the data array memory management is tested thoroughly by Dask so that we don't need to effectively test over a range of sizes of partitions of the sub-arrays.

As a result, three extra tests pass (after being un-skipped) relative to those passing before, since the non-existent attribute of `chunk_sizes` being called in the test turned out to be the only migration issue left in those small number of cases.

### DIFF
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -22,8 +22,16 @@ faulthandler.enable()  # to debug seg faults and timeouts
 
 import cf
 
-# TODODASK: these can be moved into the lone tests that use them now
+# Variables for _collapse
 a = np.arange(-100, 200.0, dtype=float).reshape(3, 4, 5, 5)
+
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Opening and closing to re-trigger the CI after I fixed a small issue, squashed the commit down into the previous related one, and force-pushed the branch...

sadielbartholomew:
No peer-review necessary since this is a setup/admin PR rather than a method migration and the general principle of removing the explicit chunksize handling in the test was agreed with David in a chat today. CI jobs all pass, so this is good to merge.

PR #249, opened 2021-08-24
Restrict size 1 limitation to source field in regridding

### LINKED ISSUE #250: Restrict size 1 dimension to source field in regridding
Hi David, Hi Sadie,

In the regridding functionality, there currently is a restriction on the size of the dimensions of the grid: any dimension needs to be greater than one. 

But after some testing, I found out that it is only problematic to have a size 1 dimension on the source grid, but not on the destination grid. Indeed, ESMF only raises an error when the source grid has a size 1 dimension (see https://github.com/esmf-org/esmf/blob/b4c535d3b3925f43ccad78191f40d6c51f6f2e7e/src/Infrastructure/Field/src/ESMF_FieldRegrid.F90#L2840-L2847). I am not sure what is the underlying reason for this restriction in ESMF, and it is surprising that it complains only for the source grid.

Nevertheless, as a user, I see some situations where I could benefit from the ability to regrid to a destination resolution composed of a single grid cell (e.g. pre-process some gridded data to get a lumped value for a given region). Since it seems to be working for all regridding methods, I thought it may be good to lift the restriction on the destination grid in `cf-python`.

What do you think? 
Also, do you have any idea why ESMF would restrict the size 1 dimension on the source grid only?

ThibHlln:
resolves #250 

Hi David, Hi Sadie,

With the simple change contained in this PR, the code below would work, which is not currently the case (i.e. a regridding towards a single grid cell domain).

```python
import cf
import numpy

# source field (2x3 grid)
fld1 = cf.Field()
fld1.set_construct(
    cf.DimensionCoordinate(
        properties={'standard_name': 'longitude',
                    'units': 'degrees_east',
                    'axis': 'X'},
        data=cf.Data([5, 15, 25]),
        bounds=cf.Bounds(data=cf.Data([[0, 10], [10, 20], [20, 30]]))
    ),
    axes=fld1.set_construct(cf.DomainAxis(size=3))
)
fld1.set_construct(
    cf.DimensionCoordinate(
        properties={'standard_name': 'latitude',
                    'units': 'degrees_north',
                    'axis': 'Y'},
        data=cf.Data([55, 65]),
        bounds=cf.Bounds(data=cf.Data([[50, 60], [60, 70]]))
    ),
    axes=fld1.set_construct(cf.DomainAxis(size=2))
)

# destination field (1x1 grid)
fld2 = cf.Field()
fld2.set_construct(
    cf.DimensionCoordinate(
        properties={'standard_name': 'longitude',
                    'units': 'degrees_east',
                    'axis': 'X'},
        data=cf.Data([15]),
        bounds=cf.Bounds(data=cf.Data([[0, 30]]))
    ),
    axes=fld2.set_construct(cf.DomainAxis(size=1))
)
fld2.set_construct(
    cf.DimensionCoordinate(
        properties={'standard_name': 'latitude',
                    'units': 'degrees_north',
                    'axis': 'Y'},
        data=cf.Data([60]),
        bounds=cf.Bounds(data=cf.Data([[50, 70]]))
    ),
    axes=fld2.set_construct(cf.DomainAxis(size=1))
)

# set some data
fld1.set_data(numpy.arange(6).reshape(2, 3), axes=('Y', 'X'))
fld2.set_data(numpy.ones((1, 1)), axes=('Y', 'X'))

# test regridding on all existing methods
for m in ['linear', 'conservative', 'conservative_2nd', 
          'patch', 'nearest_stod', 'nearest_dtos']:
    
    fld3 = fld1.regrids(fld2, method=m)
    
    print(fld3.data)
```

### DIFF
--- a/cf/regrid/utils.py
+++ b/cf/regrid/utils.py
@@ -71,7 +71,7 @@ def regrid_compute_mass_grid(
     return mass
 
 
-def regrid_get_latlon(f, name, axes=None):
+def regrid_get_latlon(f, name, method, axes=None):
     """Retrieve the latitude and longitude coordinates of this field and
     associated information. If 1-d lat/long coordinates are found then
     these are returned. Otherwise, 2-d lat/long coordinates are searched
@@ -85,6 +85,9 @@ def regrid_get_latlon(f, name, axes=None):
         name: `str`
             A name to identify the field in error messages. Either
             ``'source'`` or ``'destination'``.
+
+        method: `str`
+            The regridding method.
 
         axes: `dict`, optional
             A dictionary specifying the X and Y axes, with keys
@@ -259,13 +262,15 @@ def regrid_get_latlon(f, name, axes=None):
     else:
         coords_2D = False
 
-        # Check for size 1 latitude or longitude dimensions if source grid
-        # (a size 1 dimension is only problematic for the source grid in ESMF)
-        if name == "source":
+    # Check for size 1 latitude or longitude dimensions if source grid
+    # (a size 1 dimension is only problematic for the source grid in ESMF)
+    if name == "source":
+        if method in ("linear", "bilinear", "patch"):
             if x_size == 1 or y_size == 1:
                 raise ValueError(
                     "Neither the longitude nor latitude dimension coordinates "
-                    f"of the {name} field can be of size 1."
+                    f"of the {name} field can be of size 1 for {method} "
+                    f"regridding."
                 )
 
     coord_keys = [x_key, y_key]
--- a/cf/regrid/utils.py
+++ b/cf/regrid/utils.py
@@ -72,15 +72,18 @@ def regrid_compute_mass_grid(
 
 
 def regrid_get_latlon(f, name, method, axes=None):
-    """Retrieve the latitude and longitude coordinates of this field and
-    associated information. If 1-d lat/long coordinates are found then
-    these are returned. Otherwise, 2-d lat/long coordinates are searched
-    for and if found returned.
+    """Get latitude and longitude coordinate information.
+
+    Retrieve the latitude and longitude coordinates of a field, as
+    well as some associated information. If 1-d lat/lon coordinates
+    are found then these are returned. Otherwise if 2-d lat/lon
+    coordinates found then these are returned.
 
     :Parameters:
 
         f: `Field`
-            The source or destination field.
+            The source or destination field from which to get the
+            information.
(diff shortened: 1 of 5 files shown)


davidhassell:
Hi Thibault,

Would you mind copying the above into an issue and referencing that issue from this PR? (https://ncas-cms.github.io/cf-python/contributing.html#preparing-pull-requests) - I'll reply over there ...

Thanks, David

cf/regrid/utils.py
```diff
@@ -258,12 +258,15 @@ def regrid_get_latlon(f, name, axes=None):
...
-            )
+
+        # Check for size 1 latitude or longitude dimensions if source grid
+        # (a size 1 dimension is only problematic for the source grid in ESMF)
+        if name == "source":
+            if x_size == 1 or y_size == 1:
+                raise ValueError(
+                    "Neither the longitude nor latitude dimension coordinates "
+                    f"of the {name} field can be of size 1."
+                )
```
davidhassell, line 269: I'm wondering why this highlighted block of code is inside the "else" clause, and is not indented one tab to the left ... This has always been the case, but looking at it afresh it'd be nice to know why the check is only done when `coords_2D = False`, and if that's correct.

davidhassell:
Hi @ThibHlln, I wasn't sure that anything would happen here, but I've managed to push to this PR of yours. This commit  (https://github.com/NCAS-CMS/cf-python/pull/249/commits/7dfb455d3f973d721560ddaee962007cdb1a1d4c) tidies up the removal of the "squeeze", which needs to go back in sometimes to get rid of size 1 dimensions that are not being regridded. Hope that's OK! (this commit can always be nuked, of course).

cf/regrid/utils.py
davidhassell: This is good now, I think

davidhassell:
Hi @ThibHlln  - I think we can merge this, now. I'll await your OK, though ...

ThibHlln:
Hi @davidhassell,

I was not sure what the squeezing was for, so thank you for fixing this.

Yes, I think it can be merged now, I ran the tests above successfully after your commits. 
So, it looks all good to me. :rocket:

davidhassell:
Hi @ThibHlln  - I think the squeezing is all about making sure we pass the correctly shaped numpy array down to the ESMF layer. Thanks for your help with this. We expect a release of 3.11.0 in a couple of weeks (ish). Merging now ...

PR #254, opened 2021-08-27
`Data.equals`: add unit test & migrate to Dask

### LINKED ISSUE #251: Add unit test for `Data.equals` method
Particularly to confirm the validity of the migration from LAMA to Dask in #182, we should add a new test method to explicitly test the `Data.equals` method as it does not seem to be unit tested yet, despite the other tests relying on it for many test assertions.

This would be a test akin to `test_Field_equals` and similar test methods as existing for `CellMethod` and `CoordinateReference` currently, so the name `test_Data_equals` would be consistent with those.

sadielbartholomew:
Fix #251 & address the `Data.equals` method for #182.


cf/test/test_Data.py
```diff
@@ -131,6 +131,149 @@ class DataTest(unittest.TestCase):
...
+                        "Data: Different data types: float32 != int64"
+                        in log_msg
+                        for log_msg in catch.output
+                    ),
+                    expect_to_see_msg,
+                )
+
+        # Test ignore_data_type parameter
+        # TODODASK - this one needs documenting in the method docstring.
+        # Question for DH: is ignore_type an alias?
```
sadielbartholomew, line 274: Hi @davidhassell, when you review this,  it would be good to get your advice regarding deprecated and potentially-deprecated arguments and how they should be managed, perhaps in a more general sense because a lot of them will apply across multiple methods.

Specifically for the `equals` method, the following are not documented:

* `ignore_data_type`;
* `ignore_type`;
*`ignore_compression`;

and I am not sure of the role of `ignore_compression` is not used in the method, suggesting it can be dropped?

Also, generally for this probable-`4.0.0` version, can we drop (some or all) long-deprecated methods like `traceback` completely or do we still want to handle these with the `_DEPRECATION*` functions, etc.?

davidhassell: Hi Sadie, 

The undocumented keywords can easily be documented with docstring rewriter keys defined in cfdm, as is done in  `cfdm.Data.equals`.

`ignore_compression` is interesting, because it _is_ used in `cfdm.Data.equals` (which is a super method of `cf.Data,.equals`) so we need to review that separately 

I think you're right - at 4.0.0 we can drop all of the keywords that were deprecated at 3.0.0, and possibly some that were deprecated at later 3.x versions.

Thanks,
David

cf/data/data.py
```diff
@@ -8811,32 +8812,27 @@ def equals(
...
-
-            if not _numpy_allclose(
-                array0, array1, rtol=float(rtol), atol=float(atol)
-            ):
-                logger.info(
-                    "{0}: Different array values (atol={1}, "
-                    "rtol={2})".format(self.__class__.__name__, atol, rtol)
-                )
+        # Finally check that corresponding elements are equal to a tolerance
+        if not da.allclose(
```
davidhassell, line 8827: Not sure if `da.allcose` is enough here, e.g. the following gives `True` in dask (and numpy), but we want it to be `False`:
```python
>>> from dask import array as da
>>> import numpy as np
>>> dx = da.from_array(np.ma.array([1, 999, 3], mask=[0, 1, 0]))
>>> dy = da.from_array(np.ma.array([999, 2, 3], mask=[1, 0, 0]))
>>> da.allclose(dx, dy).compute()
True
```
This is why we use the (terribly named) `cf.functions._numpy_allclose` function instead, which also accounts for non-numeric data types.

Daskifying  `cf.functions._numpy_allclose` will need a little thought to see if we can avoid computing the two operands multiple times each (once to get the mask, and once to compare the data).

cf/data/data.py
```diff
@@ -8811,32 +8812,27 @@ def equals(
...
+            self_dx, other_dx, rtol=float(rtol), atol=float(atol)
+        ):
```
davidhassell, line 8829: (we would need a `.compute()` on the `da.allclose` call, but see previous comment!)

davidhassell (changes requested):

cf/test/test_Data.py
sadielbartholomew: Thanks for clarifying, @davidhassell, and for your review.

> The undocumented keywords can easily be documented with docstring rewriter keys defined in cfdm, as is done in cfdm.Data.equals.

Noted, thanks. I'll go in and do that where appropriate going forwards.

> ignore_compression is interesting, because it is used in cfdm.Data.equals (which is a super method of cf.Data,.equals) so we need to review that separately

Sure, I'll put in a specific `TODODASK` for that.

> I think you're right - at 4.0.0 we can drop all of the keywords that were deprecated at 3.0.0

That makes sense.

> and possibly some that were deprecated at later 3.x versions.

 I'll initiate a list covering such keywords in the main #182 thread, and add to it as I encounter them. We can then discuss them when there's a good time to to agree which to fully deprecate.

cf/data/data.py
```diff
@@ -8811,32 +8812,27 @@ def equals(
         other_Units = other.Units
         if self_Units != other_Units:
             logger.info(
-                "{}: Different Units ({!r}, {!r}".format(
+                "{}: Different Units ({!r}, {!r})".format(
```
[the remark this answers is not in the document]
sadielbartholomew: I was planning to do the f-string conversions separately, as I am doing with the "Incremental conversion to f-strings: ..." daily commits on the `master` branch, to keep the functional changes separate as I thought it might make reviewing of the migration proper somewhat easier, but if you prefer those to be bundled in with each migration PR, that's fine by me. I guess if I do them as a separate commit for each PR, that gives us the best of both worlds, so that'll be my plan going forwards.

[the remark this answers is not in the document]
davidhassell: OK, that is good - Feel free to adopt whatever approach to f-strings you like!

cf/data/data.py
sadielbartholomew: Ah, OK, good points. It sounds like, firstly, my unit test needs to be more thorough so it tests for cases like that, so that currently with the change set here it will fail, then I need to fix the `equals` in the way you state is required.

> Daskifying cf.functions._numpy_allclose will need a little thought to see if we can avoid computing the two operands multiple times each (once to get the mask, and once to compare the data).

I'll take a further look and have a think about a good approach, though I think we discussed a similar challenge back in March or around then, so that might guide us as to a good solution. I'll look through my notes to see what we thought back then and report back briefly and/or add a new commit with a suggestion.

sadielbartholomew: > This is why we use the (terribly named) cf.functions._numpy_allclose function instead

I realised now (whilst having a more detailed look) that I assumed `_numpy_allclose` was an alias to `numpy.allclose` and hence the direct and seemingly obvious conversion I suggested to `da.allclose`. (So yes, it is somewhat more complicated than I thought to migrate this.) I think, even though those `_numpy_*` aliases have been removed for the `lama-to-dask` branch, it would be best to rename `_numpy_allclose` since the name is rather confusing.

I'll get my migration fix suggest committed and in the meantime if you have any thoughts about a less confusing name, do let me know. It does seem difficult to name, though. Maybe `_inclusive_allclose` or something similar? (It isn't too important since it is an internal function.)

sadielbartholomew:
Hi @davidhassell, RE https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r699977411 and the (intermediate, towards this PR) daskification of `_numpy_allclose` (which we should rename I think as per our comments in that thread), I have been investigating approaches and the following is a summary of my thoughts so far.

## On `_numpy_allclose`

### Currently (pre-migration):
1. Checks if a and b are `MaskedArray` objects.
2. If neither is `MaskedArray`: `np.allclose(a, b)` -> done.
3. Else if both are a `MaskedArray`: compare masks, if different -> `False`, done.
4. Otherwise, if only one is a `MaskedArray`: check if either has a non-trivial (`False`) mask and if so, they have different masks, so `False` -> done.
5. Otherwise, at least one is a `MaskedArray` but the masks are the same, so need to run `np.ma.allclose` to compare data. -> done.

### To migrate:

I am not yet sure if this suggestion is the best approach, particularly as I am just getting to grips with all of the possibilities provided with Dask and which are more efficient to use and advisable to use in what cases, but it is at least a starting point...

1. Check the type of the underlying array on a and b (this is `type(dx._meta)`, as reported in the `repr` under the name `chunktype`) to determine if either are `numpy.MaskedArray` rather than `numpy.ndarray` (those are the only two options, here, right, since it the method is only applied for `numpy` arrays I guess?).
2. If one is `MaskedArray` and the other `ndarray`, is `False` -> done.
3. If both are `numpy.ndarray`, just use `dask.allclose` -> done.
4. If both are `MaskedArray`, it's a bit more complicated, but doing something along the lines of the following does seem to work and the `visualize` graph (included below) seems promising, though of course at the `equal` and `isclose` step we have to wait on both possible parallel task routes which is not ideal:

```python
In [1]: import dask.array as da

In [2]: import numpy as np

In [3]: x = np.ma.array([1, 2, 3], mask=[0, 1, 0])

In [4]: y = np.ma.array([999, 2, 3], mask=[1, 0, 0])

In [5]: dx = da.from_array(x)

In [6]: dy = da.from_array(y)

In [7]: dx_mask = da.ma.getmaskarray(dx)

In [8]: dy_mask = da.ma.getmaskarray(dy)

In [9]: dx_data = da.ma.getdata(dx)

In [10]: dy_data = da.ma.getdata(dy)

In [11]: mask_comparison = da.equal(dx_mask, dy_mask)

In [13]: data_comparison = da.isclose(dx_data, dy_data)

In [14]: result = da.all(da.logical_and(mask_comparison, data_comparison))

In [15]: result.visualize(filename="test007.png")
Out[15]: <IPython.core.display.Image object>

In [16]: result.compute()
Out[16]: False

In [17]: def compare(f, g):
    ...:     mask_comparison = da.equal(da.ma.getmaskarray(f), da.ma.getmaskarray(g))
    ...:     data_comparison = da.isclose(da.ma.getdata(f), da.ma.getdata(g))
    ...:     result = da.all(da.logical_and(mask_comparison, data_comparison))
    ...:     return result
    ...: 

In [19]: compare(np.ma.array([1, 2, 3], mask=[0, 1, 0]), np.ma.array([1, 2, 3],
    ...: mask=[0, 1, 0])).compute()
Out[19]: True

In [20]: compare(np.ma.array([1, 2, 3], mask=[0, 1, 0]), np.ma.array([1, 2, 3],
    ...: mask=[0, 1, 1])).compute()
Out[20]: False

In [21]: compare(np.ma.array([1, 2, 3], mask=[0, 1, 0]), np.ma.array([5, 2, 3],
    ...: mask=[0, 1, 0])).compute()
Out[21]: False

In [22]: compare(np.ma.array([1, 2, 3]), np.ma.array([5, 2, 3])).compute()
Out[22]: False

In [23]: compare(np.ma.array([1, 2, 3]), np.ma.array([1, 2, 3])).compute()
Out[23]: True

```

As the outputs indicate, the process given in (4) could also apply for the general case, i.e. of any combo. of `ndarray` or `MaskedArray`, but given the need to process both a mask and a data array it won't be as efficient even under a `dask` task workflow, surely.

### Task graph from the above process for step (4)

test007

sadielbartholomew:
(FYI @davidhassell, as you may have seen from notifications, I have pushed some new commits here, but please don't look at this until I tag you again to let you know it is ready - I have prepared review aids that I will share but I am thinking over some aspects before I finalise everything and open this for review.)

sadielbartholomew:
Hi again @davidhassell, all ready now, though depending on your thoughts regarding the approach to take with regards to laziness of `cf.Data` (see the questions and comments I have added to our developer notes in 7a01cbe0f926f675713efc54eb763cbe29e4c15f) then I may want to revert the penultimate commit which converts the overall approach to Case 1 (as covered in those notes) whereas previously I was abiding by Case 2.

Please also see below for notes conveying the context of the PR and our discussions, since they took place a while back now so it is good to note down our thoughts.

Note also I will push up a commit to add a test for the new `_da_ma_allclose` internal method in a moment, but otherwise I am done here bar any changes relating to the discussion of approach to laziness, as above.

********************************************

#### Notes for context

After our September discussion we realised we should amend the task graph from that suggested in https://github.com/NCAS-CMS/cf-python/pull/254#issuecomment-912230797 to the following, where I have drawn a red box to outline the key part and that which is different to the previous graph:



and I have implemented that. The overall logic corresponds to this interactive example, in case you wished to play around interactively with it:

```pycon
In [1]: import dask.array as da
   ...: import numpy as np
   ...: x = np.ma.array([1, 2, 3], mask=[1, 0, 0])
   ...: y = np.ma.array([999, 2, 3], mask=[1, 0, 0])
   ...: dx = da.from_array(x)
   ...: dy = da.from_array(y)

In [2]: dx_mask = da.ma.getmaskarray(dx)
   ...: dy_mask = da.ma.getmaskarray(dy)
   ...: mask_comparison = da.equal(dx_mask, dy_mask)

In [3]: def da_ma_allclose(x, y, masked_equal=True, rtol=1e-05, atol=1e-08):
   ...:     x = da.asanyarray(x)
   ...:     y = da.asanyarray(y)
   ...:     return da.map_blocks(
   ...:         np.ma.allclose, x, y, masked_equal=masked_equal, rtol=rtol,
   ...:         atol=atol,
   ...:     )
   ...: 

In [4]: data_comparison = da_ma_allclose(dx, dy, masked_equal=False)

In [5]: result = da.all(da.logical_and(mask_comparison, data_comparison))

In [6]: result.visualize(filename="cf_equals_test_003.png")
Out[6]: <IPython.core.display.Image object>

In [7]: result.compute()
Out[7]: False
```

Notice the graph from the above is as desired:

cf_equals_test_003

davidhassell:
Hi Sadie - I think the logic is sound, but unfortunately it doesn't work for me if `dx` and `dy` have different  chunk sizes

```python
dx = da.from_array(x, chunks=(1, 2))
dy = da.from_array(y, chunks=(3,))
```

which is a limitation/feature of `map_blocks`, I think.

After some messing about I came up with:

```python
import dask.array as da
import numpy as np

def allclose(a_blocks, b_blocks, rtol=1e-05, atol=1e-08):
    result = True
    for a, b in zip(a_blocks, b_blocks):
        result &= np.ma.allclose(a, b, rtol=rtol, atol=atol, masked_equal=True)

    return result

# -----------------------------------------
# Case 1
# -----------------------------------------
x = np.ma.array([1,   2, 3], mask=[1, 0, 0])
y = np.ma.array([999, 2, 3], mask=[1, 0, 0])
dx = da.from_array(x, chunks=(1, 2))
dy = da.from_array(y, chunks=(3,))

dx_mask = da.ma.getmaskarray(dx)
dy_mask = da.ma.getmaskarray(dy)
mask_comparison = da.allclose(dx_mask, dy_mask)

axes = tuple(range(dx.ndim))

data_comparison = da.blockwise(
    allclose, '', dx, axes, dy, axes, dtype=bool,
    rtol=1e-05, atol=1e-08
)

result = mask_comparison & data_comparison

print('CASE 1')
print('mask equal?', mask_comparison.compute())
print('data equal?', data_comparison.compute())
print('     equal?', result.compute())

# -----------------------------------------
# Case 2
# -----------------------------------------
x = np.ma.array([1,   999, 3], mask=[0, 1, 0])   # Different
y = np.ma.array([999,   2, 3], mask=[1, 0, 0])
dx = da.from_array(x, chunks=(1, 2))
dy = da.from_array(y, chunks=(3,))

dx_mask = da.ma.getmaskarray(dx)
dy_mask = da.ma.getmaskarray(dy)
mask_comparison = da.allclose(dx_mask, dy_mask)

axes = tuple(range(dx.ndim))

data_comparison = da.blockwise(
    allclose, '', dx, axes, dy, axes, dtype=bool,
    rtol=1e-05, atol=1e-08
)

result = mask_comparison & data_comparison

print('\nCASE 2')
print('mask equal?', mask_comparison.compute())
print('data equal?', data_comparison.compute())
print('     equal?', result.compute())
```

which produces, as required:

```python
CASE 1
mask equal? True
data equal? True
     equal? True

CASE 2
mask equal? False
data equal? True
     equal? False
```

This assume that `dx` and `dy` have the same shape, but we'll already have checked that and short circuited if they haven't.

What do you think?

sadielbartholomew:
HI @davidhassell, thanks for your detailed (re-)review, as ever. Good spot regarding the chunk sizes - I hoped Dask would handle all of that, but I guess with our custom Dask method we need to be careful. I will add in a test for the chunk sizes akin to your examples, and a proper unit test for the new `allclose` method, first of all. Then try out your suggested solution - thanks for investigating and sharing it.

I'll start that all after my dinner. I am also about to push up further developer notes as quite a lot of other important aspects were discussed in relation to this PR!

davidhassell:
I wonder how the `allclose` thing actually works - It'd be nice to see a graph visualisation of this.

davidhassell:
Here's the dask graph for the non-matching chunks of case 2.

It looks OK to me. The chunk structures of `dx` and `dy` are made the same, and then a `list` of its chunks (np arrays) is passed to the `allclose` function.  I'm not wholly sure how you're meant to know that such a function takes such an input,but I guess that's what they mean by "Function to apply to individual tuples of blocks".

mydask

sadielbartholomew:
> Here's the dask graph for the non-matching chunks of case 2.

I imagine you've found, as I have, that we need to choose the non-trivial case to visualise and test for one because the graph engine seems to simplify the case where two identical arrays are input and show them as one! Though I guess that's because they are known (to be the same) at the start, whereas we of course need to look at the general case where there may be many arbitrary operation steps leading up to the start of the task graph in question...

***

I will push some progress on testing shortly, but just to clarify why I chose initially to go with `da.map_blocks(np.ma.allclose, x, y,...` for our own `_da_ma_allclose` implementation: I was looking at [the Dask masked array module, `dask/dask/array/ma.py`](https://github.com/dask/dask/blob/main/dask/array/ma.py), and saw that they mostly used that approach (e.g. [in these methods](https://github.com/dask/dask/blob/84da277b9739beae66380d2c0f53753487720656/dask/array/ma.py#L98-L113)).

Notably, though, I did notice they only defined across the module `da.ma` methods that involved a single masked array input, not more than one, and after your comments I think that is probably because of the complication of two inputs having possibly different chunk sizes.

Maybe we could open up an issue on Dask to ask about this, and go on to ask about adding in further masked methods, such as `da.ma.allclose` ourselves, as I believe is the plan anyway? (I did check the Dask issue tracker for any issues and notes on the topic, but couldn't find much we don't know already.)

Some comments regarding your suggestion for a solution above:

1. You are applying `allclose` for the mask, but I believe we agreed that `equal` is sufficient and indeed more efficient for the mask, since it will be Boolean so precision isn't relevant?
2. You may have used this one for inspiration and therefore be aware anyway, but looking at the docs, but the example [here](https://docs.dask.org/en/stable/generated/dask.array.blockwise.html#dask.array.blockwise) after "Inner product multiplying a by b, two 1-d vectors" (can't link to it directly sadly) corresponds, so I am even more confident your approach to the data comparison is both sound and optimal. :+1:

sadielbartholomew:
@davidhassell (note I've commented above also) I have just seen your latest comment regarding the new notes (sorry I missed it before).

> I think the .array attribute could be thought of as cf-python's version of .compute(). For methods that return new cf.Data objects (such as transpose) I think they should be lazy.Other methods, perhaps not. __repr__ and equals come to mind.

Ah, I like that solution very much! When I was considering the possibility of the `precompute` kwarg I did come to conclude in my mind that the best default would depend on case, and methods returning Booleans such as `equals` would probably be used as final, terminating operations so it would be best to pre-compute there. And I think your idea encapsulates the division of methods that are likely to be endpoints of any task graph rather than somewhere in the middle.

Excellent - that's decided I think! I'll add a brief note to summarise the plan there. For the lazy methods, we still have the complications of delaying error handling, and so on, to contemplate, but that doesn't stop me forging ahead with the coding for now...

sadielbartholomew:
FYI also David I am suggesting to create a new test module, `test_Data_utils.py`, with a test class `DataUtilsTest` to test the data utility functions we think are important to cover (probably all of them, eventually), as per the latest commit.

Let me know if you think that is OK as a location for such testing - I am happy to take a different approach to that, but as well as with the motivation of modularity, the `test_Data` module is already pretty long at ~3500 lines long at the moment so I would prefer to use a separate module for tests that aren't on the user-facing Data methods.

sadielbartholomew:
[This proposal](https://github.com/data-apis/array-api/issues/170) over on the Python Array API Standard is interesting RE `isclose` methods, as covered in our discussions today @davidhassell!

cf/test/test_Data.py
```diff
@@ -131,6 +131,293 @@ class DataTest(unittest.TestCase):
...
+        self.assertTrue(j6.equals(j6))
+        with self.assertLogs(level=cf.log_level().value) as catch:
+            print("J5 EQUALS J6 NOW <<<<<<<<<<<<<<<<<<<<<<<<<")
+            self.assertFalse(j5.equals(j6))
+            self.assertTrue(
+                any(
+                    "Data: Different array values" in log_msg
+                    for log_msg in catch.output
+                )
+            )
```
sadielbartholomew, line 280: #### Outstanding issue 1

Currently, I can't get all of the masked array tests to pass (for known reasons but your advice would be appreciated). Either the 'extra special case' here (`self.assertFalse(j5.equals(j6))` which tests that key case which makes our equality with masked arrays differ in behaviour to `np.ma.allclose` for example) fails, when `masked_equal=True` is input to the `_da_ma_allclose`, or all of the self-equality checks on masked Data arrays fail, e.g. `self.assertFalse(j4.equals(j4))`, when we instead set `masked_equal=False`.

I wasn't sure what the appropriate choice for `masked_equal` would be, but I don't think it is the issue at hand anyway, because upon investigation I found that, with the print statement in the final debugging commit, the underlying issue is that the data comparison is comparing the data *with masked elements* (e.g. `[-- 2.0 3.0]` for the above example), and not the data extracted separately from the mask (e.g. `[1.0 2.0 3.0]`).

Surely we want to test on the latter, so need to call some method or attribute to extract the data from the mask for our `data_comparison` result? But how would we do that - I've tried using the `getdata` method but that still masks elements where the mask is `True` for the result...

As well as that seeming wrong to me, there is the issue of self-equality, for example with `masked_equal=False`, my print statements show:

```
SELF-EQUALS J1 NOW <<<<<<<<<<<<<<<<<<<<<<<<<
DATA COMP'ING IS [-- 2.0 3.0] [-- 2.0 3.0]
DATA COMP RESULT IS False
MASK COMP RESULT IS True
OVERALL COMP RESULT IS False
```
indicating that the data comparison does not xonsider `[-- 2.0 3.0]` allclose to itself due to the masked element.

What are your thoughts on this? (Ca take a video call if that seems easiest for discussion.)

cf/test/test_Data.py
```diff
@@ -131,6 +131,293 @@ class DataTest(unittest.TestCase):
...
+
+        # Test NaN and inf values
+        d3 = cf.Data(a.astype(np.float64), "m")
+        h = cf.Data(np.full(shape, np.nan), "m")
+        # TODODASK: is this OK given that NaN in NumPy aren't equal to e/o
+        # or should this be assertTrue to expect equality behaviour?
+        self.assertFalse(h.equals(h))
```
sadielbartholomew, line 193: #### Outstanding issue 2

Notice that this is the lone self-equality which is `assertFalse` rather than `assertTrue`, and (as with everything except those assertions discussed in my other in-line comment for issue 1) it passes at the moment, because it is an array with NaN elements and naturally in NumPy NaN is not equal to itself.

Any cf.Data object not being equal to itself seems either wrong or potentially confusing to me. Do you think we should keep this behaviour, or change it so that NaNs are considered equal (eay enough to do with the `equal_nan` kwarg to `allclose`, etc.)?

I haven't converged on a strong preference either way, though I do think if we keep the behaviour of the self-equality being False we should at least provide a warning, e.g. via the logging: something like "The data array contains NaN elements which are not considered equal to themselves". And, of course, there's the option of providing our own `equal_nan` kwarg for flexibility for users to choose?

sadielbartholomew (commented):
Hi again @davidhassell, happy to report that this is nearly ready for re-review, with everything covered and decided on in our latest discussion addressed, I believe, in the set of commits starting at 2b73fad (see the commit messages for summaries of those), though note the final commit is a temporary one for debugging to help me to demonstrate the first outstanding issue as detailed in-line.

'Nearly ready' accounts for the fact there are two questions I wanted to ask you about. In general these relate to the fact that I added checks for self-equality i.e. `assertTrue(<cf.Data object>.equals(<same cf.Data object>))` (because those seem like an important edge-like case) and the results were quite interesting. See the in-line comments regarding those.

cf/test/test_Data.py
```diff
@@ -131,6 +131,293 @@ class DataTest(unittest.TestCase):
...
 
+    def test_Data_equals(self):
+        if self.test_only and inspect.stack()[0][3] not in self.test_only:
+            return
+
+        shape = 3, 4
+        a = np.arange(12).reshape(*shape)
+
+        d = cf.Data(a, "m")
+        self.assertTrue(d.equals(d))  # also do self-equality checks!
```
davidhassell, line 142: In general , should do self equality with `d` and `d.copy()`, as the test short circuits if `self is other`: https://github.com/NCAS-CMS/cfdm/blob/master/cfdm/mixin/container.py#L213
```suggestion
        self.assertTrue(d.equals(d.copy()))  # also do self-equality checks!
```

cf/test/test_Data.py
davidhassell: I'll need to think about this, but will do after submitting the other review comments

cf/test/test_Data.py
```diff
@@ -131,6 +131,293 @@ class DataTest(unittest.TestCase):
...
+        # Note these should *not* be considered equal inside cf.Data, whereas
+        # np.ma.allclose and indeed our own _da_ma_allclose methods do hold
+        # these to be 'allclose': Data.equals is stricter than _da_ma_allclose.
+        j5 = cf.Data(np.ma.array([1.0, 2.0, 3.0], mask=[1, 0, 0]), "m")
+        self.assertTrue(j5.equals(j5))
+        j6 = cf.Data(np.ma.array([10.0, 2.0, 3.0], mask=[1, 0, 0]), "m")
+        self.assertTrue(j6.equals(j6))
+        with self.assertLogs(level=cf.log_level().value) as catch:
+            print("J5 EQUALS J6 NOW <<<<<<<<<<<<<<<<<<<<<<<<<")
+            self.assertFalse(j5.equals(j6))
```
davidhassell, line 274: Should be True!

davidhassell: ... and might well be when `masked_equal=True` (see other comment on cf.Data.equals)

davidhassell (commented):
I don't think I've reviewed everything, yet, but I'll submit what I've thought of so far. Thanks.

cf/data/dask_utils.py
```diff
@@ -5,9 +5,101 @@
...
+
+            y: a dask array to compare with x
+
+            masked_equal:
+                Whether masked values in a and b are considered
+                equal (True) or not (False). They are considered equal
+                by default.
+
+            rtol:
+                Relative tolerance. Default is 1e-05.
```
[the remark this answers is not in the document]
sadielbartholomew: Ah, good point - I assumed we would be consistent with NumPy but definitely should have checked! Will get those changed to the cf values. Thanks.

cf/data/utils.py
```diff
@@ -10,6 +10,56 @@
...
+    # "2000-12-3 12:00", yet could be considered, or encoded as, numeric?
+    dtype = array.dtype
+    # This checks if the dtype is either a standard "numeric" type (i.e.
+    # int types, floating point types or complex floating point types)
+    # or Boolean, which are effectively a restricted int type (0 or 1).
+    # We determine the former by seeing if it sits under the 'np.number'
+    # top-level dtype in the NumPy dtype hierarchy; see the
+    # 'Hierarchy of type objects' figure diagram under:
+    # https://numpy.org/doc/stable/reference/arrays.scalars.html#scalars
+    return np.issubdtype(dtype, np.number) or np.issubdtype(dtype, np.bool_)
```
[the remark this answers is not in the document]
sadielbartholomew: I tried to cover that further into the comment above, "or Boolean, which are effectively a restricted int type (0 or 1)." Zeroes and ones are numeric to me, fundamentally, and indeed I noticed when checking for any similar functions existing in Dask or NumPy that (they don't seem to define their own, but) Dask imports the pandas function `is_numeric_dtype` (same name I have used, but internally marked) and they [also consider Boolean as numeric](https://github.com/pandas-dev/pandas/blob/945c9ed766a61c7d2c0a7cbb251b6edebf9cb7d5/pandas/core/dtypes/common.py#L1188-L1227 ).

Happy to remove it though if you have a decent reason that Booleans shouldn't be considered numeric? Or could add a flag kwarg to toggle such behaviour?

cf/data/data.py
```diff
@@ -9110,37 +9123,57 @@ def equals(
...
+            return False
 
-            if not _numpy_allclose(
-                array0, array1, rtol=float(rtol), atol=float(atol)
-            ):
-                logger.info(
-                    "{0}: Different array values (atol={1}, "
-                    "rtol={2})".format(self.__class__.__name__, atol, rtol)
-                )
+        mask_comparison = da.all(
```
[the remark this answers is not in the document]
sadielbartholomew: I am not convinced it would be more efficient because I would have thought `all` would be most efficient due to its nature, unless I am missing something here? I can profile them to see... will let you know and choose the most efficient one.

cf/test/test_Data.py
```diff
@@ -131,6 +131,293 @@ class DataTest(unittest.TestCase):
...
+        shape = 3, 4
+        a = np.arange(12).reshape(*shape)
+
+        d = cf.Data(a, "m")
+        self.assertTrue(d.equals(d))  # also do self-equality checks!
+
+        d2 = cf.Data(a.astype(np.float32), "m")  # different datatype to d
+        self.assertTrue(d2.equals(d2))
+        with self.assertLogs(level=cf.log_level().value) as catch:
+            self.assertFalse(d2.equals(d))
```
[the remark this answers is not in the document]
sadielbartholomew: Ah true, there is the issue of convertible datatypes with equivalent numeric data still to address, then, I think! Thanks for the important spot. I will get on it...

cf/test/test_Data.py
sadielbartholomew: That's wise, very good point. Thanks, I'll get those statements all converted in that way.

cf/test/test_Data.py
sadielbartholomew: Yes, as I covered in [my comment here](https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r753277279) (sorry if it was a bit verbose!) with `masked_equal=True` it does become `True`. Is that the most appropriate choice to set? I wasn't sure, as indicated, especially given my uncertainty about the nature of the input we intend for the `data_comparison`.

For now (until we decide on the `masked_equal` value, etc.), I'll convert this line to as `assertTrue` so we are testing for the behaviour we want.

cf/test/test_Data.py
davidhassell: ... although the "are Nan/infs equal" question may have some bearing on using `self is other`

cf/data/data.py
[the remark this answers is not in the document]
davidhassell: You're quite right. I withdraw my suggestion. `da.all(da.equal())` is certainly faster for boolean arrays than `da.allclose()`

cf/data/dask_utils.py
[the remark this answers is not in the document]
davidhassell: I think that for the range of meteorological inputs we expect the numpy defaults give too many false `True`s. Safer to have a false `False` in my book. By the way, that thread that you put me onto (https://github.com/numpy/numpy/issues/10161) was very interesting.

cf/test/test_Data.py
davidhassell: (discussing off-line - will report back later ...)

cf/test/test_Data.py
davidhassell: It was mentioned elsewhere that "`cf.Data.equals` has, up till now at least, short circuited if `self is other`". But this is not the case_, i.e. `cf.Data,equals` always compares the data values, even when "self is other"

It is the case, however, that `cfdm.Data.equals` short circuits when self is other, and I incorrectly  assumed that this was inherited by cf. I think that I probably meant to make sure that `cf.Data,equals` does short circuit in this case, but overlooked it.

So in the status quo:
```python
>>> import numpy as np; import cf, cfdm
>>> h = cfdm.Data(np.full((3, 4), np.nan), "m")    ## cfdm
>>> h.equals(h, verbose=-1)
True
>>> h = cf.Data(np.full((3, 4), np.nan), "m")      ## cf
>>> h.equals(h, verbose=-1)
Data: Different array values (atol=2.220446049250313e-16, rtol=2.220446049250313e-16)
False
```

I.e. cfdm short cricuits and gives `True`, but cf does not and give `False`, because numpy itself deems NaNs no equal.

As this falls in the realms of cock up rather than conspiracy, I think that we should consider  ourselves free to define a consistent behaviour across cfdm and cf, regardless of whcih each library does at present.

That's a good place to stop. Now I'm up to speed I can start to think about the problem as stated ...

sadielbartholomew: Agreed. Whatever we go for should be the case for both libraries. Good thinking to consider the cfdm behaviour whilst we are considering this, too.

cf/test/test_Data.py
sadielbartholomew: Thanks David, all of your thoughts are much appreciated.

cf/data/utils.py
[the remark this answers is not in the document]
davidhassell: Hmm, more confusion, but not of your making! cfdm and cf short circuit on non-equal data types: https://github.com/NCAS-CMS/cfdm/blob/master/cfdm/data/data.py#L2256

I recall that that was deliberate, as cf equality goes beyond just numerical equivalence. For example, the units have to be equal and not just equivalent. Image that we have data from two netCDF files that have different precisions but equivalent values, are these variables equal? We thought not, since precision can sometimes matter beyond just noisy bits when you actually do stuff with your data - e.g. the global integral of a 64-bit field would not be the same as the same field truncated to 32-bits, even though numpy might say that the fields are equal element-wise.

What do you think of that history?

Your suggestion of new keywords to control this (NaNa, infs, dtypes, and even units) are sounding like good ideas to me right now ... but what would the defaults be?

cf/test/test_Data.py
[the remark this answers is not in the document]
sadielbartholomew: Hi @davidhassell, after resolving (most) of your other comments (the rest I will shortly summarise where I am at with), this one still gives `False`, and it turns out that it does so because it is hitting the `cfdm` `Data.equals` check which is giving False:

https://github.com/NCAS-CMS/cf-python/blob/1aa619def8d9d89071619a6d7df97b219e3f936a/cf/data/data.py#L10979-L10989

so ultimately this is an issue due to inconsistency with the `cfdm` logic, i.e. basically the issue you pointed out in https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r756266071. I will open up an issue over in `cfdm` for this., For now I have commented out this one test and added a warning `print`ed whenever we hit the above logic to give a `False` result via `cfdm`.

cf/test/test_Data.py
sadielbartholomew: This is now resolved, I believe, after discussion in other feedback comments.

davidhassell:
Hi Sadie,

> whether we should consider non-equal but equivalent data types, having equivalent data after conversion, as 'equal' or not (see Data.equals: add unit test & migrate to Dask #254 (comment)).

I think that different-but-equivalent data-types should be *not equal* (by default) in the eyes of `Data.equals`. The method is meant to be strict (by default) about things like this, as there are times when different data-types are important. If none of the "similar" things matter, then one should use `(d == e).all()` to test equality - this accounts for different shapes, data-types, the works.

davidhassell:
Hi Sadie,

> whether we should consider NaNs equal for self-equality of arrays with any NaNs in (see Data.equals: add unit test & migrate to Dask #254 (comment)) (I can't see anything in my notes about thoughts on this so I am not sure if we have covered this before at all?)

This may be controversial, but I think that that NaNs should be *equal* by default to `Data.equals`. However, `==` should do whatever ever numpy does normally (is that `NaN != NaN`?). This is beacuse whilst I wouldn't like to say any one NaN is equal to any other NaN, at a higher level, and array full of NaNs (for instance) is "the same" as another array full of NaNs. This behaviour allows this:
```python
>>> f = cf.read('file_containing_NaNs.nc')
>>> g = cf.read('file_containing_NaNs.nc')
>>> f.equals(g)
True
```
which is what I would expect. Does any of that make sense?

sadielbartholomew:
HI again David, thanks for the new clarifications.

> I think that different-but-equivalent data-types should be not equal (by default) in the eyes of Data.equals

Understood, that seems very sensible, and noted RE comparisons for something more akin to 'equivalency' - so now I need to check that this PR implements that behaviour, like cfdm does. I think I was a bit confused up to now as to your wishes on this, since your new comment, and some older ones, seem to contradict [this comment](https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r754202778)?

> This may be controversial, but I think that that NaNs should be equal by default to Data.equals. However, == should do whatever ever numpy does normally (is that NaN != NaN?).

That makes perfect sense and is more than fine by me; that usage of NaNs is probably better suited to the context of the cf libraries and our specific usage of, and distinction between, `equals` and `==`.

I'll add a new commit or two to ensure the behaviour matches that agreed above.

Have just realised one thing I may have missed, namely your question https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r754175121 and my response: it isn't clear from your later comment on that thread whether you are happy for me to leave the "or Boolean" check in?

sadielbartholomew:
(Discussed externally, now I will push some new commits to tie up the loose ends in the ways agreed from that conversation...)

sadielbartholomew:
Ready for review @davidhassell, with the caveat that I am not adding an `equal_nan` kwarg in this PR. I actually went in to add it but found that it won't be trivial because `np.ma.allclose` [doesn't have an `equal_nan` kwarg itself](https://numpy.org/doc/stable/reference/generated/numpy.ma.allclose.html) (although `np.allclose` [does](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html#numpy.allclose)) so in order to implement it I would have to use some other means to do the ultimate `allclose` inside `_da_ma_allclose`, or some pre-or post-processing. How about we think about that as part of a possible follow-up PR? For this PR, the behaviour is that of `numpy`, ultimately that any array with `NaN`s in with not be equal to another object.

cf/test/test_Data_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        c = np.ma.array([1.0, 2.0, 100.0], mask=[1, 0, 0])
+        d = np.array([1.0, 2.0, 3.0])
+        e = a + 5e-04  # outside of tolerance to set, namely rtol=1e-05
+        f = a + 5e-06  # within set tolerance to be specified, as above
+
+        # Test the function with these inputs as both numpy and dask arrays...
+        allclose = cf.data.dask_utils._da_ma_allclose
+        da_ = da.from_array(a)
+
+        self.assertTrue(allclose(a, a).compute())
```
davidhassell, line 36: Do we need to test `_da_ma_allclose` for non-dask array inputs?

davidhassell (commented):
Hi  Sadie - I think we're pretty much there, excellent work.

I have "commented" rather than "requested changes", as we might need to sign off on a couple of things (perhaps the string data type question?).

Thanks,
David

cf/test/test_Data.py
```diff
@@ -131,6 +131,289 @@ class DataTest(unittest.TestCase):
...
     #    test_only = ['test_Data__init__dtype_mask']
 
+    def test_Data_equals(self):
+        if self.test_only and inspect.stack()[0][3] not in self.test_only:
+            return
+
+        shape = 3, 4
+        a = np.arange(12).reshape(*shape)
+
+        d = cf.Data(a, "m")
```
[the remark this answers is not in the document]
sadielbartholomew: Aha, good thinking to test against `self` as well as the copy. Committing your suggestion. Thanks.

cf/test/test_Data.py
```diff
@@ -131,6 +131,289 @@ class DataTest(unittest.TestCase):
...
+                    for log_msg in catch.output
+                )
+            )
+
+        # Test rtol and atol parameters
+        k1 = cf.Data(np.array([10.0, 20.0]))
+        self.assertTrue(k1.equals(k1.copy()))
+        k2 = cf.Data(np.array([10.01, 20.01]))
+        self.assertTrue(k2.equals(k2.copy()))
+        for ks in [(k1, k2), (k2, k1)]:  # to test symmetry/commutativity
```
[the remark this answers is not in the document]
sadielbartholomew: Good spot. I'll remove the loop over the pairs.

cf/test/test_Data.py
```diff
@@ -131,6 +131,289 @@ class DataTest(unittest.TestCase):
...
+        # Note these *should* be considered equal inside cf.Data, and indeed
+        # np.ma.allclose and our own _da_ma_allclose methods also hold
+        # these to be 'allclose'.
+        j5 = cf.Data(np.ma.array([1.0, 2.0, 3.0], mask=[1, 0, 0]), "m")
+        self.assertTrue(j5.equals(j5.copy()))
+        j6 = cf.Data(np.ma.array([10.0, 2.0, 3.0], mask=[1, 0, 0]), "m")
+        self.assertTrue(j6.equals(j6.copy()))
+        self.assertTrue(j5.equals(j6))
+
+        # Test non-numeric dtype arrays
```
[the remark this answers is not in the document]
sadielbartholomew: Well-spotted. I remember you mentioning that case when we discussed this a while back, but somehow have neglected to incorporate it into the unit test, apologies. I'll add a test in now and see what currently happens. As for what behaviour we should have and it should test for:

> "do different string data types matter if the data are the same?"

I suggest we be consistent with the behaviour now implemented for numeric dtypes, namely that a different string data type will be considered unequal, even with the same data. Happy to do differently if you prefer (with or without justification), though!

cf/test/test_Data_utils.py
davidhassell: I think the function is fine, just that the `a` in `self.assertTrue(allclose(a, a).compute())` is a numpy array, not a dask array. Apologies if I'm missing something

cf/test/test_Data.py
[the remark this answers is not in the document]
davidhassell: Hmm. Having a had a tinker with `netCDF4.Dataset`, the numpy data type is lost upon when written to a file as a VLEN, i.e. the Python encoding of the string is immaterial. Perhaps this becomes clear if we consider two strings encoded as `U` and `S`. These can be the same in a way that a `float32` can never be the same as `float64`. 

So, I think that for the string case, perhaps we should allow any two string-like data types to be equal for out purposes.

What do you think?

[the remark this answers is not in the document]
davidhassell: Does https://github.com/NCAS-CMS/cfdm/blob/master/cfdm/data/data.py#L2256-L2261 imply that this while test can be dealt with in cfdm? I.e. we modify cfdm to let different string types through?

cf/test/test_Data_utils.py
sadielbartholomew: Ah, I see - in that case I will amend the test. Thanks.

cf/test/test_Data.py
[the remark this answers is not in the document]
sadielbartholomew: > So, I think that for the string case, perhaps we should allow any two string-like data types to be equal for out purposes.

Makes sense! Let me implement that in a new commit.

sadielbartholomew:
Thanks once again for the feedback, @davidhassell. All should be resolved now, noting that for the issue outlined in [the thread here](https://github.com/NCAS-CMS/cf-python/pull/254#discussion_r786631596) I have now adjusted the test now so it tests for the agreed behaviour, but commented out the line with the assertion in question, which currently fails, with a note that the correct behaviour will be added via cfdm:

https://github.com/NCAS-CMS/cf-python/pull/254/files#diff-bc5ca24d93eb1dce7c459430c3decfc73dd85fded3face2985e76b9565cf6881R284-R287

So it isn't resolved in the sense of being implemented, but is ready in terms of this PR: I will open up a follow-up issue on cfdm to cover it.

davidhassell:
Just had a thought - it'd be good to set all of the dask arrays in all of the tests to have multiple dask chunks:

```python
d = cf.Data(a, "m", chunks=(2, 2)
```

sadielbartholomew:
> Just had a thought - it'd be good to set all of the dask arrays in all of the tests to have multiple dask chunks:

Right-o, good idea. One moment and I'll push a commit doing that.

sadielbartholomew:
@davidhassell since my last comment I've:

* added dask chunks on all `cf.Data` calls in `test_Data_equals`, and I will add chunks to the other Data-module tests as I go along with daskification (and can throw chunks into the other already-ready tests in a lone commit to push post-merge);
* handled the usgae of the `daskified` decorator to reduce the verbosity of the daskification monitoring messages;
* resolved most of the linting failures, many of which appear to be residual from the recent merge of `master`, and some of which I imagine were from your initial LAMA to Dask work, though I couldn't fully resolve the linting to get `flake8` to pass via `pre-commit run --all-files` because I wasn't sure of the nature of two undefined variables, please could you clarify so I can rectify it and the linting tests will pass?

#### Outstanding linting failures

What are/were `is_small` and `is_very_small`? Can they go or 

```console
$ pre-commit run --all-files
Check python ast.........................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
black....................................................................Passed
docformatter.............................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

cf/data/creation.py:137:12: F821 undefined name 'is_small'
cf/data/creation.py:196:12: F821 undefined name 'is_very_small'
cf/data/creation.py:234:12: F821 undefined name 'is_small'
cf/data/creation.py:240:12: F821 undefined name 'is_small'
cf/data/creation.py:266:20: F821 undefined name 'is_small'

isort (python)...........................................................Passed
isort (cython).......................................(no files to check)Skipped
isort (pyi)..........................................(no files to check)Skipped
```

davidhassell:
Hi Sadie,

> What are/were is_small and is_very_small? Can they go or does some logic need to be (re-)added?

I think "were"! and the easiest thing is probably just to comment out those lines - that won't stop any (relevant) units tests from passing, and they'll get wiped anyway when #297 is merged.

sadielbartholomew:
> I think "were"!

Sorry, I realise now I had worded that badly - obviously they have gone as they are not defined, but by 'were' I meant to question whether it has gone AWOL and how to re-add it :slightly_smiling_face:

> and the easiest thing is probably just to comment out those lines - that won't stop any (relevant) units tests from passing, and they'll get wiped anyway when #297 is merged.

Aha, sure - I didn't realise #297 would touch them. Will remove those and then we might be ready to merge...

sadielbartholomew:
OK just going to trigger the CI jobs, which should now pass as everything is good locally, including the linting...

davidhassell (approved):
All done!

Issue #263, opened 2021-09-23
Import fails with `NameError: name 'ESMF' is not defined`


lguo-uk:
Hi there,

I have trouble to make cf python run on my Mac. Below is the details.

Any idea?

Thanks,
Liang

__Installation information:__
```shell
/bin/bash:/Users/liangguo$ pip3 install cf-python
Collecting cf-python
  Downloading cf-python-3.10.0.tar.gz (1.7 MB)
     |████████████████████████████████| 1.7 MB 3.3 MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied: numpy>=1.15 in /Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages (from cf-python) (1.21.2)
Collecting psutil>=0.6.0
  Downloading psutil-5.8.0-cp39-cp39-macosx_10_9_x86_64.whl (236 kB)
     |████████████████████████████████| 236 kB 4.6 MB/s
Collecting cfunits>=3.3.3
  Downloading cfunits-3.3.4.tar.gz (41 kB)
     |████████████████████████████████| 41 kB 1.0 MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting cfdm<1.8.10.0,>=1.8.9.0
  Downloading cfdm-1.8.9.0.tar.gz (358 kB)
     |████████████████████████████████| 358 kB 1.7 MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Collecting cftime>=1.5.0
  Downloading cftime-1.5.0-cp39-cp39-macosx_10_9_x86_64.whl (214 kB)
     |████████████████████████████████| 214 kB 4.1 MB/s
Collecting netCDF4>=1.5.4
  Downloading netCDF4-1.5.7-cp39-cp39-macosx_10_9_x86_64.whl (4.0 MB)
     |████████████████████████████████| 4.0 MB 3.6 MB/s
Collecting netcdf-flattener>=1.2.0
  Downloading netcdf-flattener-1.2.0.tar.gz (26 kB)
Building wheels for collected packages: cf-python, cfdm, cfunits, netcdf-flattener
  Building wheel for cf-python (PEP 517) ... done
[...]

PR #264, opened 2021-09-23
Fix treatment of missing ESMF library

davidhassell:
Fixes: #263

### DIFF
--- a/cf/regrid/utils.py
+++ b/cf/regrid/utils.py
@@ -11,8 +11,8 @@ from ..functions import regrid_logging
 if _found_ESMF:
     try:
         import ESMF
-    except Exception as error:
-        print(f"WARNING: Can not import ESMF for regridding: {error}")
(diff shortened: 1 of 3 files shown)


cf/regrid/utils.py
```diff
@@ -1214,6 +1214,12 @@ def regrid_initialize():
             A singleton instance of the `ESMF` manager.
 
     """
+    if not _found_ESMF:
+        raise RuntimeError(
+            "Regridding methods will work unless "
```
sadielbartholomew, line 1219:
```suggestion
            "Regridding methods will not work unless "
```
Surely?!

sadielbartholomew (changes requested):
All good, fixing the issue at hand, except that the new error message introduced needs negating (see inline comment) else much confusion may ensue! Once that's addressed, feel free to merge (else I can do it when I check back).

PR #253, opened 2021-08-27
cf.Data.can_compute always returns True (for now)

davidhassell:
Fixes (for now) #252

Issue #252, opened 2021-08-27
Missing array in representation after a `__setitem__`


sadielbartholomew:
For the `lama-to-dask` branch for #182, assignment for some array value(s) seems to result in the string representation for both  `repr` and `str` losing the component that renders the (truncated etc.) array output, for example:

```pycon
>>> import cf
>>> import numpy as np
>>> d = cf.Data(np.arange(12).reshape(3, 4), "m")
>>> print(d)
[[0, ..., 11]] m
>>> d
<CF Data(3, 4): [[0, ..., 11]] m>
>>> d[0, 0] = 1
>>> d.array
array([[ 1,  1,  2,  3],
       [ 4,  5,  6,  7],
       [ 8,  9, 10, 11]])
>>> print(d)
 m
>>> d
<CF Data(3, 4):  m>
```
where only the units are included in the output for a `Data` object that is otherwise unchanged to one giving the expected output, except for one assignment performed.

I haven't had time to look into why this might be occurring, nor investigate the extent of cases in which it does occur, but it looks like it is likely something to do with `__setitem__`. @davidhassell, since you volunteered to take charge of `setitem` matters for the Dask migration, do you want to look into this? (Happy to try to resolve it myself, if not.)

davidhassell:
Superceded by #493

PR #265, opened 2021-09-30
dask: Force `Data.can_compute` to return True (for now ...)

davidhassell:
Fixes (for now) #252

Replace the messy #253


sadielbartholomew (approved):
Perfect. Thanks for creating the new, clean PR.

PR #267, opened 2021-09-30
Fix autocyclic bug

### LINKED ISSUE #237: set period if dimension cyclicity automatically set
Hi Sadie, Hi David,

I've noticed that when I define a field with a cyclic spatial dimension (e.g. -180 degrees east to +180 degrees east), I get an error message in `cf-python==3.10.0` that I was not getting in `cf-python==3.8.0` because my cyclic dimension does not have a period. See example below:

```python
import cf

f = cf.Field()

lon = f.set_construct(
    cf.DimensionCoordinate(
        properties={'standard_name': 'longitude',
                    'units': 'degrees_east',
                    'axis': 'X'},
        data=cf.Data([-150, -90, -30, 30, 90, 150]),
        bounds=cf.Bounds(data=cf.Data([[-180, -120], [-120, -60], [-60, 0], [0, 60], [60, 120], [120, 180]]))
    ),
    axes=f.set_construct(cf.DomainAxis(size=6))
)
```

With this example, when setting the longitude construct, I get thrown the following:

> Traceback (most recent call last):
  File "/Users/thibhlln/opt/miniconda3/envs/hj38/lib/python3.8/site-packages/cf/mixin/fielddomain.py", line 2651, in set_construct
    self.autocyclic(key=out, coord=construct, config=autocyclic)
  File "/Users/thibhlln/opt/miniconda3/envs/hj38/lib/python3.8/site-packages/cfdm/decorators.py", line 181, in verbose_override_wrapper
  File "/Users/thibhlln/opt/miniconda3/envs/hj38/lib/python3.8/site-packages/cf/mixin/fielddomain.py", line 1244, in autocyclic
    self.cyclic(key, iscyclic=True, period=period, config=config)
  File "/Users/thibhlln/opt/miniconda3/envs/hj38/lib/python3.8/site-packages/cf/field.py", line 5116, in cyclic
    raise ValueError(
ValueError: A cyclic dimension coordinate must have a period

I don't know if cyclicity was automatically checked for and set before, and I believe that this is a good thing that it is if the bounds show that it is a cyclic dimension, however, I am wondering whether the period could not also be set automatically at the same time? This would be very convenient.

Thank you.
Thibault

davidhassell:
Fixes #237

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -14,7 +14,7 @@ version 3.11.0
   (https://github.com/NCAS-CMS/cf-python/issues/230)
 * Allow for regridding using a destination field featuring size 1 dimension(s)
   (https://github.com/NCAS-CMS/cf-python/issues/250)
-* Fix but that sometimes caused `cf.Field.autocyclic` to fail when the
+* Fix bug that sometimes caused `cf.Field.autocyclic` to fail when
   setting a construct that is cyclic and has a defined period
   (https://github.com/NCAS-CMS/cf-python/issues/237)
 


cf/mixin/fielddomain.py
```diff
@@ -1207,13 +1207,15 @@ def autocyclic(self, key=None, coord=None, verbose=None, config={}):
...
 
         period = coord.period()
+
         if period is not None:
             has_period = True
         else:
             period = config.get("period")
-            has_period = False
+            if period is None:
+                has_period = False
```
sadielbartholomew, line 1216: By the logic here I think there can be a case where we could run into a `NameError` because `has_period` would not be defined (if `config.get("period")` is not `None`?), so to protect against that, I would add something like:

```suggestion
            if period is None:
                has_period = False
            else:
                has_period = True
              
```

sadielbartholomew (changes requested):
One small change is required I think, though the logic does fix the issue at hand so otherwise this is all good.

cf/mixin/fielddomain.py
davidhassell: Good spot

Changelog.rst
```diff
@@ -14,6 +14,9 @@ version 3.??.?
   (https://github.com/NCAS-CMS/cf-python/issues/230)
 * Allow for regridding using a destination field featuring size 1 dimension(s)
   (https://github.com/NCAS-CMS/cf-python/issues/250)
+* Fix but that sometimes caused `cf.Field.autocyclic` to fail when the
+  setting a construct that is cyclic and has a defined period
```
sadielbartholomew, line 18:
```suggestion
* Fix bug that sometimes caused `cf.Field.autocyclic` to fail when
  setting a construct that is cyclic and has a defined period
```

sadielbartholomew (commented):
One more minor thing, then we're good to merge...

sadielbartholomew (approved):

PR #269, opened 2021-10-05
dask: Move dask utils to a new file

davidhassell:


sadielbartholomew (approved):
The `test_Data` module passes with known skips locally and on the CI here. All good!

(Let's keep chipping away to move logic out of the bloated `data.data` :slightly_smiling_face: .)

PR #282, opened 2021-11-25
Allo element-wise comparisons with `None`

### LINKED ISSUE #281: attempting to compare a cf field with None returns an error
I have some code which has a comparator that looks like

```
x = None
loop over stuff
    if x == None:
       # create a cf field called x
   use x in each instance of the loop thereafter.
```
(I actually do this with `if (x,y) == (None,None)` but I think it's the simpler problem. My error message is this:

```
  File "/home/users/lawrence/GitRepos/cmip-play/regrid.py", line 71, in <module>
    if (land, sea) ==  (None, None):
  File "/home/users/lawrence/miniconda3v/envs/py39n/lib/python3.9/site-packages/cf/mixin/propertiesdata.py", line 342, in __eq__
    return self._binary_operation(y, "__eq__")
  File "/home/users/lawrence/miniconda3v/envs/py39n/lib/python3.9/site-packages/cf/field.py", line 1918, in _binary_operation
    if other.ndim > 0:
  File "/home/users/lawrence/miniconda3v/envs/py39n/lib/python3.9/site-packages/cf/data/data.py", line 7679, in ndim
    return self._ndim
  File "/home/users/lawrence/miniconda3v/envs/py39n/lib/python3.9/site-packages/cf/data/data.py", line 7256, in _ndim
    return self._custom["_ndim"]
KeyError: '_ndim'
```
It looks like the __eq__ method needs to start by seeing if the comparator is of the same class before testing attributes.

I would expect this to work. Clearly it's easy to work around, but ...
 
```
cf.environment(paths=False)
Platform: Linux-3.10.0-1160.42.2.el7.x86_64-x86_64-with-glibc2.17
HDF5 library: 1.10.6
netcdf library: 4.7.4
udunits2 library: /home/users/lawrence/miniconda3v/envs/py39n/lib/libudunits2.so.0
Python: 3.9.7
netCDF4: 1.5.6
cftime: 1.5.1.1
numpy: 1.21.2
psutil: 5.8.0
scipy: 1.5.3
matplotlib: 3.4.3
ESMF: 8.0.1
cfdm: 1.9.0.1
cfunits: 3.3.4
cfplot: 3.1.17
cf: 3.11.0
```

davidhassell:
Fixes #281

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -4654,7 +4654,11 @@ class Data(Container, cfdm.Data):
                     # .timetuple()[0:6], microsecond=other.microsecond,
                     calendar=getattr(self.Units, "calendar", "standard"),
                 )
-
+            elif other is None:
+                # Can't sensibly initialize a Data object from a bare
+                # `None` (issue #281)
+                other = numpy_array(None, dtype=object)
+   
             other = type(self).asdata(other)
 
         #            if other._isdt and self.Units.isreftime:
--- a/cf/field.py
+++ b/cf/field.py
@@ -1914,6 +1914,9 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
                 # required so can use the metadata-unaware parent
                 # method
                 # ----------------------------------------------------
+                if other is None:
+                    other = numpy_array(None, dtype=object)
+                    
                 other = Data(other)
                 if other.ndim > 0:
                     other.squeeze(inplace=True)
--- a/cf/mixin/propertiesdata.py
+++ b/cf/mixin/propertiesdata.py
@@ -586,6 +586,8 @@ class PropertiesData(Properties):
 
         if isinstance(y, self.__class__):
             y = y.data
+        elif y is None:
+            y = Data(numpy_array(None, dtype=object))
 
         if not inplace:
             new = self.copy()  # data=False) TODO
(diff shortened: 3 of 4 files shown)


davidhassell:
Note to selves: think about this as part of the dask work.

sadielbartholomew (approved):
All good, assuming the desired behaviour for the specific case Bryan raised (as I am guessing from the thread on the corresponding Issue), along with the obvious general behaviour, is that:

> In any case, I'm glad you're putting in a PR so the right error is raised rather than a stackdump.

namely that we return the same error as `numpy` (https://github.com/NCAS-CMS/cf-python/issues/281#issuecomment-978944031), which is indeed implemented:

```pycon
>>> d = cf.Data(np.array([1, 2, 3]))
>>> if d != None:
...     print("nope")
... 
  File "/home/sadie/cf-python/cf/data/data.py", line 1544, in __bool__
    raise ValueError(
ValueError: The truth value of Data with more than one element is ambiguous. Use d.any() or d.all()
```

sadielbartholomew:
(The linting workflow is failing on some spurious whitespace, which I can get later if you prefer to merge this as-is. There is a 3.6 workflow running, too, when it shouldn't as we dropped 3.6, hence the failure of it, so I'll deal with that.)

PR #284, opened 2021-11-26
Fix `cf.aggregate` failures when using the 'dimension' keyword

### LINKED ISSUE #283: `cf.aggregate` fails when creating new axes with the `dimension` keyword
At v3.11.0 (and all other versions since 3.0.0) , `cf.aggregate` fails when creating new axes for each input field from one or more given properties, using the `dimension` keyword argument:

```python
>>> import cf
>>> print (cf.__version__)
3.11.0
>>> f = cf.example_field(0)
>>> g = f.copy()
>>> f.set_property('sim', 'r1i1p1f1')
>>> g.set_property('sim', 'r2i1p1f1')
>>> a = cf.aggregate([f, g], dimension='sim')
  File "agg.py", line 20, in <module>
    a = cf.aggregate([f, g], dimension='sim')[0]
  File "/home/david/NCAS-CMS/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
  File "/home/david/cf-python/cf/aggregate.py", line 1721, in aggregate
    meta = _Meta(
  File "/home/david/cf-python/cf/aggregate.py", line 355, in __init__
    aux_coords = {
  File "/home/david/cf-python/cf/aggregate.py", line 358, in <dictcomp>
    if axis in construct_axes[aux]
KeyError: 'auxiliarycoordinate0'
```

This is simply a bug in code that was not covered by the units tests. PR fix to follow.

Thank you to @bnlawrence for originally spotting this.

davidhassell:
Fixes #283

### DIFF
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -138,6 +138,7 @@ class _Meta:
         relaxed_identities=False,
         ncvar_identities=False,
         field_identity=None,
+        copy=True,
     ):
         """**initialization**
 
@@ -163,7 +164,8 @@ class _Meta:
             dimension: (sequence of) `str`, optional
                 See the `aggregate` function for details.
 
-        **Examples:**
+            copy: `bool` optional
+                See the `aggregate` function for details.
 
         """
         self._bool = False
@@ -245,6 +247,7 @@ class _Meta:
         # Promote selected properties to 1-d, size 1 auxiliary
         # coordinates
         # ------------------------------------------------------------
+        _copy = copy
         for prop in dimension:
             value = f.get_property(prop, None)
             if value is None:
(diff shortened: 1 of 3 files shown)


cf/aggregate.py
```diff
@@ -258,10 +261,19 @@ def __init__(
             aux_coord.nc_set_variable(prop)
             aux_coord.id = prop
 
+            if _copy:
+                # Copy the field, as we're about to change it.
+                f = f.copy()
+                self.field = f
+                _copy = False
```
sadielbartholomew, line 268: Is this line strictly necessary? `_copy` doesn't seem to be used again in the constructor method and it doesn't seem to make any difference from what I can see with some `print` statements, but I could be missing something.

sadielbartholomew (commented):
One comment (see in-line), otherwise all good:

> This is simply a bug in code that was not covered by the units tests.

Bug fixed and a new unit test added which captures the bug (fails on the current master and not here).

cf/aggregate.py
davidhassell: Hi Sadie, setting `_copy = False` is to prevent the field from being copied multiple times in the event that `dimension` has more than one element. We initialise it to the "copy" keyword because we don't mind about changing the input fields in place if the user has specified `cf.aggregate(..., copy=False)`.

sadielbartholomew: Aha, I see now. Thanks for clarifying David. All good, approving now!

sadielbartholomew (approved):

sadielbartholomew:
Note that the CI test failures are both solely due to the following error:

> FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/cf-python/cf-python/main/cf/test/test_file.nc'

so somehow the given test file is not being generated. I will look into that separately. Merging now.

PR #290, opened 2021-12-23
Fix bug in `cf.Field.derivative` and `cf.relative_vorticity` when   wrapping with periodic coordinates

### LINKED ISSUE #289: Error in `cf.Field.derivative` and `cf.relative_vorticity` when wrapping with periodic coordinates
At v3.11.0, there is a bug in `cf.Field.derivative` which causes the returned boundary values to be incorrect when wrapping with periodic coordinates (such as longitudes).

The centred differences of the coordinates are used, so the wrapped values need to be adjusted by the period in order to be correct. This is not currently the case. 

For example, if there are longitude coordinates of `(22.5, 67.5, 112.5, 157.5, 202.5, 247.5, 292.5, 337.5)` in degrees east, then their centred differences should be `(90, 90, 90, 90,  90, 90, 90, 90)`. However, the current derivative method gives them as `(-270, 90, 90, 90, 90, 90, 90, -270)`.

```python
>>> # v3.11.0 situation - INCORRECT
>>> import cf
>>> f = cf.example_field(0)
>>> print(f)
Cell methods    : area: mean

>>> print(f.derivative('X').array[[0, -1]))
[[-1.85185185e-05 -4.44444444e-05 -2.22222222e-04  1.66666667e-04
   2.55555556e-04  6.66666667e-05 -8.88888889e-05  6.29629630e-05]
 [-8.51851852e-05  1.44444444e-04 -1.11111111e-05 -1.11111111e-05
   2.22222222e-05  1.77777778e-04 -2.66666667e-04  1.03703704e-04]]
```

```python
# CORRECT VALUES:
[[ 5.55555556e-05 -4.44444444e-05 -2.22222222e-04  1.66666667e-04
   2.55555556e-04  6.66666667e-05 -8.88888889e-05 -1.88888889e-04]
 [ 2.55555556e-04  1.44444444e-04 -1.11111111e-05 -1.11111111e-05
   2.22222222e-05  1.77777778e-04 -2.66666667e-04 -3.11111111e-04]]
```

PR to follow

davidhassell:
Fixes #289

### DIFF
--- a/cf/field.py
+++ b/cf/field.py
@@ -16733,7 +16733,7 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
             period = coord.period()
             if period is not None:
                 # Fix the boundary differences for cyclic periodic
-                # coordinates. Need to add extend the coordinates to
+                # coordinates. Need to extend the coordinates to
                 # include a dummy value at each end, grabbed from the
                 # other end, that maintains strict monotonicity.
                 c_data = coord.data
--- a/cf/field.py
+++ b/cf/field.py
@@ -16609,12 +16609,11 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
     ):
         """Return the derivative along the specified axis.
 
-        The derivative is calculated by centred finite differences along
-        the specified axis.
-
-        If the axis is cyclic then the boundary is wrapped around,
-        otherwise missing values are used at the boundary unless one-sided
-        finite differences are requested.
+        The derivative is calculated using centred finite differences
+        apart from at the boundaries (see the *one_sided_at_boundary*
+        parameter). If missing values are present then missing values
+        will be returned at all points where a centred finite
+        difference could not be calculated.
 
         :Parameters:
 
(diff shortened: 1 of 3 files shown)


cf/field.py
```diff
@@ -16690,31 +16722,49 @@ def derivative(
...
-        d = coord.data.convolution_filter(
-            window=[1, 0, -1], axis=0, mode=mode, cval=numpy_nan
-        )
+        # Find the differences of the coordinates
+        d = None
+        if wrap and f.iscyclic(axis):
+            period = coord.period()
+            if period is not None:
+                # Fix the boundary differences for cyclic periodic
+                # coordinates. Need to add extend the coordinates to
```
sadielbartholomew, line 16736:
```suggestion
                # coordinates. Need to extend the coordinates to
```

sadielbartholomew (approved):
Great: fixes the issue at hand and covers the case with new testing. The CI failures are again due to the `test_file.nc` not being generated or recognised (yet to solve that one). (I've made one minor suggestion but it only relates to a commenting typo so feel free to ignore if simpler to merge as-is.)

davidhassell:
Thanks, Sadie, for the quick review. I'll merge, now.

sadielbartholomew:
@davidhassell, apologies since this only came to my attention now due to #291, but there appears to be a test failure related to the changes made in this PR that appears on multiple Actions CI workflows, though it doesn't fail locally (not sure why there is the difference and don't have time to investigate pre-Christmas): see https://github.com/NCAS-CMS/cf-python/pull/291#issuecomment-1000565898. If you could investigate why that failure has emerged after this PR, any insight would be helpful. Thanks.

davidhassell:
https://github.com/NCAS-CMS/cf-python/commit/b77e5a2cd3c6ff9525fcfe161fe93d0cf17b7f1f should fix this for now. The whole vorticity thing needs overhauling, so it's not worth doing a more thorough job at this time.

PR #291, opened 2021-12-23
CI: create `test_file.nc` before step to run test suite

sadielbartholomew:
A netCDF file for testing on, `test_file.nc` is written out and used by the `run_test.py` script that runs the full test suite, and locally this works fine, but for the Actions jobs it is not being recognised, at least within the `test_Field` module:

```bash
Running tests from /home/runner/work/cf-python/cf-python/main/cf/test

    class FieldTest(unittest.TestCase):
  File "/home/runner/work/cf-python/cf-python/main/cf/test/test_Field.py", line 80, in FieldTest
    f = cf.read(filename)[0]
  File "/home/runner/work/cf-python/cf-python/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
  File "/home/runner/work/cf-python/cf-python/main/cf/read_write/read.py", line 699, in read
    open(file_glob, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/cf-python/cf-python/main/cf/test/test_file.nc'
```

This PR aims to fix this so that the required file is generated by the Actions jobs and available for use by that script (as wrapped by the coverage-checking script), as opposed to committing it permanently to the repo which is a simple solution but is not preferred for various reasons.

Some trial-and-error is anticipated to get this working.

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -111,6 +111,16 @@ jobs:
     - name: Notify about starting testing
       run: echo Setup complete. Now starting to run the cf-python test suite...
 
+    # Create a file needed for testing. A separate step is required for this
+    # so the file can be registered and recognised first. Locally, the file
+    # is created and used on-the-fly by the 'run_tests_and_coverage' script.
+    - name: Create test_file.nc
+      shell: bash -l {0}
+      run: |
+        cd ${{ github.workspace }}/main/cf/test
+        python create_test_files.py
+        ls -la
+
     # Finally run the test suite and generate a coverage report!
     - name: Run test suite and generate a coverage report
       shell: bash -l {0}
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -119,6 +119,7 @@ jobs:
       run: |
         cd ${{ github.workspace }}/main/cf/test
         python create_test_files.py
+        python setup_create_field.py
         ls -la
 
     # Finally run the test suite and generate a coverage report!
--- a/.github/workflows/test-source-dist.yml
+++ b/.github/workflows/test-source-dist.yml
@@ -94,6 +94,17 @@ jobs:
     - name: Notify about starting the sdist test
       run: echo Setup complete. Now creating and testing the source dist...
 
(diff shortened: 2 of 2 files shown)


sadielbartholomew:
Opening and closing to trigger the CI jobs...

sadielbartholomew:
The fix applied, which seems to work now for the `run_test_suite` workflow, is to run a separate step to create the test file before the step that runs `run_test.py` via the `run_tests_and_coverage` bash script wrapper. Locally that script generates the test file on-the-fly but I think it doesn't work with the Actions workflow because files need to be persisted explicitly to be recognised and I can't find a way in the documentation to register it as it is created, so a separated step before is harmless.

Now applying the working fix to the `test-source-dist` which is hitting the same issue.

sadielbartholomew:
Re-triggering...

sadielbartholomew:
Again...

sadielbartholomew:
Now the `test-source-dist` workflow single job creates and uses the `test_file.nc` appropriately, but fails for a reason unrelated to this PR, namely:

```pytb
FAIL: test_relative_vorticity_latlong (test_Maths.MathTest)
  File "/home/runner/tmp/test_cf-python/cf/test/test_Maths.py", line 105, in test_relative_vorticity_latlong
    self.assertTrue(numpy.allclose(rv.array, rv_array))

Ran 337 tests in 793.486s

FAILED (failures=1)
```

which I will raise elsewhere. So that workflow can be considered done for the purposes of this PR.

sadielbartholomew:
> Now the test-source-dist workflow single job creates and uses the test_file.nc appropriately, but fails for a reason unrelated to this PR

Ditto for the `run-test-suite`... so all good to go!

davidhassell:
Hi Sadie. Ah - that makes sense, and I should have tested for it in #290. The fix there will change the results of relative vorticity calculations. I'll have a look at `test_relative_vorticity_latlong`

davidhassell:
https://github.com/NCAS-CMS/cf-python/commit/b77e5a2cd3c6ff9525fcfe161fe93d0cf17b7f1f should fix this for now. The whole vorticity thing needs overhauling, so it's not worth doing a more thorough job at this time.

PR #294, opened 2022-01-05
dask: `Data.convolution_filter`

davidhassell:


davidhassell:
Note: I changed the cfdm version to >= 1.9.0.1, as a convenience for my environment - I don't think that there are any knock-on effects from this ....

sadielbartholomew (approved):
Perfect: works, with updated testing, and is very thoroughly commented. Thanks!

sadielbartholomew:
> Note: I changed the cfdm version to >= 1.9.0.1, as a convenience for my environment - I don't think that there are any knock-on effects from this ....

No code-level effects though it does break the CI jobs because they are set to install the previous version specifically. Just to be rigorous I'll push a reversion of the version and re-trigger the CI jobs now. The `test_Data` module passes locally for me but it is always nice to check everything works across each environment as we go...

sadielbartholomew:
Opening and closing to re-trigger to CI jobs...

sadielbartholomew:
Clean sweep of passes, so I'll merge now and update the grand table...

PR #296, opened 2022-01-07
Reduce duplication in `TimeDuration` via internal methods

sadielbartholomew:
After #288 I noted (to myself) it would be helpful to consolidate the `timeduration` module as there was considerable duplication of logic, and best do that for the next release and before I focus in on the Dask migration work fully, hence a quick PR now.

The corresponding changes are trivial enough that peer review seems unnecessary and there are no functional changes so no changelog update required (etc.).

### DIFF
--- a/cf/timeduration.py
+++ b/cf/timeduration.py
@@ -540,10 +540,9 @@ class TimeDuration:
         x.__iadd__(y) <==> x+=y
 
         """
-        if isinstance(other, (self.__class__, int, float)):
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
I have cancelled four of the CI jobs because they were hanging on `test_Field_collapse` for reasons I will investigate elsewhere. All others have passed, which is sufficient given that there is a full test suite pass for me locally, and notably the `test_TimeDuration` module consistently passes.

PR #300, opened 2022-01-24
`Data.func` and friends: migrate to Dask

sadielbartholomew:
Convert the `Data.func` method towards #182, in doing so daskifying several other element-wise operation methods that use it to perform the underlying operation, as follows (though note the below context since in short most of these can and will be then converted to use Dask built-ins rather than calling `func`):

- the trigonometric and hyperbolic methods and their inverses (12 in total: `sin`, `arcsin`, `sinh`, `arcsin` etc.);
- `ceil`;
- `exp`;
- `floor`;
- `rint`;
- `round`;
- `trunc`;

and daskifying `log` except in the case of setting a base that isn't `e`, `2` or `10` which requires which requires `__itruediv__` to be daskified to work.

### Context and plan/proposal for follow-on work

As well as daskifying `func` this PR effectively daskifies the other methods listed above in doing so because they use `func` to operate in the code at present, hence I have removed the test-skipping decorators from the unit tests for, and added the `daskified` marker to, each of the methods (except `log` where a comment has been made to note the nearly-daskified state), **however** such methods are more efficiently daskified by conversion to using the existing Dask built-in equivalents instead of calling `func` this way.

So the plan is to use follow-on PRs to convert from use of `func` to the appropriate built-in in each method. It [turns out](https://docs.dask.org/en/stable/array-api.html) there are Dask built-ins for every one listed above available for us to use directly.

A small complication is in the case of the trig. and hyperbolic methods with a restricted domain (4 of the 12, e.g. `arcsin` which is undefined outside of `[-1, 1]`), since we support a `preserve_invalid` keyword which preserves any `NaN` or `inf` like values rather than masking them as important for the use context of cf-python. In order to still support that for methods we think should have it as a keyword flag, we should use the daskified `func` rather than the Dask built-in, as it already implements that keyword. Overall, therefore, my proposal is as follows, to be applied in a series of new PRs:

*For data methods that apply a trivial operation in an element-by-element manner (see list above), use the following:*

1. *If there is a Dask built-in equivalent method, use that directly, except where there is a valid reason we can't, notably as an example for methods where there is a restricted domain such that the output may have new masked elements, where we want to have a `preserve_invalid` option that NumPy does not support in any trivial way.*
2. *Otherwise use `func` to apply the operation.*

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -2855,7 +2855,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         [-1. -1. -1. -1.  0.  1.  2.  2.  2.]
 
         """
-        # TODODASK: do we need 'out' specification to 'func'? Was out=True
         return self.func(np.ceil, inplace=inplace)
 
     @daskified(_DASKIFIED_VERBOSE)
@@ -10164,7 +10163,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         [-2. -2. -2. -1.  0.  1.  1.  1.  1.]
 
         """
-        # TODODASK: do we need 'out' specification to 'func'? Was out=True
         return self.func(np.floor, inplace=inplace)
 
     @_deprecated_kwarg_check("i")
@@ -11052,7 +11050,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         [-2. -2. -1. -1.  0.  1.  1.  2.  2.]
 
         """
-        # TODODASK: do we need 'out' specification to 'func'? Was out=True
         return self.func(np.rint, inplace=inplace)
 
     def root_mean_square(
@@ -11179,7 +11176,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         [-0., -0., -0., -0.,  0.,  0.,  0.,  0.,  0.]
 
         """
-        # TODODASK: do we need 'out' specification to 'func'? Was out=True
         return self.func(np.round, inplace=inplace, decimals=decimals)
 
     def stats(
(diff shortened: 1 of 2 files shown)


cf/data/data.py
```diff
@@ -10153,7 +10157,8 @@ def floor(self, inplace=False, i=False):
         [-2. -2. -2. -1.  0.  1.  1.  1.  1.]
 
         """
-        return self.func(np.floor, out=True, inplace=inplace)
+        # TODODASK: do we need 'out' specification to 'func'? Was out=True
```
davidhassell, line 10160: In the light of the rewritten `Data.func`, I don't think we need `out` in the `func` call anymore.

cf/data/data.py
```diff
@@ -12516,44 +12529,29 @@ def func(
```
davidhassell, line 12529: I think we can do a `@_deprecated_kwarg_check("out")`

davidhassell (approved):
Hi Sadie, not much to say here, as it all looks good. I agree with you analysis/context/plan/proposal (https://github.com/NCAS-CMS/cf-python/pull/300#issue-1113270230). I can confirm that the tests all pass, and have noted the TODOs in them to be sorted out when other functionality comes on line.

Thanks - please merge when you're ready.

cf/data/data.py
sadielbartholomew: Thanks for the review, David. With quite a bit to note in the opening comment I forgot to mention 'out' so thanks for picking up on it!

I worked out that it was surely there for the optional kwarg to pass through to NumPy `ufunc`-like functions to specifiy the 'alternate array object(s) in which to put the result' but wasn't entirely sure why it was used in cases such as this. I suspected it could go, so glad to hear that is right. I'll get rid of those with a deprecation warning on the kwarg in a new commit.

sadielbartholomew:
Thanks for the feedback, @davidhassell. I've deprecated the 'out' keyword as suggested, and also added a minimal unit test for `Data.func` as an explicit test was missing (and I realised I had not added the 'daskified' marker decorator to the trig. and hyperbolic methods, so did that too in a third commit). None of those updates require a re-review, as you implied in your review comment. I'll just check the CI jobs pass and then merge - re-triggering those via open-close...

PR #301, opened 2022-01-25
`Data.mask`: migrate to Dask

sadielbartholomew:
Convert the `mask` managed attribute towards #182.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -6373,13 +6373,12 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         """
         mask_data_obj = self.copy()
 
-        dx = self._get_dask().copy()
+        dx = self._get_dask()
         mask = da.ma.getmaskarray(dx)
-        mask_data_obj[...] = mask
 
-        mask_data_obj.dtype = _dtype_bool
-        mask_data_obj._Units = _units_None
-        mask_data_obj._hardmask = True
+        mask_data_obj._set_dask(mask, reset_mask_hardness=True)
+        mask_data_obj.override_units(_units_None, inplace=True)
+        mask_data_obj.hardmask = True
 
         return mask_data_obj
 
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -6371,17 +6371,17 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         (12, 73, 96)
 
         """
+        mask_data_obj = self.copy()
+
         dx = self._get_dask().copy()
         mask = da.ma.getmaskarray(dx)
+        mask_data_obj[...] = mask
 
-        mask._Units = _units_None
-        mask._hardmask = True
+        mask_data_obj.dtype = _dtype_bool
+        mask_data_obj._Units = _units_None
+        mask_data_obj._hardmask = True
 
-        # Note pre-daskification this would return a mask object where would
-        # need to apply 'array' prop. to get the array itself, but now, as
-        # consistent with NumPy, will instead return the mask array directly.
-        # Therefore in test suite, previous `x.mask.array` becomes `x.mask`.
-        return mask
+        return mask_data_obj
 
     @staticmethod
     def mask_fpe(*arg):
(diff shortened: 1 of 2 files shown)


davidhassell:
Hi Sadie, I'm afraid that I favour the existing behaviour, which I also see as being consistent with numpy. For a  numpy array `a`, `a.mask` returns a boolean numpy array; similarly for a cf.Data object `d`, `d.mask` returns a boolean cf.Data object. I'm not sure what `d.mask` should return if not a cf.Data object. The cf.Data object mask is lazy (e.g. hasn't triggered a read from disk at definition time), and also supports extra features (such as cyclic axes) which are not supported elsewhere. 

The following also works:
```python
>>> import cf
>>> d = cf.example_field(0).data
>>> a = d.mask.array
>>> (a == d.mask).all()
True
```
i.e. we don't need the `.array` bit for everything to work through.

sadielbartholomew:
Hi @davidhassell, fair enough, thanks for explaining the context - I think in hindsight I was getting confused over the nature of `mask` and after your explanations I agree about the behaviour it should have. I will update this PR accordingly shortly.

cf/data/data.py
```diff
@@ -6366,42 +6366,22 @@ def mask(self):
         (12, 73, 96)
         >>> m = d.mask
         >>> m.dtype
-        dtype('bool')
+        bool
```
davidhassell, line 6369:
```suggestion
        dtype('bool')
```
since:
```python
>>> d = cf.Data([1, 0, 0], dtype=bool)
>>> d.dtype
dtype('bool')

File: cf/data/data.py [hunk 2]
```diff
@@ -6366,42 +6366,22 @@ def mask(self):
...
+        mask = da.ma.getmaskarray(dx)
+        mask_data_obj[...] = mask
 
-        mask._hardmask = True
+        mask_data_obj.dtype = _dtype_bool
+        mask_data_obj._Units = _units_None
+        mask_data_obj._hardmask = True
 
-        return mask
+        return mask_data_obj
```
davidhassell, line 6384: This might not work in all cases, e.g. I'm not sure what happens is `self` is an array of strings. How about:
```python
mask_data_obj = self.copy()

dx = self._get_dask()
mask = da.ma.getmaskarray(dx)
mask_data_obj._set_dask(mask, reset_mask_hardness=True)
mask_data_obj.override_units(_units_None, inplace=True)
mask_data_obj.hardmask = True

return mask_data_obj
```
This is more efficient, too, as there is no `__setitem__` nor `astype` operation. We need to set `hardmask` (without the underscore) so that the `cf_harden_mask` operation is added to the dask graph. The data type is automatically taken from the underlying new Boolean dask array.

davidhassell (changes requested):
Hi Sadie, thanks - I've suggested a different tack on the main method, we can discuss this afternoon, perhaps.

cf/data/data.py
sadielbartholomew: Ah yes, apologies, this was left over from my initial change-of-behaviour idea which was rightly scrapped...

cf/data/data.py
sadielbartholomew: Aha, much better idea than `[...]`, thanks!

I did first have a poke through the code and docs to check we didn't have something that could effectively replace all of the data with new data, rather than setting it all to something new (if the difference makes sense?), so it is strange I didn't arrive at `_set_dask` from that!

sadielbartholomew: > I'm not sure what happens is self is an array of strings.

Sounds like we should add a dedicated test and cover this. I was thinking of extending `test_Data_apply_masking` to first test `.mask`.

sadielbartholomew:
All feedback has now been addressed. Triggering the CI jobs via open-close for a final sanity check...

sadielbartholomew:
All good: local test passes, CI jobs pass, & all feedback has been addressed, including adding a new unit test. Clearance to merge assuming similar conditions from David given externally prior to his taking leave, so merging now.

Issue #302, opened 2022-01-26
LAMA to Dask: replace `func` usage with Dask built-ins


sadielbartholomew:
To do the follow-up work summarised under the heading 'Context and plan/proposal for follow-on work' in https://github.com/NCAS-CMS/cf-python/pull/300#issue-1113270230, i.e:

> ... such methods are more efficiently daskified by conversion to using the existing Dask built-in equivalents instead of calling func this way.
> 
> So the plan is to use follow-on PRs to convert from use of func to the appropriate built-in in each method. It turns out there are Dask built-ins for every one listed above available for us to use directly.

### Methods to convert

- [x] the eight trigonometric and hyperbolic methods and their inverses which don't have restricted domains (all except `arccos`, `arcsin`, `arctanh` and `arccosh`);
- [x] `ceil`;
- [x] `exp`;
- [x] `floor`;
- [x] `rint`;
- [x] `round`;
- [x] `trunc`;
- [x] `log`.

sadielbartholomew:
Closed by #308 and #309.

PR #308, opened 2022-02-02
LAMA to Dask: `func` -> built-in conversion for non-trig. methods

### LINKED ISSUE #303: LAMA to Dask: reinstate testing of arbitrary-base `Data.log`
Un-comment to re-add the following test, deactivated in #300, under the `test_Data_log` unit test, which can be done with the corresponding assertions passing once `__itruediv__` is converted ('mask' in the true comment below should read '\_\_itruediv\_\_', that's a copy-and-paste typo):

https://github.com/NCAS-CMS/cf-python/blob/0206338037b27cb8f0a0d3fb30da2b15a09c41a3/cf/test/test_Data.py#L3393-L3401

sadielbartholomew:
Partially address #302, to tick all sub-issue boxes except the topmost\* and in doing so close #303, since `Data.log` is now fully migrated (and marked as such) because the `/=` operator it uses with a non-`e`/`10`/`2` based is applied in 'Dask space' rather than `cf` space.

##### Note
\* the topmost item,  '_the eight trigonometric and hyperbolic methods and their inverses which don't have restricted domains_', will be tackled as a separate PR to make it easier for the reviewer to check that the correct eight methods have been converted and in a consistent way, with commenting to explain why the other four such methods have been left as `func` calls. And since the code changes required on all of the methods listed in #302 are largely the same, such that any review feedback can be done as one on them all, and it would be more time-consuming to make seven separate PRs for each, I thought it best to split into just two PRs to cover the Issue (second PR for the trig. methods to follow).

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -10138,6 +10138,7 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
 
     @daskified(_DASKIFIED_VERBOSE)
     @_deprecated_kwarg_check("i")
+    @_inplace_enabled(default=False)
     def floor(self, inplace=False, i=False):
         """Return the floor of the data array.
 
@@ -10164,7 +10165,10 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         [-2. -2. -2. -1.  0.  1.  1.  1.  1.]
 
         """
-        return self.func(np.floor, inplace=inplace)
+        d = _inplace_enabled_define_and_cleanup(self)
+        dx = d._get_dask()
+        d._set_dask(da.floor(dx), reset_mask_hardness=False)
+        return d
 
     @_deprecated_kwarg_check("i")
     def outerproduct(self, e, inplace=False, i=False):
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -9219,7 +9219,8 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         if d.Units:
             d.Units = _units_1
 
-        d.func(np.exp, inplace=True)
+        dx = d._get_dask()
+        d._set_dask(da.exp(dx), reset_mask_hardness=False)
 
         return d
 
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -12315,6 +12315,7 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
 
     @daskified(_DASKIFIED_VERBOSE)
     @_deprecated_kwarg_check("i")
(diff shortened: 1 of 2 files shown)


cf/data/data.py
```diff
@@ -12021,16 +12036,21 @@ def log(self, base=None, inplace=False, i=False):
...
+            dx = da.log2(dx)
         else:
-            d.func(np.log, units=_units_1, inplace=True)
-            d /= np.log(base)
+            dx = da.log(dx)
+            dx /= da.log(base)
+
+        d._set_dask(dx, reset_mask_hardness=False)
+
+        d._Units = _units_1  # all logarithm outputs are unitless
```
davidhassell, line 12053: This is nice-ifying the old code, but this might be better (signals intent and avoids explicit underscore access):

```suggestion
        d.override_units(_units_1, inplace=True)  # all logarithm outputs are unitless
```

cf/data/data.py
```diff
@@ -9219,7 +9223,8 @@ def exp(self, inplace=False, i=False):
         if d.Units:
```
davidhassell, line 9223: Could we f-string-ify the above error message?

cf/test/test_Data.py
```diff
@@ -3391,14 +3391,12 @@ def test_Data_log(self):
...
-        # c = cf.Data(a, "s")
-        # d = c.log(base=4)
-        # self.assertTrue((d.array == b).all())
-        # self.assertEqual(d.shape, b.shape)
+        a = np.array([[4, 16, 4 ** 3.5], [0, 1, 0.25]])
+        b = np.log(a) / np.log(4)  # the numpy way, using log rules from school
+        c = cf.Data(a, "s")
+        d = c.log(base=4)
+        self.assertTrue((d.array == b).all())
+        self.assertEqual(d.shape, b.shape)
```
davidhassell, line 3399: Might be nice to test the units:
```suggestion
        self.assertEqual(d.shape, b.shape)
        self.assertEqual(d.Units, cf.Units("1"))
```

davidhassell (changes requested):
Nice job. A few minor suggestions, but all tests pass for me. After you've dealt with the comments (in whatever way you want), please go ahead and merge.

cf/data/data.py
sadielbartholomew: We can indeed. I will add that in a new commit in a moment...

sadielbartholomew:
Quickly re-running he CI jobs to double check they are still passing, triggering via open-close...

sadielbartholomew:
All checks passing, good to go...

PR #309, opened 2022-02-03
LAMA to Dask: func -> built-in conversion for trig. methods

### LINKED ISSUE #302: LAMA to Dask: replace `func` usage with Dask built-ins

sadielbartholomew:
Close #302 by addressing the final item, namely to convert 'the eight trigonometric and hyperbolic methods and their inverses which don't have restricted domains' and also, for completeness, adding comments to the other four such methods to indicate why they haven't been converted in this way.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -12025,7 +12025,10 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         if d.Units.equivalent(_units_radians):
             d.Units = _units_radians
 
-        d.func(np.tanh, units=_units_1, inplace=True)
+        dx = d._get_dask()
+        d._set_dask(da.tanh(dx), reset_mask_hardness=False)
+
+        d.override_units(_units_1, inplace=True)
 
         return d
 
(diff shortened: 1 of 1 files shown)


davidhassell (approved):
Looks good to me.

PR #315, opened 2022-02-07
dask: `Data.empty`, `Data.full`, `Data.ones`, `Data.zeros`

davidhassell:


cf/data/data.py
```diff
@@ -12463,49 +12446,143 @@ def full(
                 The fill value.
 
             dtype: data-type
-                The data-type of the new array. By default the data-type
-                is ``float``.
+                The data-type of the new array. By default the
+                data-type is ``float``.
```
sadielbartholomew, line 12450: I think it is important to note `optional` on the`dtype` kwarg (even if the 'By default' statement implies this), mainly because our behaviour differs from the `numpy` behaviour here with respect to a given `fill_value`, e.g:

```python
>>> d_cf = cf.Data.full((2, 3), fill_value=10)
>>> d_cf
<CF Data(2, 3): [[10.0, ..., 10.0]]>
>>> d_np = np.full((2, 3), fill_value=10)
>>> d_np
array([[10, 10, 10],
       [10, 10, 10]])
```

That should be sufficient though, as you've included a good equivalent example later in the docstring.

cf/data/data.py
```diff
@@ -3179,28 +3167,29 @@ def rechunk(
...
-        balance: `bool`, optional
-            If True, try to make each chunk to be the same size. By
-            default this is not attempted.
+                TODODASK - how to use/import dask config items??
 
-            This means ``balance=True`` will remove any small leftover
-            chunks, so using ``x.rechunk(chunks=len(x) // N,
-            balance=True)`` will almost certainly result in ``N`` chunks.
+            balance: `bool`, optional
+                If True, try to make each chunk to be the same
```
sadielbartholomew, line 3186:
```suggestion
                If True, try to make each chunk the same
```

cf/data/data.py
```diff
@@ -12383,61 +12372,55 @@ def empty(
...
-                *Parameter example:*
-                  ``fill_value=-999.``
+            fill_value: deprecated at version 4.0.0
+                Use `set_fill_value` instead.
 
         :Returns:
 
             `Data`
+                Array of uninitialized (arbitrary) data of the given
+                shape and dtype, and order.
```
sadielbartholomew, line 12408:
```suggestion
                shape and dtype.
```

since I guess reference to 'order' was from copy-and-pasting from `numpy` docstrings and [their `order` keyword](https://numpy.org/doc/stable/reference/generated/numpy.empty.html), which we aren't supporting as far as I see from this PR (though we could maybe eventually add one to these methods to pass through?).

sadielbartholomew (approved):
Some minor comments you may wish to address, but this looks great (and we can ignore the failing 'Run test suite' and similar CI jobs which shouldn't have run for this PR, I will check why they triggered). Thanks. Merge at will!

cf/data/data.py
davidhassell: Interesting. Couldn't we also follow numpy? and do:
```
    dtype : data-type, optional
        The desired data-type for the array  The default, None, means
         ``np.array(fill_value).dtype``.
```

davidhassell: OK - let's go with the numpy behaviour: https://github.com/NCAS-CMS/cf-python/pull/315/commits/8ab7733fdb5e6937ade7f2e770109d2d9678bde6

sadielbartholomew (approved):
All previous feedback addressed. All good, thanks.

sadielbartholomew:
I did notice a few missing `>>> print(d.array)` in the `Data.full` docstring, but that's so trivial and non-controversial I will add them in as a follow-on commit to save time.

PR #319, opened 2022-02-10
dask: `Data.__bool__`

davidhassell:


sadielbartholomew (approved):
Perfect: migrated sensibly and the new unit test covers everything it should. Thanks.

PR #320, opened 2022-02-10
dask: `Data.__contains__`

davidhassell:
`Data.__contains__` was already daskified, but the unit test was being skipped. This PR re-instates the unit test, and takes an opportunity to improve the code (notable adding some short circuits).

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -105,6 +105,7 @@ from .creation import (
 )
 from .dask_utils import (
     _da_ma_allclose,
+    cf_contains,
     cf_harden_mask,
     cf_percentile,
     cf_soften_mask,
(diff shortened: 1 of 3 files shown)


cf/data/data.py
```diff
@@ -648,44 +649,65 @@ def __contains__(self, value):
...
+        >>> 99 in d
+        False
+        >>> [1, 2] in d
+        False
+        >>> cf.Data(2, 'seconds') in d
         False
 
         """
+        # No need to check the dask array if the size of value is
+        # greater than 1
```
sadielbartholomew, line 688: I was wondering about the scope of `__contains__` for arrays, particularly what the intended behaviour is with respect to a `value` that is not a scalar or an array holding only one element (i.e. a sub-array), something like `[0, 1, 2]` in `[[0, 1,  2], [3, 4, 5]]`.

With the logic here, we are short circuiting on anything of size greater than one, so really the scope is limited to single-element values, but I am not sure if that is clearly documented, especially since `>>> [1, 2] in d` is given as a docstring example which leaves me wondering what the result would be when using a sub-arrays e.g. `[0, 1, 2]` (`False`, from this short-circuiting of course).

So overall I advise that we add a parameter signature docstring item to clarify the intended nature of the `value` input. And possibly we could add a warning or even return an error with a value with more than a single element?

Interestingly on the topic, I was looking the the NumPy behaviour and found empirically that they don't short-circuit on such values, but otherwise the behaviour  is very strange, such that they return `True` for sub-arrays having at last one element in the corresponding position with matching value:

```pycon
>>> d_np = np.array([[0, 1,  2], [3, 4, 5]])
>>> [1, 2] in d_np
<stdin>:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
False
>>> d_np = np.array([[0, 1,  2], [3, 4, 5]])
>>> [0 , 1, 2] in d_np
True
>>> [0 , 1, 3] in d_np
True
>>> [0 , 1, 7] in d_np
True
>>> [7 , 7, 7] in d_np
False
>>> [0, 7, 7] in d_np
True
>>> [0, 7, 7, 7] in d_np
False
>>> [0, 0, 5] in d_np
True
>>> [0, 4, 0] in d_np
True
>>> [0, 4, 0, 0] in d_np
False
```
though note[ from the issue here](https://github.com/numpy/numpy/issues/3016) (and maybe the deprecation error below), it looks like it might be about to change to raise an error when a value that has more than a single element is used.

So perhaps we can keep an eye on that and when they have settled on a more well-defined behaviour, we can also make ourselves consistent with that?

sadielbartholomew (approved):
All good, with nice improvements to the test and docstring examples, though one of the new short circuiting cases left me pondering the scope and nature of `__contains__` with `cf` and `numpy`, so please see my longer in-line comment in case it might elicit some changes, minor or otherwise (after you return from leave of course!).

cf/data/data.py
davidhassell: This is very interesting! I think that raising an exception if the value is *not a scalar* (instead of returning False) - is easiest and safest. So `3 in d` would return True or False, but `[3] in d` and `[3, 4] in d` would raise a TypeError. What do you think?

If/when numpy sorts itself out, we can revisit, as you say. However, because the membership test is done in parallel on a chunkwise basis, we'll need to be careful if numpy allows vector membership ... 

This implemented in https://github.com/NCAS-CMS/cf-python/pull/320/commits/7cef4874a8a1420c6be301ac27f232754f424058

sadielbartholomew: > I think that raising an exception if the value is not a scalar (instead of returning False) - is easiest and safest. So 3 in d would return True or False, but [3] in d and [3, 4] in d would raise a TypeError. What do you think?

That behaviour sounds ideal to me and the new commit implements and tests it very well.

> However, because the membership test is done in parallel on a chunkwise basis, we'll need to be careful if numpy allows vector membership ...

Ah, indeed; it won't be simple to implement anything more complicated, so it's good there's no compelling reason to yet!

I will put up an issue to remind us to keep an eye on the situation regarding:

> If/when numpy sorts itself out, we can revisit, as you say.

Issue #295, opened 2022-01-06
Replace LAMA with Dask: grouping methods to migrate


sadielbartholomew:
# Table for #182 

### (See comment update datetime for author and datestamp of last table update)

Note: ~~**306 methods in total**~~ now **315 methods in total** in the table.

| `Data` method name                 | Done? (Bool as emoji, PR tag)   | Group with (if at all)   | Notes 1   | Notes 2   | Notes 3   |
|------------------------------------|---------------------------|--------------------------|-----------|-----------|-----------|
| `HDF_chunks`                       | :heavy_check_mark:  #338  | .                        | .         | Deprecated       | API change      |
| `Units`                            | :heavy_check_mark:   #405      | .                        | .         | .         | .         |
| `_HDF_chunks`                      | :heavy_check_mark:  #338  | .                        | .         | Deprecated        | API change      |
| `_Units`                           | :heavy_check_mark:  | .                        | .         | .         | .         |
| `_YMDhms`                          | :heavy_check_mark: #322   | .                        |  .          | .         | .         |
| `__abs__`                          | :heavy_check_mark:, #409  | UA                        | .  | .         | .         |
| `__add__`                          | :heavy_check_mark:, #409  | A                        | .  | .         | .         |
| `__and__`                          | :heavy_check_mark:, #409  | L                        | . | .         | .         |
| `__array__`                        | :heavy_check_mark:  | CFDM                      | .        | .         | .         |
| `__bool__`                         | :heavy_check_mark:  #319 | .                        | .         | .         | .         |
| `__contains__`                     | :heavy_check_mark:  #320       | .                        | .         | .         | .         |
| `__data__`                         | :heavy_check_mark:  | .                        | .         | .         | .         |
| `__deepcopy__`                     | :heavy_check_mark:  | CFDM                      | .          | .         | .         |
| `__div__`                          | :heavy_check_mark:, #409  | A                        | .  | .         | .         |
| `__doc_template__`                 | :heavy_check_mark: | CFDM                       |  .         | .         | .         |
| `__docstring_package_depth__`      | :heavy_check_mark: | CFDM                       |  .         | .         | .         |
| `__docstring_substitutions__`      | :heavy_check_mark:  | CFDM                     |   .        | .         | .         |
| `__eq__`                           | :heavy_check_mark:, #409  | C                        | . | .         | .         |
| `__float__`                        | :heavy_check_mark:  #337  | .                        | .         | .         | .         |
[...]

PR #321, opened 2022-02-10
dask: `Data.__iter__`, `Data.__len__`

davidhassell:


cf/data/data.py
```diff
@@ -830,78 +830,75 @@ def __iter__(self):
...
+        Traceback (most recent call last):
+            ...
+        TypeError: len() of unsized object
 
         """
-        shape = self._shape
-        if shape:
-            return shape[0]
+        dx = self._get_dask()
+        if math.isnan(dx.size):
```
sadielbartholomew, line 897: `isnan` seems to be the only method we use from `math` across the Data module and to avoid this for full use of `numpy`/`dask` over `math`, perhaps with some performance benefit, can we not instead use (here and the other methods  which use it, e.g. `size`, `shape`, `nbytes`, etc.):

```suggestion
        if da.isnan(dx.size):
```
(else `np.isnan`)?

sadielbartholomew (commented):
All good. I've raised one question, so please see that before merging, and generally don't forget to mark migrated methods as 'daskified' via the decorator of that name to help us keep track of the migrated methods in the Data module itself (to complement the big table in #295); ideally add in those decorators on each PR, including this PR, before merging. Thanks.

cf/data/data.py
davidhassell: It turns out that dask itself seems very confused and uses math.isnan and np.isnan seemingly randomly for scalars (you need np.isnan for array checks) - I also just chose one at random:

```shell
$ grep isnan dask/array/slicing.py 
        if np.isnan(dim) and ind != slice(None, None, None):
    if math.isnan(other_numel) or other_numel == 0:
        if ind < 0 and not math.isnan(shape):
    if isinstance(ind, (np.ndarray, list)) and not math.isnan(shape):
        if math.isnan(dim):
    if np.isnan(dimension):
    if np.isnan(x.chunks[axis]).any():
        if not np.isnan(x.shape).any() and not np.isnan(index[0].shape).any():
                and not math.isnan(index_size)
            if np.isnan(index.size):
            if math.isnan(b) or b <= index.size:
        elif math.isnan(a):
$ 
```
A quick performance test shows that for scalars `math.isnan` is ~8x faster than np.isnan, and ~15x faster than da.isnan (which itself is just  a wrapper to np.isnan). So (by luck rather than judgement!) I think it probably is worth going with math.isnan. Is that OK?

sadielbartholomew (approved):
There is now a merge conflict to resolve, but otherwise this is all good to merge.

PR #322, opened 2022-02-11
dask: `Data_asreftime`; `Data._asdatetime`; `Data.year` & friends

davidhassell:


cf/data/data.py
```diff
@@ -8793,181 +8776,137 @@ def cyclic(self, axes=None, iscyclic=True):
...
+        return YMDhms(self, "minute")
 
     @property
     def second(self):
-        """The second of each data array element.
+        """The second of each date-time value.
 
-        Only applicable for reference time units.
+        Only applicable for data with reference time units. The
+        returnedx `Data` will have the same mask hardness as the
```
sadielbartholomew, line 8894:
```suggestion
        returned `Data` will have the same mask hardness as the
```

cf/test/test_Data.py
```diff
@@ -2015,6 +2008,10 @@ def test_Data_year_month_day_hour_minute_second(self):
         self.assertTrue(d.minute.equals(cf.Data([[37, 25]])))
         self.assertTrue(d.second.equals(cf.Data([[26, 26]])))
 
+        # Can't get year from with non-reference time units
```
sadielbartholomew, line 2011:
```suggestion
        # Can't get year from data with non-reference time units
```

sadielbartholomew (approved):
Fantastic. All good, save some typos and possibly one minor aspect as raised in-line. Thanks again David, please merge when ready.

davidhassell:
Hi Sadie, thanks for the review. I found that the @daskified decorator didn't work with methods decorated with `@property`:

```python
# Not @daskified
>>> cf.Data([1, 2], 'days since 2000-01-01').second
<CF Data(2): [0, 0]>

# @daskified
>>> cf.Data([1, 2], 'days since 2000-01-01').second
<bound method daskified.<locals>.decorator.<locals>.wrapper of <CF Data(2): [2000-01-02 00:00:00, 2000-01-03 00:00:00]>>
```

so I'll leave that for now.

I'm inclined to leave the numpy imports :) I find it useful to for everything to be defined for the reader. Happy to be persuaded otherwise, though!

davidhassell:
Whilst testing the decorator, I found that initialising a Data object with a size 1 datetime wasn't working. I'll put in a fix as another commit on this PR:

```python
>>> cf.Data([1, 2], 'days since 2000-01-01')
<CF Data(2): [2000-01-02 00:00:00, 2000-01-03 00:00:00]>
>>> cf.Data([1], 'days since 2000-01-01')
traceback
    ...
AttributeError: 'cftime._cftime.DatetimeGregorian' object has no attribute 'dtype'
```

davidhassell:
... OK, it's not in the __repr__ that the problem lies, and that has (probably) been fixed in another PR that I've got in the wings. So I'll merge this now, afterall.

sadielbartholomew:
Hi David, that's all fine to me!

Regarding:

> I found that the @daskified decorator didn't work with methods decorated with @property

I guess you tried to add it *above* the `property` decorators? I realised when creating `@daskified` that this order wouldn't work, so added a note to the docstring, which covers how it needs to be done in these cases:

> Note: for properties the decorator must be placed underneath the property decorator so it is called before and not after it.

So hopefully just popping it below will work. To help us keep track I'll try adding those in now as a follow-up commit.

davidhassell:
Ah ha! Thanks for showing me how to decorate the properties.

PR #324, opened 2022-02-17
Update deprecated `distutils.version` to `packaging.version`

sadielbartholomew:
When running test modules e.g. `test_Data` under `pytest` (as can be done via `$ pytest <test module>` even though we don't strictly use `pytest` but `unittest`), I get multiple deprecation warnings on several uses of `distutils.version.LooseVersion`, all in `cf/__init__.py`:

```console
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    g["version"][version] = LooseVersion(version)
```

I have updated these here to `packaging.version.Version`; it actually wasn't at all clear in any of the relevant documentation as to which class from that module would be the appropriate replacement, but from doing a full-GitHub search on relevant keywords I found the following comments ([this](https://github.com/pydata/xarray/pull/6096#issue-1086346755) and [this](https://github.com/pydata/xarray/pull/6096#issuecomment-1000339097)) and corresponding PR to xarray, along with a few other similar PRs on major libraries, to confirm that `Version` is the best one to go with.

Unfortunately `packaging` isn't a built-in, so this change prompted by the deprecation warning does necessitate another dependency, which I have added to the requirements list and documentation of those.

@davidhassell can I check with you that it is OK to add this new dependency? (I am not too pleased that an entire new dependency is required to upgrade deprecated API usage, but this is what the warning prompts us to use. If you can think of an alternative approach, do let me know...)

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -7,6 +7,7 @@ version 3.13.0
   ``um`` keyword has been set, but without the ``'version'`` key
   (https://github.com/NCAS-CMS/cf-python/issues/306)
 * Changed dependency: ``1.9.0.2<=cfdm<1.9.1.0``
+* New dependency: ``packaging>=20.0``
 
 ----
 
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -96,13 +96,13 @@ except ImportError as error1:
 
 __cf_version__ = cfdm.core.__cf_version__
 
-from distutils.version import LooseVersion
+from packaging.version import Version
 import importlib.util
 import platform
 
 # Check the version of Python
 _minimum_vn = "3.7.0"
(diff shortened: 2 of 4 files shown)


davidhassell (approved):
All looks good. Thanks, Sadie.

I agree with you that we have to accept this new import. I see that it is even recommended by [PEP 632](https://www.python.org/dev/peps/pep-0632/), which I can hope means that it will soon be part of the standard library ...?

sadielbartholomew:
> I see that it is even recommended by [PEP 632](https://www.python.org/dev/peps/pep-0632/), which I can hope means that it will soon be part of the standard library ...?

I'm not sure whether it will either (it was hard to find anything concrete on that from a quick-ish search) but I hope so too! I guess we'll find out eventually...

PR #331, opened 2022-03-01
dask: `Data.halo`

davidhassell:


sadielbartholomew (approved):
Excellent. There's one reference to the deprecated parameter name that needs updating, as raised in-line, but then please merge.

PR #333, opened 2022-03-01
dask: `Data.flatten`

davidhassell:


cf/data/data.py
```diff
@@ -10078,67 +10074,52 @@ def flatten(self, axes=None, inplace=False):
...
+        d = _inplace_enabled_define_and_cleanup(self)
 
-        ndim = self._ndim
+        ndim = d.ndim
         if not ndim:
             if axes or axes == 0:
                 raise ValueError(
-                    "Can't flatten: Can't remove an axis from "
-                    "scalar {}".format(self.__class__.__name__)
+                    "Can't flatten: Can't remove an axes from "
```
sadielbartholomew, line 10083: (Feels a bit pedantic to mention, but since you changed it from one to the other here) surely the original, `axis`, is in fact the right one?

sadielbartholomew (approved):
A few minor comments, but all good to go, assuming `TODODASK - check against daask flatten behaviour` has been covered (as per the removed line)?

cf/data/data.py
davidhassell: Not at all. Fixed as: `"Can't flatten: Can't remove axes from "`

davidhassell:
Thanks Sadie - you're right that the `TODODASK - check against dask flatten behaviour` has been covered. It behaves like numpy in that it doesn't allow subsets of axes to be flattened, so that code that does that (the transposing and reshaping) is still there in `Data.flatten`.

PR #338, opened 2022-03-04
dask: `Data.HDF_chunks`

davidhassell:
No dask implications here, just tidying up something that never really got finished from way back.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -498,13 +498,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
 
         super().__init__(array=array, fill_value=fill_value, _use_array=False)
 
-        # Create the _HDF_chunks attribute: defines HDF chunking when
-        # writing to disk.
-        #
-        # Never change the value of the _HDF_chunks attribute
-        # in-place.
-        self._HDF_chunks = None
-
         if loadd is not None:
             self.loadd(loadd)
             return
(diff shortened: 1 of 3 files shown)


davidhassell:
Hi Sadie, I've just noticed that we have `nc_clear_hdf5_chunksizes`, `nc_hdf5_chunksizes`, and `nc_set_hdf5_chunksizes` inherited from cfdm, which might make `HDF_chunks` redundant. Please don't look at this yet while I see if that's the case. Thanks.

davidhassell:
OK - using the cfdm methods is the way to go, as it's only `cfdm.NetCDFWrite` that knows how to dealy with the HDF chunk sizes.

davidhassell:
I've left some uses of `HDF_chunks` in, as they are in LAMA-based serialisation methods that are not daskified yet and will need a complete overhaul anyway.

cf/data/data.py
```diff
@@ -10965,41 +10941,85 @@ def flip(self, axes=None, inplace=False, i=False):
...
-            self._HDF_chunks = None
-            return org_HDF_chunks
+            chunks: `dict` or `None`, *optional*
+                Specify HDF chunk sizes.
 
-        axes = self._axes
-        for axis, size in chunks.items():
-            _HDF_chunks[axes[axis]] = size
+                When no positional argument is provided, the HDF chunk
+                sizes are unchanged
```
sadielbartholomew, line 10964:
```suggestion
                sizes are unchanged.
```

sadielbartholomew (approved):
All good (and very dedicated of you to have added new docstring examples to a deprecated method)!

cf/data/data.py
davidhassell: ... likewise for typos in deprecated methods. Someone somewhere might will be glad for our efforts!

davidhassell:
Just checking - is good to merge? Thanks.

sadielbartholomew:
> Just checking - is good to merge? Thanks.

Yes, please go ahead.

PR #339, opened 2022-03-04
dask: `Data.argmax`

davidhassell:


cf/data/data.py
```diff
@@ -7542,98 +7542,69 @@ def argmax(self, axis=None, unravel=False):
...
+        >>> index = d.argmax(unravel=True)
+        >>> index
+        (1, 2)
+        >>> d[index]
+        <CF Data(1, 1): [[5]]>
 
-            return np.ravel_multi_index(index, self.shape)
+        >>> d.argmax(axis=0)
+        <CF Data(3): [1, 1, 1]>
+        d.argmax(axis=1)
```
sadielbartholomew, line 7590:
```suggestion
        >>> d.argmax(axis=1)
```

cf/data/data.py
```diff
@@ -7542,98 +7542,69 @@ def argmax(self, axis=None, unravel=False):
...
-        # Parse axis
-        ndim = self._ndim
-        if -ndim - 1 <= axis < 0:
-            axis += ndim + 1
-        elif not 0 <= axis <= ndim:
-            raise ValueError(
-                "Can't argmax: Invalid axis specification: Expected "
-                "-{0}<=axis<{0}, got axis={1}".format(ndim, axis)
-            )
+        Only the first occurrence is returned.
```
sadielbartholomew, line 7593: I think this note should also be covered in the description somehow, rather than only (somewhat hidden) here in the examples. Maybe adding a few words to the second sentence RE use without provision of an axis: "... locates the _first occurrence of the_ maximum of the whole data."

Also, if this stays, perhaps use a colon, or otherwise, to emphasise you are introducing/describing the example afterwards (I for one was a bit confused as to why the sentence was there on its own)?

```suggestion
        Only the first occurrence is returned:
```

sadielbartholomew (approved):
Some very minor comments registered, but generally all good. Nice, please merge!

cf/data/data.py
davidhassell: Thanks. How about 927aa0b54 ?

sadielbartholomew: Perfect, thanks. Marking as resolved.

PR #342, opened 2022-03-07
Fixes setting both relaxed_identities=True and ncvar_identities=True in `cf.aggregate`

### LINKED ISSUE #341: Setting both `relaxed_identities=True`and `ncvar_identities=True` in` cf.aggregate` doesn't work.
As of v3.12.0, `cf.aggregate` with `relaxed_identities=True` raises an obscure exception when `ncvar_identities=True` is also set. This is not necessary. The following should aggregate nicely:
```python
>>> import cf
>>> f = cf.example_field(0)
>>> cf.aggregate([f[:2], f[2:]], relaxed_identities=True, ncvar_identities=True) 
    ...
VaueError: 'relaxed' and 'nc_only' parameters cannot both be True
```

davidhassell:
Fixes #341

Fixes the issue, and a bit of code tidy thrown in that resulted from a red herring that I pursued before properly understanding the requirements.

### DIFF
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -214,7 +214,7 @@ class _Meta:
         self.has_data = f.has_data()
         self.identity = f.identity(
             strict=strict_identities,
-            relaxed=relaxed_identities,
+            relaxed=relaxed_identities and not ncvar_identities,
             nc_only=ncvar_identities,
             default=None,
         )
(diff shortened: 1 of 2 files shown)


sadielbartholomew (approved):
Fixes the issue at hand in a sensible way, adding a new test which exposes the original issue. The CI job failures are unrelated, due to a single flaky test which I have registered in a new issue (#348), and can be considered passes when that is taken into consideration.

Excellent! Merge at will.

PR #343, opened 2022-03-08
dask: `Dask cumsum`

davidhassell:
The `masked_as_zero` keyword has been deprecated because the person who requested it originally a) hasn't ever used it, b) says he doesn't need it and c) I think that its functionality was a bit confused. So, all in all I hope we can make this API change without any worries.


sadielbartholomew:
> The masked_as_zero keyword has been deprecated because the person who requested it originally a) hasn't ever used it, b) says he doesn't need it and c) I think that its functionality was a bit confused.

:disappointed: Sad to hear, but fair enough. Thanks for the context.

sadielbartholomew (approved):
All great, please merge.

PR #347, opened 2022-03-09
Cfa base

### LINKED ISSUE #346: Setting the CFA "base" option to an empty string in `cf.write` results in incorrect relative path names in the output `cfa_array` attribute
PR to fix this to follow.

davidhassell:
Fixes #346

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -6,6 +6,8 @@ version 3.13.0
 * Fix bug in `cf.read` when reading PP or fields files for which the
   ``um`` keyword has been set, but without the ``'version'`` key
   (https://github.com/NCAS-CMS/cf-python/issues/306)
+* Fix bug when setting the CFA "base" option to an empty string in
+  `cf.write` (https://github.com/NCAS-CMS/cf-python/issues/346)
 * Changed dependency: ``1.9.0.2<=cfdm<1.9.1.0``
 * New dependency: ``packaging>=20.0``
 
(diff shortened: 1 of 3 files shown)


sadielbartholomew (changes requested):
The fix looks sound, but I think the test (though promising and better than nothing, as-is) may need adjusting, because if I take this new `test_cfa.py` and try it on the `master` branch, `test_cfa_base` passes, implying the bug is not being exposed by it.

davidhassell:
Good point on the test. I realise that do it wholly properly we would need to move the original and cfa files to a new location and read them. I'll try and add that.

davidhassell:
How about: f4824d757? (I had to add a "u+w" chmod option so that `test_cfa.sh` didn't lose write permissions every time the test was run.)

sadielbartholomew:
> How about: https://github.com/NCAS-CMS/cf-python/commit/f4824d7571a6220ae71249dcf9adf117f9dec4dc? (I had to add a "u+w" chmod option so that test_cfa.sh didn't lose write permissions every time the test was run.)

Hmmm :thinking: , it's still passing for me when I copy both the `test_cfa.py` and `test_cfa.sh` over to `master`, including re-installing via `pip install -e .` to be certain it is running the `master`-state codebase. Is there maybe another subtlety being missed? If nothing comes to mind, you could also try these tests on `master` to confirm the results, as then I can check we are seeing the same outcome?

sadielbartholomew (approved):
After external discussions, we've agreed that this is good enough to merge despite the remaining deficiency that:

> if I take this new test_cfa.py and try it on the master branch, test_cfa_base passes, implying the bug is not being exposed by it

given the context (as part of code that won't be around for very much longer) and given that producing such a failure would be quite difficult. I am satisfied that the new unit test documents the issue at hand, which is better than nothing.

Thanks David, please merge.

PR #355, opened 2022-03-11
dask: `Data.chunks` and `Data.rechunk`

davidhassell:
Just docs and tests. Code changes were already there.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -3236,7 +3236,7 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
         balance=False,
         inplace=False,
     ):
-        """Change the chunk struture of the data.
+        """Change the chunk structure of the data.
 
         .. versionadded:: TODODASK
 
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -3226,6 +3226,7 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
 
         return d
 
+    @daskified(_DASKIFIED_VERBOSE)
     @_inplace_enabled(default=False)
     def rechunk(
         self,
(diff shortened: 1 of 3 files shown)


cf/data/data.py
```diff
@@ -3235,45 +3236,44 @@ def rechunk(
         balance=False,
         inplace=False,
     ):
-        """Convert blocks in the dask array for new chunks.
-
-        See `dask.array.rechunk`for more details.
+        """Change the chunk struture of the data.
```
sadielbartholomew, line 3239:
```suggestion
        """Change the chunk structure of the data.
```

cf/data/data.py
```diff
@@ -3235,45 +3236,44 @@ def rechunk(
...
 
-        .. versionadded:: 4.0.0
+        .. versionadded:: TODODASK
 
-        .. seealso:: `chunks`
+        .. seealso:: `chunks`, `dask.array.rechunk`
```
sadielbartholomew, line 3243: I hadn't thought about this before, but tried building a development version of the docs from this branch and such references all work seamlessly with intersphinx to link directly to relevant methods under the Dask docs, which is very neat. I guess you realised this - good thinking!

It might be useful to add similar 'see also' references to relevant Dask methods throughout the Data module where appropriate, if we aren't already doing so already (I've made a mental note to do so).

cf/data/data.py
```diff
@@ -3285,10 +3285,10 @@ def rechunk(
...
         >>> y = x.rechunk({0: 1000})
 
-        Use the value ``-1`` to specify that you want a single chunk along
-        a dimension or the value ``"auto"`` to specify that dask can
-        freely rechunk a dimension to attain blocks of a uniform block
-        size
+        Use the value ``-1`` to specify that you want a single chunk
+        along a dimension or the value ``"auto"`` to specify that dask
+        can freely rechunk a dimension to attain blocks of a uniform
+        block size
```
sadielbartholomew, line 3291:
```suggestion
        block size.
```

sadielbartholomew (approved):
All good, thanks!

> Code changes were already there.

I also did some manual testing on the code you migrated previously and it is watertight, as the new test helps to confirm.

Merge at will.

PR #362, opened 2022-03-22
`data` batch docstring re-formatting: 'Examples:' -> 'Examples'

sadielbartholomew:
As for NCAS-CMS#182 over on cfdm, but note that this is **only across the data module**, i.e:

```console
$ pwd
<path to cf repo>/cf-python/cf/data
$ find . -type f | xargs sed -i 's/Examples:/Examples/g'
```

in line with our approach to managing development of mostly `data` modules on `lama-to-dask` and all other modules on `master`. I am about to open a corresponding PR on `master` to cover all non-`data` modules.

@davidhassell I have opened this PR so the batch update gets done at the first available point, but please prioritise merging of your currently-open PRs before this - it is trivial to manage code updates and to resolve merge conflicts here, much less so (probably) on your PRs.

### DIFF
--- a/cf/data/abstract/compressedsubarray.py
+++ b/cf/data/abstract/compressedsubarray.py
@@ -78,7 +78,7 @@ class CompressedSubarray(abc.ABC):
         """The file on disk which contains the compressed array, or
         `None` of the array is in memory.
 
-        **Examples:**
+        **Examples**
 
          >>> self.file
          '/home/foo/bar.nc'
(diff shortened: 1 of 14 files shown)


sadielbartholomew:
> Great. Can we just merge this now and I'll check for conflicts in the PRs that are currently up?

Sure, whatever you prefer - I just didn't want to create merge difficulties for you, though `git` may handle it neatly (I suspect it will but I'm not certain)!

PR #366, opened 2022-03-30
dask: `Data.__hash__`

davidhassell:
Hi Sadie, I think deprecating `__hash__` is the good thing to do because it was never used in the library code; it's exact functionality is still in `cf.hash_array`; and it was never a good hash of a `Data` object, anyway, as it it didn't include units, cyclic axes, etc.

### DIFF
--- a/cf/data/mixin/deprecations.py
+++ b/cf/data/mixin/deprecations.py
@@ -63,7 +63,8 @@ class DataClassDeprecationsMixin:
         _DEPRECATION_ERROR_METHOD(
             self,
             "__hash__",
-            message="Consider using 'cf.hash_array' instead.",
+            message="Consider using 'cf.hash_array' on the underlying "
+            "array instead.",
             version="TODODASK",
             removed_at="5.0.0",
         )
--- a/cf/data/mixin/deprecations.py
+++ b/cf/data/mixin/deprecations.py
@@ -10,7 +10,7 @@ class DataClassDeprecationsMixin:
     def __hash__(self):
         """The built-in function `hash`.
 
-        Depreacted at version TODODASK. Consider using the
+        Deprecated at version TODODASK. Consider using the
         `cf.hash_array` function instead.
 
         Generating the hash temporarily realizes the entire array in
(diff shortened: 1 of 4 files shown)


cf/data/mixin/deprecations.py
```diff
@@ -7,6 +7,67 @@
...
+        >>> print(d.array)
+        [0.0 -- 2.0 3.0]
+        >>> hash(d)
+        -4816859207969696442
+
+        """
+        _DEPRECATION_ERROR_METHOD(
+            self,
+            "__hash__",
+            message="Consider using 'cf.hash_array' instead.",
```
sadielbartholomew, line 66: I just tried this, following this instruction assuming `cf.hash_array(d)` would work (for a `cf.Data` object `d`), but that runs into `AttributeError: 'Data' object has no attribute 'flags'`, whereas `cf.hash_array(d.array)` works as it should, so perhaps it would be helpful top be more specific and say something along the lines of "Consider using 'cf.hash_array' **on the underlying array** instead."?

cf/functions.py
```diff
@@ -2611,80 +2609,76 @@ def pathjoin(path1, path2):
...
 
-    The hash value is not guaranteed to be portable across versions of
-    Python, numpy and cf.
-
     :Parameters:
 
         array: `numpy.ndarray`
             The numpy array to be hashed. May be a masked array.
 
+        algorthim: `hashlib` constructor function
```
sadielbartholomew, line 2625:
```suggestion
        algorithm: `hashlib` constructor function
```

cf/functions.py
```diff
@@ -2611,80 +2609,76 @@ def pathjoin(path1, path2):
...
+            Constructor function for the desired hash algorithm,
+            e.g. `hashlib.md5`, `hashlib.sha256`, etc.
+
+             .. versionadded:: TODODASK
```
sadielbartholomew, line 2629:
```suggestion
            .. versionadded:: TODODASK
```

sadielbartholomew (approved):
Minor comments raised in-line for your consideration pre-merge, but this looks solid overall and I agree that deprecating `__hash__` seems sensible given your justifications.

cf/data/mixin/deprecations.py
davidhassell: Good idea. Done.

PR #368, opened 2022-03-31
dask: `Data` query interface

davidhassell:
Just deleting these methods, as they merely duplicate code in `cf.Query` (https://github.com/NCAS-CMS/cf-python/blob/v3.12.0/cf/query.py#L667-L710). That code is already as efficient as it can be for dask arrays.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -4272,25 +4272,6 @@ class Data(Container, cfdm.Data, DataClassDeprecationsMixin):
 
             return self
 
-    def __query_set__(self, values):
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
> Just deleting these methods, as they merely duplicate code

Very sensible!

I do notice there are still such methods in

davidhassell:
> I do notice there are still such methods in mixin.propertiesdata but as long as that's known and/or OK, please merge!

This was not known! These other methods can be scrapped to - I'll put in another commit that deletes them before merging.

PR #369, opened 2022-03-31
dask: `Data._parse indices`

davidhassell:
`_parse_indices` inherited from cfdm has been replaced with `cf.parse_indices` (because the functionality is also used elsewhere), and so the former should not be used.


sadielbartholomew (approved):
Very much does what it says on the tin! Thanks, please merge.

PR #370, opened 2022-04-04
dask: `Data.__round__`

davidhassell:
`__round__` is not implemented by numpy nor dask - I reckon it's OK to also not implement it here.

### DIFF
--- a/cf/data/mixin/deprecations.py
+++ b/cf/data/mixin/deprecations.py
@@ -7,6 +7,22 @@ from ...functions import (
 class DataClassDeprecationsMixin:
     """Deprecated attributes and methods for the Data class."""
 
+    def __round__(self, *ndigits):
+        """Called to implement the built-in function `round`
+
+        Deprecated at version TODODASK, use method `round` instead.
+
+        x.__round__(*ndigits) <==> round(x, *ndigits)
+
+        """
+        _DEPRECATION_ERROR_METHOD(
+            self,
+            "__round__",
+            "Use method 'round' instead.",
+            version="TODODASK",
+            removed_at="5.0.0",
+        )  # pragma: no cover
+
     @property
     def Data(self):
         """Deprecated at version 3.0.0, use attribute `data` instead."""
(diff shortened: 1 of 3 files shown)


cf/data/mixin/deprecations.py
```diff
@@ -69,6 +69,22 @@ def __hash__(self):
...
+
+        Deprecated at version TODODASK, use method `round` instead.
+
+        x.__round__(*ndigits) <==> round(x, *ndigits)
+      
+        """
+        _DEPRECATION_ERROR_METHOD(
+            self,
+            "__round__",
+            "Use method 'round' instead.",
```
sadielbartholomew, line 83: I don't think this is an appropriate message, since the `round` method with `cf.Data` ultimately just calls `Data.__round__`? Therefore I suspect the best message to provide in context for the deprecation is something along the lines of that raised when you try to do the equivalent with `numpy`, namely `"cf.Data doesn't define a __round__ method"`?:

```python
>>> n = np.arange(120).reshape(2, 6, 10)
>>> round(n, 5)
  ...
TypeError: type numpy.ndarray doesn't define __round__ method
>>> d = cf.Data(n)
>>> round(d, 5)
  ...
cf.functions.DeprecationError: Data method '__round__' has been deprecated at version TODODASK and is no longer available and will be removed at version 5.0.0. Use method 'round' instead.
```

sadielbartholomew (changes requested):
One issue, as raised in-line, otherwise this looks good to me.

cf/data/mixin/deprecations.py
davidhassell: Hi Sadie - sorted: https://github.com/NCAS-CMS/cf-python/pull/370/commits/eff87caa1a99f1356ba10e2bd80d7731ee9febaf

sadielbartholomew: Removing the method is indeed a simpler way to produce that message, I suppose! Thanks, marking as resolved.

davidhassell:
Hi Sadie - just checking if this is good to merge, now?

sadielbartholomew:
> Hi Sadie - just checking if this is good to merge, now?

Yes, good to go in my eyes. Thanks.

sadielbartholomew (approved):

PR #373, opened 2022-04-05
dask: `Data.all` and `Data.any`

davidhassell:


sadielbartholomew (commented):
The migration implementation is sound, and the tests have been updated nicely, though before we merge this I think we should discuss the question you pose here in the notes under 'Returned Booleans', because I am not convinced about some cases (see in-line comments) for returning `Data` as opposed to the `Boolean` in these methods, and would like to hear more about the motivation.

sadielbartholomew (approved):
@davidhassell apologies for the delay in reporting back, this fell off my radar after submitting my review. After our discussions last week, my concerns about the return type were quashed, especially from:

* it being pointed out that `numpy` returns its own custom type for a Boolean, rather than a Python built-in `bool`, so there is an appropriate consistency here as you originally saw and implemented:

    ```python
    >>> import numpy as np
    >>> import cf
    >>> n = np.array([0, 1, 2, 3])
    >>> type(n.all())
    <class 'numpy.bool_'>
    >>> type(n.any())
    <class 'numpy.bool_'>
    >>> c = cf.Data(n)
    >>> type(c.any())
    <class 'cf.data.data.Data'>
    >>> type(c.all())
    <class 'cf.data.data.Data'>
    >>> bool(c.any())
    True
    >>> bool(c.all())
    False
    ```

* and from the reference to relevant parts of the array API in https://github.com/NCAS-CMS/cf-python/pull/373#discussion_r848206491 which I hadn't properly looked at.

So, overall, I now agree with your behaviour choice and have marked my previously opened comments on the matter as resolved. All good, please merge.

PR #374, opened 2022-04-05
dask: `Data.apply_masking`

davidhassell:


cf/test/test_Data.py
```diff
@@ -492,9 +492,6 @@ def test_Data_halo(self):
             d.halo(4)
 
     def test_Data_mask(self):
-        if self.test_only and inspect.stack()[0][3] not in self.test_only:
-            return
-
```
sadielbartholomew, line 497: This is an appropriate time to consider my `TODODASK` note just under this line, namely that it may make sense, given the similarities in the methods, to combine `test_Data_apply_masking` and `test_Data_mask`. What do you think? I'm happy to either combine or not, so please either go ahead and do that, else just remove the `TODODASK` comment lines just below. Thanks.

cf/data/data.py
```diff
@@ -7412,45 +7411,45 @@ def apply_masking(
...
                     "'fill_values' parameter must be a sequence or "
-                    "of type bool. Got type {}".format(type(fill_values))
+                    f"of type bool. Got type {type(fill_values)}"
                 )
             else:
                 if isinstance(fill_values, str):
                     raise TypeError(
                         "'fill_values' parameter must be a sequence or "
-                        "of type bool. Got type {}".format(type(fill_values))
+                        f"of type bool. Got type {type(fill_values)}"
```
sadielbartholomew, line 7424: Perhaps we can also prevent the duplication of this message whilst we are here, by assigning it to a variable just before the `try/except`?

sadielbartholomew (approved):
I've left two suggestions that are unrelated to the core of the PR, but that we may want to include here. All good, though, so please merge at will.

cf/data/data.py
davidhassell: Given that almost all uses will not raise an exception, I'm slightly against creating the message which almost always never get used.

cf/test/test_Data.py
davidhassell: Perhaps go with the easy option of not combining at this time - so I'll just remove the message!

cf/data/data.py
sadielbartholomew: Fair enough! I appreciate your justifications. Marking as resolved.

PR #376, opened 2022-04-06
dask: `Data.data`

davidhassell:


cf/test/test_Data.py
```diff
@@ -3938,6 +3938,10 @@ def test_Data_tolist(self):
             self.assertEqual(e, np.array(x).tolist())
             self.assertTrue(d.equals(cf.Data(e)))
 
+    def test_Data_data(self):
+        d = cf.Data([1, 2], "m")
+        self.assertIs(d.data, d)
```
sadielbartholomew, line 3943: I'd be happier if, to be safe, we can add some further illustrative cases covering the various possibilities for `Data` objects (mainly `Data` with a mask and `Data` chunked in a non-default way) to this check. How about something like:

```suggestion
        for d in [
            cf.Data(1),
            cf.Data([1, 2], fill_value=0),
            cf.Data([1, 2], "m"),
            cf.Data([1, 2], mask=[1,0], units="m"),
            cf.Data([[0, 1, 2], [3, 4, 5]], chunks=2),
        ]:
            self.assertIs(d.data, d)
```
?

sadielbartholomew (commented):
One suggestion, but overall this is all good so please merge when ready in light of that comment.

PR #380, opened 2022-04-12
dask: `Data.mask_fpe`

davidhassell:


davidhassell:
@sadielbartholomew  - just a note that this PR may affect the daskification of `_binary_arithmetic`.

cf/data/mixin/deprecations.py
```diff
@@ -332,6 +333,80 @@ def add_partitions(self, extra_boundaries, pdim):
...
+        >>> e/d
+        <CF Data: [inf, 2.0] >
+        >>> e**123456
+        <CF Data: [1.0, inf] >
+
+        """
+        raise DeprecationError(
+            "Data method 'mask_fpe' has been deprecated at version TODODASK "
+            "and is not available.\n\n"
+            "It is currently not possible to control how floating-point errors are handled, due to the use of `dask` for handling all array manipulations. This may change in the future (see https://github.com/dask/dask/issues/3245 for more details)."
```
sadielbartholomew, line 407: The `black` formatter shouldn't be happy with this line length and yet it passes locally for me - strange! Regardless, if we can break it down, that would be great, thanks. For ease of committing:

```suggestion
            "It is currently not possible to control how floating-point errors "
            "are handled, due to the use of `dask` for handling all array "
            "manipulations. This may change in the future (see "
            "https://github.com/dask/dask/issues/3245 for more details)."
```

sadielbartholomew (approved):
Overall:

I have raised a few minor comments but otherwise this is ready to merge, with +1 for a clear and detailed deprecation message.

Notes:

`flake8` uncovers that that there is still a reference to `_mask_fpe`, but as you have implied it is in `_binary_operation` which I am covering (PR to go up today or tomorrow) so not touching it here to avoid the merge conflict is wise. I've already removed the reference to it in my branch for the PR.

There is one other undefined name though, maybe we can deal with it here whilst we are here? It's in `get_filenames` which is marked as migrated in the big table:


```console
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

cf/data/data.py:4181:20: F821 undefined name '_mask_fpe'
cf/data/data.py:9406:30: F821 undefined name 'FileArray'

```

davidhassell:
Thanks, Sadie. The `FileArray` is annoying, and wholly my fault, but it has now been resolved in other PRs. (Also, I've spotted some holes in `get_filenames` which need to be revisited ...)

PR #381, opened 2022-04-12
`data` batch docstring re-formatting: 'Examples:' -> 'Examples'

sadielbartholomew:
Re-apply #362 (see details there), which we reverted because the merge, at the time, caused some unanticipated merge conflicts that it was simplest to avoid. After #356 has been merged, which should be imminent, will be a convenient time to make this update. (Then there will be no need to update the docstrings manually in this respect.)

~~**Note: do not merge until #356 has been merged, since that is a large PR in terms of line changes and code movement (any conflicts emerging on other open Dask migration PRs due to this, if any, will be minor, so they don't need to by considered).**~~ (see comment below RE this)

### DIFF
--- a/cf/data/abstract/compressedsubarray.py
+++ b/cf/data/abstract/compressedsubarray.py
@@ -78,7 +78,7 @@ class CompressedSubarray(abc.ABC):
         """The file on disk which contains the compressed array, or
         `None` of the array is in memory.
 
-        **Examples:**
+        **Examples**
 
          >>> self.file
          '/home/foo/bar.nc'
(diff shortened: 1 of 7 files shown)


sadielbartholomew:
In the end I have re-applied the `sed` batch replacement on the `lama-to-dask` branch as updated to the present state, since that was easier than resolving the corresponding merge conflicts. I have double-checked that the line count of the changes is the same as the number of `Example:` cases under `data/` and that the tests still pass, to ensure I am not touching anything I shouldn't be for this scope. Good to merge now.

davidhassell:
Great!

PR #382, opened 2022-04-19
dask: `Data.close`

davidhassell:


cf/data/mixin/deprecations.py
```diff
@@ -155,6 +155,25 @@ def ispartitioned(self):
...
+        """
+        _DEPRECATION_ERROR_METHOD(
+            self,
+            "close",
+            "All files are now automatically closed when not being accessed.",
+            version="TODODASK",
+            removed_at="5.0.0",
+        )  # pragma: no cover
+
+        print("TODODASK - is this still needed/valid? Not needed")
```
sadielbartholomew, line 173: It's not clear whether this (first) question remains open, since you seem to have answered it yourself already in the next sentence? Just checking there is nothing left to do here, to tidy up or otherwise...

sadielbartholomew (approved):
One question, but if it doesn't indicate any tweaks that may be required, this is all good, so please merge.

cf/data/mixin/deprecations.py
davidhassell: Good spot - that deprecated code has now been deleted. Merging ...

PR #383, opened 2022-04-19
dask: `Data.compressed`

davidhassell:


sadielbartholomew (approved):
Excellent: an elegant dask solution with robust updates to the unit test. There's one test line that seems superfluous, but this is good to merge.

PR #388, opened 2022-04-20
dask: `Data.to_dask_array`

davidhassell:
The (undocumented) API seems to be `to_dask_array()`, e.g. (from `dask.array.asarray`) https://github.com/dask/dask/blob/main/dask/array/core.py#L4343

#### Need to think about when to merge this, as it could break other items ...


sadielbartholomew (commented):
All good except that three cases of (`_`)`get_dask` seem to remain (as indicated by one test failure in `test_Data_reshape`), though I suspect they were re-introduced by the latest merge commit rather than missed by you when you did this. Once you blitz those then I am happy for this to be merged:

```console
$ pwd
/home/sadie/cf-python/cf
$ git grep "get_dask"
data/data.py:    #        return self.get_dask(copy=True)
data/data.py:        dx = self._get_dask()
data/data.py:        dx = d._get_dask()
data/utils.py:        return a.data.get_dask(copy=False)
```

As indicated in our discussion last week, I agree with the API-related decision to standardise to use `get_dask_array` and not have those extra methods, where the loss of `get`, `set`, `del` symmetry in this case is justified.

davidhassell:
Thanks, Sadie. Superceded `_get_dask` occurrences removed. Merging.

PR #391, opened 2022-04-25
dask: `Data.unique`

davidhassell:


cf/data/data.py
```diff
@@ -7995,46 +7995,59 @@ def uncompress(self, inplace=False):
...
-                u.append(array.item())
+        """
+        d = self.copy()
+        hardmask = d.hardmask
+        if hardmask:
+            # Soften a hardmask so that the result doesn't contain a
+            # seperate missing value for each input chunk that
+            # contains missing values. For any number greater than 0
+            # of missing values in the original data, we only want one
+            # missing value in the result.
```
sadielbartholomew, line 8040: Very clear and useful comment :100:

sadielbartholomew (approved):
All good and ready to merge, with a nice extension to the unit test, though if I was to be harsh I might suggest first including also in the unit test a small block to cover the various data type possibilities (`"fibUS"` as set here) and in particular the case of a string-type data array, an example I checked manually in reviewing, since really we are only checking integer arrays at present?).

cf/data/data.py
sadielbartholomew: I should add that great commenting is generally the case with your Dask migration PRs, but I will mention it now and again for nice examples :)

davidhassell:
Harsh is fine! Data type unit tests:  https://github.com/NCAS-CMS/cf-python/pull/391/commits/ae13c208c10ab8c7156fbe3866f958fb249e860f

sadielbartholomew (approved):
Bonus points for adding the new commit, which I have sanity checked and has added my suggestion in an elegant way. All good to merge.

davidhassell:
**For future reference** - the `da.unique` function was not used in this PR, we used `da.reduction` instead, because the former does not deal very well with missing data. I had a look at modifying `da.unique` to cope, but a) it was not trivial; b) `cf.Data` doesn't need (yet) the `return_*` keywords; and c) the dask folk might change it to be a reduction at some stage (https://github.com/dask/dask/issues/2851). For these three reasons, it was preferable at this time to implement a missing-data-friendly "unique", without the `return_*` keywords as a standard `cf.data.Collapse` method.

PR #392, opened 2022-04-25
dask: Data dump and load serialisations

davidhassell:


sadielbartholomew (commented):
All good, though I had one question which may provoke further changes, as below.

* I see that the `loadd=None` and `loads=None` keywords remain, along with basic necessary management of those (`if loadd is not None:` statement and the equivalent), in the constructor method for the `Data` object. Is there a reason they have been left in, or can they now be cut too (I suspect they have been missed since they will throw an error as-is for this PR, I believe)?

davidhassell:
Good spot - those keywords and their associated code have gone: https://github.com/NCAS-CMS/cf-python/pull/392/commits/ed0683ddbac126ba3a816d4e4eafb4df5f5a5329

sadielbartholomew (approved):
I have sanity checked the new commit and this is ready to go. Thanks.

PR #394, opened 2022-04-27
dask: `Data.inspect`

davidhassell:


cf/test/test_Data.py
```diff
@@ -3845,6 +3846,12 @@ def test_Data_override_calendar(self):
 
         self.assertIsNone(d.override_calendar("all_leap", inplace=True))
 
+    def test_Data_inspect(self):
+        d = cf.Data([9], "m")
+        with open(os.devnull, "w") as devnull:
+            with contextlib.redirect_stdout(devnull):
+                self.assertIsNone(d.inspect())
```
sadielbartholomew, line 3853: It would be a more rigorous test if we compared against the expected string print output, please can we do that as well as just testing the return output (especially since that's the more critical aspect for this method's behaviour)? I think it can indeed be achieved via `io.StringIO` alongside your present use of that `contextlib` redirection method, as indicated for example in the first answer [here](https://stackoverflow.com/questions/56045623/how-to-capture-the-stdout-stderr-of-a-unittest-in-a-variable)?

sadielbartholomew (commented):
The daskification is all good, though I have one request regarding the new unit test. Thanks.

cf/test/test_Data.py
sadielbartholomew: > since the output contains pretty prints of lots of non-public-API stuff

Fair enough! That is indeed a good justification to not check against the precise output, which would be a bit messy anyway! How about, as a middle ground, we check for the presence of a few important keywords? For instance, that the `repr` gets included (for the header), and also check for the presence of some strings which should be there as keys, e.g. "fill_value" and "netcdf"?

Marking as resolved regardless, but the above would be nice as an extra!

davidhassell: Hi Sadie, thanks - I'll mark resolved as you suggest. I'm still not sure about pattern matching on the STDOUT, but lets keep that one going elsewhere if we want to :)

PR #395, opened 2022-04-27
dask: `Data.isscalar`

davidhassell:


cf/data/data.py
```diff
@@ -4920,18 +4920,19 @@ def is_masked(a):
         return bool(dx.any())
 
     @property
+    @daskified(_DASKIFIED_VERBOSE)
     def isscalar(self):
-        """True if the data array is a 0-d scalar array.
+        """True if the data is a 0-d scalar array.
```
sadielbartholomew, line 4925: Clear enough, though it might be useful if we clarify the situation with regards to non-numeric arrays. In fact, in reviewing I was wondering what happens for those and unsure if this current behaviour is an inconsistency that we wouldn't want (even though our result seems most intuitive to me on the face of it):

```python
In [30]: np.isscalar(np.array("A"))
Out[30]: False

In [31]: cf.Data("A").isscalar
Out[31]: True
```

(Could/should we perhaps defer to use `isscalar(x)` under-the-hood instead to ensure consistency?)

sadielbartholomew (approved):
All good. One question posed, so please merge if you don't think that calls for any changes, otherwise we can discuss the particulars.

cf/data/data.py
davidhassell: That's not nice! I see in `help(np.isscalar)` it says:

davidhassell:
```
    In most cases ``np.ndim(x) == 0`` should be used instead of this function,
    as that will also return true for 0d arrays. This is how numpy overloads
    functions in the style of the ``dx`` arguments to `gradient` and the ``bins`
```
so I'm tempted to deprecated `Data.isscalar` and suggest using "ndim == 0" instead. The method is only used once in `field.py` (https://github.com/NCAS-CMS/cf-python/blob/v3.12.0/cf/field.py#L11382)

sadielbartholomew: > so I'm tempted to deprecated Data.isscalar and suggest using "ndim == 0" instead. The method is only used once in field.py

Thanks for your interpretation. Sounds very wise to me, please go ahead with that either here or as a separate/follow-on PR.

sadielbartholomew: Perfect. Marking as resolved.

sadielbartholomew (approved):
All good to merge, thanks.

(In reviewing the final commit I did notice there is a format-string issue in the deprecation messaging which imagine was missed from a recent PR touching the deprecation module, but I will blitz a commit to fix that separately:
``DeprecationError: Data attribute 'isscalar' has been deprecated at version TODODASK and will be removed at version  and will be removed at version 5.0.0. Use 'd.ndim == 0' instead``)

PR #398, opened 2022-04-28
dask: `Data.outerproduct`

davidhassell:
Some minor changes to `reshape` crept in by accident.


cf/data/data.py
```diff
@@ -8746,22 +8747,21 @@ def floor(self, inplace=False, i=False):
...
 
         The axes of result will be the combined axes of the two input
-        arrays:
+        arrays.
 
-          >>> d.outerproduct(e).ndim == d.ndim + e.ndim
-          True
-          >>> d.outerproduct(e).shape == d.shape + e.shape
-          True
+        .. seealso :: `np.multiply.outer`
```
sadielbartholomew, line 8759:
```suggestion
        .. seealso:: `np.multiply.outer`
```

(else I don't think the Sphinx admonition will be recognised and render properly).

sadielbartholomew (approved):
Excellent, thanks. One minor formatting issue I believe, otherwise all good to merge.

PR #399, opened 2022-05-04
dask: `Data.hardmask`, `Data._set_dask`, `Data.to_dask_array`

davidhassell:
Work going on in the active storage world has made it undesirable to stack up dask operations during `Data.__init__` unless absolutely necessary.

Currently, after the data definition, `__init__` _always_ adds on a "harden/soften mask" operation, and in addition might add on an "astype" and/or a "where" operation.

The current thinking is that to implement active storage reductions, the dask graph can *only* contain the data definition. If the "astype" and "where" operations are needed then we don't want to use active storage anyway, but the hardness of the mask will not interfere with a storage-side reduction, so we don't want to add this operation by default, thereby making it very hard to tell if an active storage operation is possible, or not.

A solution is to "somewhat lazify" the setting of the mask hardness. Instead of the status quo of adding a dask graph operation every time `d.hardmask = X` is run, the `hardmask` attribute now just records the desired state of the mask hardness. It is then up to a future method to apply a mask hardening/softening operation **if and only if** it is necessary. At present, **only two methods** need this (`__setitem__` and `where`).

Not only does this pave the way for active storage reductions to be implemented, it also (in my opinion!) makes the code cleaner, too.

### DIFF
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -185,7 +185,7 @@ if not LooseVersion(_minimum_vn) <= _cfdm_version < LooseVersion(_maximum_vn):
     )
 
 # Check the version of dask
-_minimum_vn = "2019.03.0"
+_minimum_vn = "2022.03.0"
 if LooseVersion(dask.__version__) < LooseVersion(_minimum_vn):
     raise RuntimeError(
         f"Bad dask version: cf requires dask>={_minimum_vn}. "
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -185,7 +185,7 @@ if not LooseVersion(_minimum_vn) <= _cfdm_version < LooseVersion(_maximum_vn):
     )
 
 # Check the version of dask
-_minimum_vn = "2022.03.0"
+_minimum_vn = "2019.03.0"
 if LooseVersion(dask.__version__) < LooseVersion(_minimum_vn):
     raise RuntimeError(
         f"Bad dask version: cf requires dask>={_minimum_vn}. "
--- a/cf/data/dask_utils.py
+++ b/cf/data/dask_utils.py
@@ -14,8 +14,7 @@ from dask.core import flatten
 from ..cfdatetime import dt2rt, rt2dt
 from ..functions import atol as cf_atol
 from ..functions import rtol as cf_rtol
-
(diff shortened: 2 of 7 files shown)


sadielbartholomew (commented):
First of all I support the new approach underlying this PR, i.e:

> A solution is to "somewhat lazify" the setting of the mask hardness. Instead of the status quo of adding a dask graph operation every time d.hardmask = X is run, the hardmask attribute now just records the desired state of the mask hardness. It is then up to a future method to apply a mask hardening/softening operation if and only if it is necessary. At present, only two methods need this (__setitem__ and where).

and agree with your motivating conclusions:

> Not only does this pave the way for active storage reductions to be implemented, it also (in my opinion!) makes the code cleaner, too.

The code here is all good (no comments to make at all, in fact :100:) though your recent and final commit to synchronise with `lama-to-dask` here seems to have introduced (from your other recently-merged PRs) a few cases of `reset_mask_hardness` which need appropriate conversion, namely I am currently seeing:

```bash
ERROR: test_Data_masked_all (__main__.DataTest)
  File "test_Data.py", line 3912, in test_Data_masked_all
    d = cf.Data.masked_all(shape)
  File "/home/sadie/cf-python/cf/data/data.py", line 9348, in masked_all
    d._set_dask(dx, reset_mask_hardness=False)
TypeError: _set_dask() got an unexpected keyword argument 'reset_mask_hardness'

ERROR: test_Data_outerproduct (__main__.DataTest)
  File "test_Data.py", line 1431, in test_Data_outerproduct
    f = d.outerproduct(b)
  File "/home/sadie/cf-python/cf/data/data.py", line 96, in wrapper
  File "/home/sadie/cfdm/cfdm/decorators.py", line 44, in inplace_wrapper
  File "/home/sadie/cf-python/cf/decorators.py", line 62, in precede_with_kwarg_deprecation_check
  File "/home/sadie/cf-python/cf/data/data.py", line 8793, in outerproduct
    d._set_dask(dx, reset_mask_hardness=False)
TypeError: _set_dask() got an unexpected keyword argument 'reset_mask_hardness'

```

Once those are rectified this is ready to merge as far as I am concerned. Thanks.

davidhassell:
Thanks Sadie - conflict fixed, and merging ....

PR #401, opened 2022-05-09
dask: `Data.fits_in_memory`

davidhassell:


cf/data/data.py
```diff
@@ -10082,30 +10082,57 @@ def swapaxes(self, axis0, axis1, inplace=False, i=False):
...
+
+        Returns True if the size of the array with all delayed
+        operations computed, including a space for a full boolean
+        mask, is small enough to be retained in available memory.
+
+        The available memory is the actual free memory less the amount
+        reserved as a temporary work space, i.e. ``cf.free_memory() -
+        cf.fm_threshold()``
+
+        The delayed operations are actually not computed by
```
sadielbartholomew, line 10097: Perhaps this merits a `.. warning::` emphasis? (If you don't think so, it's fine as-is, though.)

cf/data/data.py
```diff
@@ -10082,30 +10082,57 @@ def swapaxes(self, axis0, axis1, inplace=False, i=False):
...
 
         **Examples**
 
-        >>> print(d.fits_in_memory(8))
+        >>> d = cf.Data(9, 'm')
+        >>> d.fits_in_memory()
+        True
+
+        >>> size = int(2 * cf.free_memory() / 8)
+        >>> d = cf.Data.empty((size,), dtype=float, chunks=-1)
```
sadielbartholomew, line 10122: Ah, that's a clever way to construct an array that doesn't fit - I will steal this technique for interactive experimentation! :100:

cf/data/data.py
```diff
@@ -10082,30 +10082,57 @@ def swapaxes(self, axis0, axis1, inplace=False, i=False):
...
+                Whether or not the computed array fits in memory.
 
         **Examples**
 
-        >>> print(d.fits_in_memory(8))
+        >>> d = cf.Data(9, 'm')
+        >>> d.fits_in_memory()
+        True
+
+        >>> size = int(2 * cf.free_memory() / 8)
```
sadielbartholomew, line 10121: It took me a while to clock that the insistence in dividing by eight (in the test as well as here) instead of simplifying to `/4` for these cases is to account for (as a guess that might be wrong?) the typical amount of bytes required per float for a `numpy.array` data structure, so perhaps a comment to clarify the relevance could be helpful to users? And/or there's the `nbytes` attribute which  you could call for these examples to show explicitly that the memory exceeds the available?

sadielbartholomew (approved):
All good and ready to merge, though I've made a few minor suggestions which you might want to incorporate so please consider those first.

cf/data/data.py
davidhassell: Thanks, Sadie - I've hopefully covered all of the good things you raise in https://github.com/NCAS-CMS/cf-python/pull/401/commits/a763e7d25a9a2c2a5f84ffc46e5f9ca70004c1dd

On reflection I've also removed the `fm_threshold` component. Given that dask essentially manages its own memory, I think that there's no point in us trying to micromanage it, too.

sadielbartholomew: Brilliant, all good thanks (and from this I noticed the "always including space for a full boolean mask" description which I missed before, apologies, which clarifies the situation at hand further!).

> On reflection I've also removed the fm_threshold component. Given that dask essentially manages its own memory, I think that there's no point in us trying to micromanage it, too.

Hear, hear!

I've sanity-checked that new commit and all feedback is addressed and everything is good to go, so marking this as resolved. Please merge.

PR #404, opened 2022-05-11
dask: `Data.get_data`

davidhassell:


sadielbartholomew (approved):
Perfect, please merge (CI jobs can be ignored as ever lately since they are failing or being subsequently cancelled on incompatible `cfdm` version availability, which I will look into separately shortly, but local tests pass).

PR #405, opened 2022-05-12
dask: `Data.Units`

davidhassell:
Code reorganisation for consistency. Also new unit test.

### DIFF
--- a/cf/data/dask_utils.py
+++ b/cf/data/dask_utils.py
@@ -617,7 +617,12 @@ def cf_units(a, from_units, to_units):
     :Returns:
 
         `numpy.ndarray`
-            An array containing values in the new units.
+            An array containing values in the new units. In order to
+            represent the new units, the returned data type may be
+            different from that of the input array. For instance, if
+            *a* has an integer data type, *from_units* are kilometres,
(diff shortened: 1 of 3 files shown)


cf/data/dask_utils.py
```diff
@@ -594,3 +593,40 @@ def cf_dt2rt(a, units):
...
+            The existing units of the array.
+
+        to_units: `Units`
+            The units that the array should be converted to. Must be
+            equivalent to *from_units*.
+
+    :Returns:
+
+        `numpy.ndarray`
+            An array containing values in the new units.
```
sadielbartholomew, line 620: It seems like the `dtype` can change (e.g. `int` to `float`) in this process, it might be worth noting that here (just to continue with the thoroughness of the documentation you're adding overall!)?

sadielbartholomew (approved):
Excellent. One very minor suggestion raised, but this is all good to merge so please go ahead when ready. Thanks.

cf/data/dask_utils.py
davidhassell: Good idea: https://github.com/NCAS-CMS/cf-python/pull/405/commits/cb617cc771f89cbd9f69107c583de3b357fb2f35

PR #407, opened 2022-06-13
dask: `Data.reconstruct_sectioned_data`

davidhassell:
This method was only ever used by `Field.regrid[s|c]`, but will not be when those methods are daskified. Therefore I think that it is OK to deprecate it at this time.


cf/test/test_Data.py
```diff
@@ -2493,27 +2493,6 @@ def test_Data_section(self):
...
-        if self.test_only and inspect.stack()[0][3] not in self.test_only:
-            return
-
-        d = cf.Data(ma)
-        self.assertEqual(d.count(), 284, d.count())
-        self.assertEqual(d.count_masked(), d.size - 284, d.count_masked())
-
-        d = cf.Data(a)
-        self.assertEqual(d.count(), d.size)
-        self.assertEqual(d.count_masked(), 0)
```
sadielbartholomew, line 2515: Did you mean to remove the `test_Data_count` method as well (asking since it doesn't come under the motivation for the PR)?

sadielbartholomew (approved):
One question raised, but otherwise all good! Thanks.

cf/test/test_Data.py
davidhassell: Oops - no I didn't! I'll reinstate it before I merge ...

Issue #365, opened 2022-03-28
How to keep track of filenames for field source?


ThibHlln:
Hi Sadie, Hi David, 🙂 

In [`unifhy`](https://github.com/unifhy-org/unifhy) we have been using [`cf.Field.get_filenames`](https://ncas-cms.github.io/cf-python/method/cf.Field.get_filenames.html?highlight=get_filenames#cf.Field.get_filenames) for a while to track down the source files of the user input fields so that they can be stored in a configuration file for potential later reuse. And I have only recently being faced with the scenario where *"If all of the data are in memory then an empty `set` is returned"*, meaning that `get_filenames` does not return the information we are looking for anymore (https://github.com/unifhy-org/unifhy/issues/80).

So, I am wondering:
- is there another attribute/property/method of `cf.Field` that always keeps the filenames of a field regardless of whether its data fits in memory?
- if not, would it make sense for `cf-python` to provide such functionality? e.g. not to drop the filenames even if the data fits in memory (I am guessing it wouldn't, otherwise you would already have implemented it 🙂)

Thank you in advance for your help,
Thibault

davidhassell:
Hi Thibault - good to hear from you. As you've guessed, it's complicated!

The short answer to your particular problem is perhaps  to manually save the file names straight after the read step:
```python
>>> import cf
>>> cf.write(cf.example_field(0), '~/delme.nc')
>>> f = cf.read('~/delme.nc')[0]
>>> f._custom['saved_filenames'] = f.get_filenames()
>>> f._custom['saved_filenames']
{'/home/david/delme.nc'}
```
Doing it this way, by adding it to the `_custom` dictionary as opposed to just setting the non-reserved attribute `f.saved_filenames = ...`, ensures that they'll get copied if you do `g = f.copy()`.

I'm can't think of a reason why we couldn't formalise this to, say:
```python
>>> f.get_filenames(save=True)
>>> f.saved_filenames  # now a reserved attribute
'/home/david/delme.nc'
```
This method with `save=True` would save the output of `get_filenames` (which could be an empty set) regardless of whether or not names had previously been saved. Would that be useful?

So what are the complications? As usual it's ambiguities and corner cases. If array values have been entirely overwritten (`f += 1`), then the presence of saved filenames could be misleading to some people, but not others. . Similarly if only some of the contributing files have been made "redundant". Note also that the files include files names which contain coordinates and other metadata - these usually, but not always, will be in the same files as the data ....

[...]

PR #408, opened 2022-06-13
dask: `Dask.get_filenames`

davidhassell:
Following on from https://github.com/NCAS-CMS/cf-python/issues/365, I think that this method is just too "dangerous", now, and it would be better to manage this from the `Field`/`Domain`, with content injected by `cf.read`/`cf.aggregate`. All of which would be the topic of a later PR.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -29,7 +29,6 @@ from ..functions import (
     _DEPRECATION_ERROR_KWARGS,
     _numpy_isclose,
     _section,
-    abspath,
     atol,
     default_netCDF_fillvals,
     free_memory,
@@ -39,7 +38,6 @@ from ..functions import (
 )
 from ..mixin_container import Container
 from ..units import Units
-from . import FileArray
 from .collapse import Collapse
 from .creation import compressed_to_dask, generate_axis_identifiers, to_dask
 from .dask_utils import (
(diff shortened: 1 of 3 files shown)


sadielbartholomew (commented):
All of the code changes here are fine and sensible, though it seems something has been missed, since the method doesn't seem to hit the deprecation error for the direct case, e.g:

```python
>>> import cf
>>> d = cf.Data(100, "m")
>>> d.get_filenames()
set()
```

which I assume is because there's still a case under the `mixin.propertiesdatabounds` module:

```console
$ pwd
/home/sadie/cf-python/cf
$ git grep "def get_filenames"
data/mixin/deprecations.py:    def get_filenames(self):
mixin/propertiesdatabounds.py:    def get_filenames(self):
```

So that probably needs to go or be deprecated also. Once that's done, this will be good to merge I think!

davidhassell:
Good spot - I didn't think to check that it actually worked :). It turns out that the issue is that the method is inherited from `cfdm.Data`, which is higher in the MRO than `DataClassDeprecationsMixin`. 

I have changed the MRO and everything seems OK (https://github.com/NCAS-CMS/cf-python/pull/408/commits/bc2bbce4c15854deecffc1bf4aaffe4860927bf5).

I'll make a note in #295 to pay attention to the `get_filenames` methods on other classes, which will need varying treatment (e.g. Field and Domain needs re-implementing, others (probably) deprecating), but that will be for another PR.

sadielbartholomew (approved):
Feedback addressed in an elegant way after:

> I have changed the MRO and everything seems OK (https://github.com/NCAS-CMS/cf-python/commit/bc2bbce4c15854deecffc1bf4aaffe4860927bf5).

All good, thanks. Merging.

PR #409, opened 2022-06-15
LAMA to Dask: arithmetic, logical & comparison operators

sadielbartholomew:
**Note**: not for review just yet (hence the draft PR form), notably since I am about to add a commit or two applying `elemwise` calls to the operations that give the `result`s, but need to do a little tweak on the code I have for that, so will push it up imminently.

For now, note that everything else here is good and ready for review as far as I am concerned, in particular the relevant test (`test_Data_BINARY_AND_UNARY_OPERATORS`) passes though I had to comment out two assertions which both fail due to `__pow__` calling `_combined_units` which raises `AttributeError: 'Data' object has no attribute '_size'` so probably needs to wait on `_combined_units` being migrated, which the big table #295 states is yet to happen (I can do it straight after this as a separate PR).


cf/data/data.py
```diff
@@ -3804,11 +3805,11 @@ def _binary_operation(self, other, method):
 
         """
         inplace = method[2] == "i"
-        method_type = method[-5:-2]
 
         # ------------------------------------------------------------
         # Ensure that other is an independent Data object
         # ------------------------------------------------------------
+        # TODODASK is this still needed? Comment below was left unfinished.
```
davidhassell, line 3812: I think that is still needed, so that combination with `cf.Query` objects works.

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+        # Parse units
+        data0, other, new_Units = data0._combined_units(other, method, True)
 
-        result = data0.copy()
-        result._shape = new_shape
-        result._ndim = new_ndim
-        result._size = new_size
-        result._axes = new_axes
+        # Cast as dask arrays
+        dx0 = data0._get_dask()
```
davidhassell, line 3843:
```suggestion
        dx0 = data0.to_dask_array()
```

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+        dx1 = other._get_dask()
```
davidhassell, line 3844:
```suggestion
        dx1 = other.to_dask_array()
```

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
-
-            array1 = other[indices].array
-
-            # UNRESOLVED ISSUE: array1 could be much larger than the
-            # chunk size.
-
-            if not inplace:
-                partition = partition_r
-                partition.update_inplace_from(partition_s)
+        try:
```
davidhassell, line 3854: I don't think we can trap FPEs anymore (https://github.com/NCAS-CMS/cf-python/pull/384#issuecomment-1102726424)

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+            if method == "__eq__":
+                result = _numpy_isclose(dx0, dx1, rtol=rtol, atol=atol)
```
davidhassell, line 3856: I'm think that for v4.0.0 we should think about not subverting `==` and `!=` with tolerance, and use our `isclose` in other parts of the code instead, in the interests of consistency with numpy, etc.

This would obvioulsy be an API change, so is not for consideration in this PR!

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
-            if broadcasting:
-                data0.partitions.set_location_map(new_axes)
-
-            self.__dict__ = data0.__dict__
+                )
+            else:
+                raise TypeError(error)
 
+        if inplace:  # in-place so concerns original self
+            self._set_dask(result, reset_mask_hardness=False)
```
davidhassell, line 3891:
```suggestion
            self._set_dask(result)
```

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+            self.override_units(new_Units, inplace=True)
             return self
+        else:  # not, so concerns a new Data object copied from self, data0
+            data0._set_dask(result, reset_mask_hardness=False)
```
davidhassell, line 3895:
```suggestion
            data0._set_dask(result)
```

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+        dx0 = data0._get_dask()
+        dx1 = other._get_dask()
 
-        # ------------------------------------------------------------
-        # Set the data-type of the result
-        # ------------------------------------------------------------
-        if method_type in ("_eq", "_ne", "_lt", "_le", "_gt", "_ge"):
-            new_dtype = np.dtype(bool)
+        # Set if applicable the tolerance levels for the result
+        if method in ("__eq__", "__ne__"):  # what about l/g-t-e (x4)?
```
davidhassell, line 3847: What does `l/g-t-e (x4)` mean?

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
-                if inplace:
-                    raise TypeError(
-                        "Incompatible result data-type ({0!r}) for "
-                        "in-place {1!r} arithmetic".format(
-                            np.result_type(array0.dtype, array1.dtype).name,
-                            array0.dtype.name,
-                        )
+                result = getattr(dx0, method)(dx1)
+
+        except FloatingPointError as error:
```
davidhassell, line 3866: This trapped `FloatingPointError` can go, it will never occur due to the dask laziness

cf/data/data.py
```diff
@@ -3831,304 +3831,70 @@ def _binary_operation(self, other, method):
...
+                # Redo the calculation ignoring the errors and
+                # then set invalid numbers to missing data
+                np.seterr(**_seterr_raise_to_ignore)
+                result = getattr(dx0, method)(dx1)
+                result = np.ma.masked_invalid(dx0, copy=False)
+                np.seterr(**_seterr)
+            else:
+                # Raise the floating point error exception
+                raise FloatingPointError(error)
+        except TypeError as error:
```
davidhassell, line 3878: I think that this exception can go, too, as `dask` handles these cases OK:
```
>>> x = da.from_array([1, 2, 3])
>>> y = da.from_array(3.14)
>>> x += y
>>> x
dask.array<add, shape=(3,), dtype=float64, chunksize=(3,), chunktype=numpy.ndarray>
```

davidhassell (commented):
Thanks, Sadie. A quick review, as I know that there is more work to done ...

> notably since I am about to add a commit or two applying elemwise calls to the operations that give the results

I don't really understand what these `elemwise` calls could be for - am I missing something?

cf/data/data.py
sadielbartholomew: Thanks for clarifying in this case as well as those in the other comments, which I will give a :+1: instead of writing the same thing! I'll either commit as suggestions or in new commit(s) to get these resolved.

cf/data/data.py
sadielbartholomew: Ah, good spot! This was a remnant of the early time I started this PR and I was due to update at the end in line with the 'new way'.

cf/data/data.py
davidhassell: Having consulted the cryptonomicon, I see now!

cf/data/data.py
sadielbartholomew: Oh, I thought not resetting the mask hardness would be appropriate here? Happy to make this change (in the two appropriate locations raised in these comments), but just wondering why this is preferred?

davidhassell: The practical reason is that `reset_mask_hardness` is no longer a keyword of `_set_dask`! See #399 for more details ...

cf/data/data.py
sadielbartholomew: Sorry, that was my very arcane way to try to convery `lt`, `gt`, `lt` and `gt` in slightly less characters :grimacing: In the original code we are assigning `rtol` and `atol` values for all of those:

https://github.com/NCAS-CMS/cf-python/blob/032eabd4ca2f56c7b52464e4eec89252cc48df3e/cf/data/data.py#L3516-L3519

but then we don't seem to use it in the inequality cases, so I don't think it is required there despite those methods being listed in the statement? Just to check I am not missing something here?

cf/data/data.py
sadielbartholomew: Good spot, thanks. Nice, simplifies the logic further! Will get these catches removed.

cf/data/data.py
davidhassell: I recalling working out that tolerant lt, le, ge, gt comparisons were not a good idea/didn't make sense. I guess setting rtol and atol in that if statement was just convenient.

cf/data/data.py
sadielbartholomew: That's a very good reason! Sorry, I do remember this - again it is something that I neglected to update given I started work on this branch months ago... I'll commit those now, thanks.

cf/data/data.py
sadielbartholomew: As discussed and agreed just now in person, we will indeed reserve (not) equals operations for strict equality checking, and use `isclose` for operations within tolerances (and our `isequals` will use `isclose` type comparisons).

I'll shortly put up a new issue to list some API changes we are planning, that we can add to as we decide on them, and I'll include this as one on the initial listing. Thanks!

sadielbartholomew:
Thanks for all of your feedback (so far) @davidhassell, which I have now addressed fully as far as I can tell.

Unless you have further comments, this should be ready to merge now, though I would like to also first check that I was OK to 'deprecate' our custom function `_numpy_isclose` by just removing it, rather than removing the functional code inside it and adding a deprecation warning, due to it being an internal function?

sadielbartholomew:
(CI jobs failing or cancelled due to trying to use the wrong `cfdm` version as a dependency, which I will fix in the workflow but outside of this PR; overall we don't have to worry about it here. Locally the relevant test, `test_Data.py`, passes.)

davidhassell:
Hi Sadie - all looks good to me, thanks. As `_numpy_isclose` is an underscore method, I'd be fine with just deleting it. I've confirmed that  the units tests pass on my laptop. Please merge!

PR #411, opened 2022-06-21
Dask isclose

davidhassell:


davidhassell:
Hi Sadie. I'm having all sort of linting issues, having upgraded `black`. I get different results when I run `pre-commit` compared with running `black` on its own, and it's all different to "before" ...

Anyway, the only interesting parts of this PR are `Data.isclose`, `Data._binary_operation`, `utils.conform_units`, `test_Data.py`. 

Sorry!

sadielbartholomew:
> Hi Sadie. I'm having all sort of linting issues, having upgraded black. I get different results when I run pre-commit compared with running black on its own, and it's all different to "before" ...

> Anyway, the only interesting parts of this PR are Data.isclose, Data._binary_operation, utils.conform_units, test_Data.py.

> Sorry!

Hi David, no worries, since I can upgrade the linting workflows and such in line with this, but sorry if it was a pain with regards to this PR. I'll take a look later today.

cf/data/data.py
```diff
@@ -3389,12 +3370,19 @@ def _binary_operation(self, other, method):
             atol = self._atol
 
         # ------------------------------------------------------------
-        # Perform the binary operation with data0 (self) and data1 (other)
+        # Perform the binary operation with data0 (self) and data1
+        # (other)
         # ------------------------------------------------------------
         if method == "__eq__":
-            result = da.isclose(dx0, dx1, rtol=rtol, atol=atol)
+            if dx0.dtype.kind in "US" or dx1.dtype.kind in "US":
```
sadielbartholomew, line 3377: Good spot and catch. Maybe we should update the test that covers the various binary operations, `test_Data_BINARY_AND_UNARY_OPERATORS`, to check some/more string type operations (not necessarily in this PR)? (I am already updating it a little to do some more complex unit checking to cover `_combined_units` (or raising an issue to register that this should be done at some point) as we discussed recently. Maybe I can do it as part of that update.)

sadielbartholomew (approved):
One probable (essentially pre-existing) typo to blitz but otherwise perfect, so please merge when ready. Thanks.

cf/data/data.py
davidhassell: Doing as part of that update sounds good. I wavered about putting the fix in to _binary_operator in this PR, but decided to go for it as it was convenient, and might have been the only update to this area.

cf/data/data.py
```diff
@@ -9216,30 +9196,32 @@ def isclose(self, y, rtol=None, atol=None):
...
+        a = np.empty((), dtype=self.dtype)
+        b = np.empty((), dtype=da.asanyarray(y).dtype)
+        try:
+            # Check if a numerical isclose is possible
+            np.isclose(a, b)
+        except TypeError:
+            # self and y do not have suitable numeric data types
+            # (e.g. both are strings)
+            return self == y
+        else:
```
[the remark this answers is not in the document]
sadielbartholomew: I prefer the phrase 'was inspired by' over 'pinched' or similar!

cf/data/data.py
sadielbartholomew: All good, might as well include minor things in other PRs to avoid a long chain of PRs :)

PR #413, opened 2022-06-22
dask: `Data.allclose`

davidhassell:


cf/data/data.py
```diff
@@ -5337,55 +5337,51 @@ def all(self, axis=None, keepdims=True, split_every=None):
...
-            rtol: `float`, optional
-                The relative tolerance for all numerical comparisons. By
-                default the value returned by the `rtol` function is used.
+            {{atol: number, optional}}
 
         :Returns:
 
-            `bool`
+            `Data`
+                A scalar boolean array that is `True if the two arrays
```
sadielbartholomew, line 5363:
```suggestion
                A scalar boolean array that is `True` if the two arrays
```

sadielbartholomew (approved):
Pretty much all good to merge. Thanks.

PR #416, opened 2022-06-24
dask: `Field.isscalar`

davidhassell:


davidhassell:
Merging without review, as it is a simple one-liner, and is a blocker for further development.

PR #418, opened 2022-06-24
Fix for setting bounds with different but equivalent units

### LINKED ISSUE #417: Creating time bounds - differing time units causes uncaptured error when collapsing
Hi Both!

  Another minor issue - it arose when working with some datasets that use an auxiliary time axis, I have a work-around but I thought it was worth highlighting as it causes errors in `collapse` that are not captured.

If I create a field with 2 time points, but no bounds, then I create month-wide bounds using cf.M() and then add these to the field.
I get a `cf.field` that has a time coordinate that looks like this:

```
Dimension coordinate: time
    calendar = '360_day'
    long_name = 'Time axis'
    standard_name = 'time'
    time_origin = '1900-01-01 00:00:00'
    units = 'seconds since 1900-01-01 00:00:00'
    Data(time(2)) = [1851-06-01 00:00:00, 1852-06-01 00:00:00] 360_day
    Bounds:calendar = '360_day'
    Bounds:units = 'days since 1851-6-1'
    Bounds:Data(time(2), 2) = [[1851-06-01 00:00:00, ..., 1852-07-01 00:00:00]] 360_day

```

As you can see, the time dimension has units of seconds, but the bounds has units of days.

If I do `field.collapse('time: mean')` I get:

```
Dimension coordinate: time
    calendar = '360_day'
    long_name = 'Time axis'
    standard_name = 'time'
    time_origin = '1900-01-01 00:00:00'
    units = 'seconds since 1900-01-01 00:00:00'
    Data(time(1)) = [1900-01-01 00:03:15] 360_day
    Bounds:calendar = '360_day'
    Bounds:units = 'seconds since 1900-01-01 00:00:00'
    Bounds:Data(time(1), 2) = [[1900-01-01 00:00:00, 1900-01-01 00:06:30]] 360_day

```

Clearly something has gone wrong here!

It seems that the bounds data (in days) has been used, but the time units (seconds) have been retained:

```
time_mean.coord('time').array
array([195.])
```

The attached code reproduces this behaviour.
[time_bounds_bug2.py.gz](https://github.com/NCAS-CMS/cf-python/files/8975955/time_bounds_bug2.py.gz)

This is probably not a major bug, most people are not going to create bounds like this! But I thought I'd draw it to you attention, just in case it might cause other problems.

Cheers,
Dan

davidhassell:
Fixes #417

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,13 @@
+version 3.13.1
+--------------
+
+**2022-06-24**
+
+* Fix bug when setting coordinate bounds with different units to their
+  parent coordinates
+  (https://github.com/NCAS-CMS/cf-python/issues/417)
+
+----
 version 3.13.0
 --------------
 
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,7 +1,7 @@
 version 3.13.1
 --------------
 
-**2022-06-24**
+**2022-??-??**
 
 * Fix bug when setting coordinate bounds with different units to their
   parent coordinates
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -74,8 +74,8 @@ installation and source code.
 
 """
 __Conventions__ = "CF-1.9"
-__date__ = "2022-06-23"
-__version__ = "3.13.0"
+__date__ = "2022-06-24"
+__version__ = "3.13.1"
 
 _requires = (
     "numpy",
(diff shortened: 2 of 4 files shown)


sadielbartholomew:
Sorry for the late review on this David, it slipped off my radar somehow! Reviewing imminently...

sadielbartholomew (approved):
All good, capturing the underlying bug, which after looking through the issue and code here I understand to be that the units assigned (with any necessary data conversion) to the coordinate bounds *when set* (by `set_bounds`) should have been those consistent with the coordinate data itself, and fixing it in the sensible (and simplest) way.

Please merge.

PR #423, opened 2022-07-13
dask: Basic daskified reading and writing

davidhassell:
* This PR requires https://github.com/NCAS-CMS/cfdm/pull/196 to be merged into the (master of) `cfdm` being used.
* Only tests `setup_create_field.py` and `test_read_write.py` have been tested. The latter has some skips relating to `cf.aggregate`, which will be fixed elsewhere.
* Running these test on my local machine "hasn't seg faulted for a while". _edit - Oh, it just did 2 out of the last 3 times I ran test_read_write.py_

### DIFF
--- a/cf/read_write/netcdf/netcdfread.py
+++ b/cf/read_write/netcdf/netcdfread.py
@@ -1,12 +1,12 @@
-import json
-from ast import literal_eval as ast_literal_eval
-
 import cfdm
-from numpy import dtype as numpy_dtype
+import netCDF4
+import numpy as np
 
-from ...constants import _file_to_fh
-from ...functions import dirname, pathjoin
-from ...units import Units
+_cfa_message = (
+    "Reading CFA files has been temporarily disabled, "
+    "and will return at version 4.0.0. "
+    "CFA-0.4 functionality is still available at version 3.13.x."
+)
 
 
 class NetCDFRead(cfdm.read_write.netcdf.NetCDFRead):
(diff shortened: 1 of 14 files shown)


davidhassell:
I'm fairly sure that the seg faults are something to do with file open/close commands on chunks

cf/read_write/netcdf/netcdfread.py
```diff
@@ -412,26 +326,76 @@ def _customize_read_vars(self):
...
+            # --------------------------------------------------------
+            # netCDF4 does not auto-mask VLEN variable, so do it here.
+            # --------------------------------------------------------
+            array = np.ma.where(array == b"", np.ma.masked, array)
+
+        elif not string_type and kind == "O":
+            # --------------------------------------------------------
+            # A netCDF string type N-d (N>=1) variable comes out as a
+            # numpy object array, so convert it to numpy string array.
+            # --------------------------------------------------------
```
sadielbartholomew, line 398: Not that it matters or we win any awards for it (though the good commenting is noted, as it often is), but there's a satisfying alignment across all of the comments here :smiley:

sadielbartholomew (approved):
Looks good, and the relevant test modules pass for me locally. The full test suite did crash my laptop, but that is unlikely to be related and we can worry about that later!

Please merge.

cf/read_write/netcdf/netcdfread.py
sadielbartholomew: On second look, partly because one comment is the same as another, but still...

PR #424, opened 2022-07-18
Fix test_decorators issue by setting correct assertLogs levels

### LINKED ISSUE #419: Upgrade to Python 3.10
cf-python won't import under Python 3.10, because `cfdm` won't (see https://github.com/NCAS-CMS/cfdm/issues/187).

When cfdm is fixed for Python 3.10 (July 2022), cf-python can be assessed for any other issues.

**Update**: `test_decorators.py` needs changing along the lines of https://github.com/NCAS-CMS/cfdm/pull/197

davidhassell:
Fixes #419

See https://github.com/NCAS-CMS/cfdm/pull/197 for details.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -4,7 +4,7 @@ version 3.13.1
 **2022-07-??**
 
 * Upgrade to allow cf to work with Python 3.10
-  (https://github.com/NCAS-CMS/cfdm/issues/419)
+  (https://github.com/NCAS-CMS/cf-python/issues/419)
 * Fix bug when setting coordinate bounds with different units to their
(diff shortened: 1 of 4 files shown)


sadielbartholomew (approved):
All good save for a tweak to fix the changelog link. Notably, the full test suite passes locally with both my Python 3.10 and 3.9 conda environments.

You got at the `test_decorators` fix for cf-python before I could (popped it on my TODO list after completing the cfdm equivalent, to tackle after finishing up more urgent tasks), so thanks sinec I can tick that off now :)

Please merge when ready.

PR #426, opened 2022-08-02
LAMA to Dask: (deprecate) `Data.concatenate_data`

sadielbartholomew:
Migrating the `Data.concatenate_data` method towards #182, by deprecating it.

I don't think this approach will be controversial (so-as to have required pre-PR discussion) since it seems to be essentially the same method as `concatenate`, just compare the inputs and outputs on the docstrings to see this, with no real difference except for the behaviour of:

> In the case that the list contains only one element, that element is simply returned.

and it is not used anywhere throughout the codebase, unlike `concatenate` which is used quite widely (note this was a post-PR grep, but regardless the only usage is to define it):

```console
$ pwd
/home/sadie/cf-python/cf
$ git grep "concatenate_data"
data/mixin/deprecations.py:    def concatenate_data(cls, data_list, axis):
data/mixin/deprecations.py:            "Data method 'concatenate_data' has been deprecated at "
```

Anyhow @davidhassell you can object at this stage and close if there is a justification to keep the method, else review this towards merging, as the change itself is almost trivial. Thanks.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -5640,37 +5640,6 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
 
         return d
 
-    @classmethod
-    def concatenate_data(cls, data_list, axis):
-        """Concatenates a list of Data objects into a single Data object
-        along the specified access (see cf.Data.concatenate for
-        details). In the case that the list contains only one element,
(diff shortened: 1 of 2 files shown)


davidhassell (approved):
Great.

PR #427, opened 2022-08-02
LAMA to Dask: fixes to `Query` & `Data.digitize`

sadielbartholomew:
Reinstate a commented-out test case, now ready since pre-requisites have been satisfied, that calls `where` and tests `digitize` and in doing so reveals a bug in the migrated state of both (:flushed:), which this PR also fixes. See below for the breakdown.

(The fixes applied may not be the most appropriate course of action, or need tweaking, so I've applied them in self-contained commits which can be reverted if necessary. The most important thing here my outline of the issues arising below, which should be addressed by some means.)

### Details

Overall this PR:

1. Reinstates the test case in question, as-was, in f86923f.

2. Fixes the first issue encountered when reinstating, namely that `NoneType` errors were being hit:

    ```pytb
    ERROR: test_Data_digitize (__main__.DataTest)
      File "test_Data.py", line 829, in test_Data_digitize
        e.where(
      File "/home/sadie/cf-python/cf/decorators.py", line 62, in precede_with_kwarg_deprecation_check
      File "/home/sadie/cfdm/cfdm/decorators.py", line 44, in inplace_wrapper
      File "/home/sadie/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
      File "/home/sadie/cf-python/cf/data/data.py", line 93, in wrapper
      File "/home/sadie/cf-python/cf/data/data.py", line 9941, in where
        condition = condition.evaluate(d)
    AttributeError: 'NoneType' object has no attribute 'evaluate'

    ```

    ultimately due to rogue assignment to `set_condition_units` which returns nothing, such that it seems highly likely that calling without assigning was the intended action, and a fix has been applied as such in e450a40.

3. Fixes the next issue revealed in the stack trace, this one occurring because `Query.set_condition_units` was introducing an extra dimension for unitless values which was manifesting as:

    ```pytb
    ERROR: test_Data_digitize (__main__.DataTest)
      File "test_Data.py", line 829, in test_Data_digitize
        e.where(
      File "/home/sadie/cf-python/cf/decorators.py", line 62, in precede_with_kwarg_deprecation_check
      File "/home/sadie/cfdm/cfdm/decorators.py", line 44, in inplace_wrapper
      File "/home/sadie/cfdm/cfdm/decorators.py", line 171, in verbose_override_wrapper
      File "/home/sadie/cf-python/cf/data/data.py", line 93, in wrapper
      File "/home/sadie/cf-python/cf/data/data.py", line 9944, in where
        _where_broadcastable(d, condition, "condition")
      File "/home/sadie/cf-python/cf/data/data.py", line 12033, in _where_broadcastable
        raise ValueError(
    ValueError: where: Broadcasting the 'condition' parameter with 4 dimensions would change the shape of the data with 3 dimensions

    ```

    with a sensible fix applied in 1aec661 (*though in this case something else may be required in the unitless case, so not sure if straight-up removing the line is suitable as the general fix, this is to be discussed...*).

4. Fixes the final issue revealed by the reinstated test, in 5bb5766, specifically that the core assertion checking equality between the intended and actual results was failing:

    ```pytb
    FAIL: test_Data_digitize (__main__.DataTest)
      File "test_Data.py", line 836, in test_Data_digitize
        self.assertTrue(e.equals(f, verbose=2))

    ```

    because (after much investigation into what was actually going wrong once that key assertion was finally reached!) the `Data.digitize` method itself created a list called `delete_bins` that was not used later in the code to do anything that had an effect on the outcome, when clearly it was designed to do something...

    I cross-referenced the migration PR (#312) and old `master` code after which it seemed that the following block had not been incorporated:

    https://github.com/NCAS-CMS/cf-python/blob/07cb04f5ea0a4e7d05c90db8366f8fab731e844b/cf/data/data.py#L2647-L2652

    so I introduced it and adapted it to the new way, and this change did indeed (finally!) get the test case to pass, without causing side effects elsewhere.

So I think (4) concludes the multi-fix, but let me know what you think, @davidhassell! Thanks.


sadielbartholomew:
@davidhassell I have made updates in line with our discussion from last week in the two latest commits. As far as I am concerned this is ready to go, so please review when you are back from leave and have time. Thanks.

cf/query.py
```diff
@@ -788,10 +788,7 @@ def set_condition_units(self, units):
             return
 
         value_units = getattr(value, "Units", None)
-        if value_units is None:
-            # Value has no units
-            value = Data(value, units=units)
-        else:
+        if value_units is not None:
```
davidhassell, line 791: Not sure about this - it doesn't work if `value` does not have a `Units` attribute: 
```python
>>> q = cf.lt(6)
>>> q._value
6
>>> q.set_condition_units('m')
>>> q._value
6  # No units here!
```
The orginal code does:
```python
>>> q._value
6
>>> q.set_condition_units('m')
>>> q._value
<CF Data(): 6 m>  # Seems right to me ...
```

How was the original code causing problems?

davidhassell (commented):
Hi Sadie - the `digitize` and `where` changes look good, but I have a question on the `Query` changes ...

cf/query.py
sadielbartholomew: Thanks David, I'm updating this in line with our discussion in the video call today which supersedes your comment here but obviously concerns the same line change and underlying issue.

sadielbartholomew:
Back onto this, with apologies for letting it fall off my radar! @davidhassell, to continue the conversation where we left off after the first round of feedback, namely regarding the outstanding piece of feedback https://github.com/NCAS-CMS/cf-python/pull/427#discussion_r946631468...

The last progress made towards this PR was back in August after your initial review where we both had a brief external pair-programming type session to investigate the `query` aspect of this PR together. Since I made written notes rather than electronic ones I've had to dig those out, and now I've found them, let me summarise what we found (hopefully it maps roughly onto any memory you might have of this conversation!):

* Overall, there *is* an underlying issue with the `query` module, ultimately manifesting in the error I stated in my opening comment (case 3), but I hadn't fixed it with https://github.com/NCAS-CMS/cf-python/commit/1aec6619861be948ec7bd9860e9409d2cd8e9642 in a correct way that accounted for all possibilities sensibly, as you pointed out for a different case.
* With some debugging we realised that the issue was due fundamentally to sequences of `value`s being put into a container array, introducing an extra dimension, which caused the test failure, when those should have been handled properly. To demonstrate this via the failing test, as demonstrated here *on the original code i.e. before any suggestions by this PR*, if I add in some `print` statements, as follows:

```diff
diff --git a/cf/query.py b/cf/query.py
index 67932f9f8..9a38cbffe 100644
--- a/cf/query.py
+++ b/cf/query.py
@@ -788,14 +788,19 @@ class Query:
             return
 
         value_units = getattr(value, "Units", None)
+        print("\n%%%%% 0 VALUE UNITS IS", value, value_units)
         if value_units is None:
             # Value has no units
+            print("\n%%%%% 1.0 IS\n", value, value_units)
             value = Data(value, units=units)
+            print("\n%%%%% 1.1 IS\n", value, value_units, value.shape)
         else:
             # Value already has units
             try:
+                print("\n%%%%% 2 IS", value, value_units, value.shape)
                 value.Units = units
             except ValueError:
+                print("\n%%%%% 3 IS", value, value_units, value.shape)
                 raise ValueError(
                     f"Units {units!r} are not equivalent to "
                     f"query condition units {value_units!r}"
```

then

```python
test_Data_digitize (__main__.DataTest) ... 
%%%%% 0 VALUE UNITS IS [<CF Data(1, 1, 1): [[[0]]]>, <CF Data(1, 1, 1): [[[5]]]>] None

%%%%% 1.0 IS
 [<CF Data(1, 1, 1): [[[0]]]>, <CF Data(1, 1, 1): [[[5]]]>] None

%%%%% 1.1 IS
 [[[[0, 5]]]] None (2, 1, 1, 1)
ERROR

```

* In terms of how to handle the sequences of `Data` case causing issue as above, I didn't write down the approach we tentatively agreed to (my bad for not finalising this PR shortly after we discussed it!) but the natural approach and the one that seems to match my vague memory of the conversation for what we agreed would be sensible as a solution is to check if the `value` is a sequence of `Data` and if so `concatenate` the separate `Data` together and then apply the specified units. This has been implemented in my new commit [793e7ce](https://github.com/NCAS-CMS/cf-python/pull/427/commits/793e7ce74280d157ed13f9a65900c1d7830de8d8) which I applied after reverting the commit making the initial changes to the `query` module.

Note it passes, but only if the current `concatenate` method code is copied from the `lama-to-dask` branch since when I started this PR that method had not yet been migrated to Dask, so otherwise the test fails on trivial migration-related issues with the call to `concatenate`.

cf/query.py
```diff
@@ -790,7 +791,11 @@ def set_condition_units(self, units):
         value_units = getattr(value, "Units", None)
         if value_units is None:
             # Value has no units
-            value = Data(value, units=units)
+            if isinstance(value, Iterable):  # may be a sequence of Data
```
davidhassell, line 794: Hi Sadie, not sure this will work in all cases:

* `Data` objects are `Iterable`
* When `value` is a sequence, we want it to remain a sequence, not convert it into a `Data` object, and `Data.concatenate` doesn't work for non-Data objects 

This code might make sense, but it's not very pretty!
```python
        value_units = getattr(value, "Units", None)
        if value_units is None:
            # Value has no units
            if isinstance(value, Iterable) and not isinstance(value, str):
                new = []
                for v in value:
                    value_units = getattr(v, "Units", None)
                    if value_units is None:
                        # Value has no units
                        v = Data(v, units=units)
                    else:
                        # Value already has units
                        try:
                            v.Units = units
                        except ValueError:
                            raise ValueError(
                                f"Units {units!r} are not equivalent to "
                                f"query condition units {value_units!r}"
                            )
                        
                    new.append(v)

                value = new
            else:
                value = Data(value, units=units)        
        else:
            # Value already has units
            try:
                value.Units = units
            except ValueError:
                raise ValueError(
                    f"Units {units!r} are not equivalent to "
                    f"query condition units {value_units!r}"
                )
```

davidhassell:
Hi Sadie - I think we have a bit of an issue here, as some of this code has already been modified in #464 - could you merge `lama-to-dask` into your branch and have another look :) Thanks

sadielbartholomew:
> could you merge lama-to-dask into your branch and have another look 

Hi David, no worries. I've done as you ask but it seems to have merged cleanly, and upon investigation it looks like the one commit where I touched `where` code, namely e450a407725914def8322d2753723318cf018a01, to fix one issue, had since I opened this PR been merged in to `lama-to-dask` elsewhere, and in fact I see you did make the exact same change in #464. So we're all good with respect to this - there's no issue!

Overall, with my new merge commit, the `test_Data` tests all still pass (apart from the one that is still skipped), so everything is good to go still as far as I am concerned, but I await your thoughts regarding:

> Hi Sadie, I think what I wrote here may well be nonsense! Please ignore, and the code snippet and I'll have another think ...

in case changes are necessary with regards to that. Thanks!

davidhassell:
Thanks, Sadie. I'm now going to stop dipping in to this piecemeal and do a proper review later today/tomorrow, Sorry for any extra confusion I have wrought!

cf/data/data.py
```diff
@@ -1787,6 +1787,18 @@ def digitize(
...
 
+        # More elegant to handle 'delete_bins' in cf- rather than Dask- space
+        # i.e. using cf.where with d in-place rather than da.where with dx
+        # just after the digitize operation above (cf.where already applies
+        # equivalent logic element-wise).
+        if delete_bins:
+            for n, db in enumerate(delete_bins):
+                db -= n
+                d.where(d == db, np.ma.masked, None, inplace=True)
+                # x = d - 1 rather than = d here since there is one less bin
```
davidhassell, line 1798:
```suggestion
                # x = d - 1 rather than = d here since there is one fewer bin
```

cf/query.py
davidhassell: OK - I think I worked out what's going on. The key is that the contents of a "set", "wi" or "wo" iterable are not necessarily `Data` objects (or rather objects with `Units`. How about this? There's scope for a bit of code re-use here, but I can't quite see how, yet.

```python

        value_units = getattr(value, "Units", None)
        if value_units is None:
            # Value has no units
            if self.operator in ("wi", "wo", "set"):
                # value is a sequence of things that may or may not
                # already have units
                new = []
                for v in value:
                    v_units = getattr(v, "Units", None)
                    if v_units is None:
                        v = Data(v, units=units)
                    else:
                        try:
                            v = v.copy()
                            v.Units = units
                        except ValueError:
                            raise ValueError(
                                f"Units {units!r} are not equivalent to "
                                f"query condition units {v_units!r}"
                            )
                        
                    new.append(v)

                value = new
            else:
                value = Data(value, units=units)
        else:
            # Value already has units
            try:
                value = value.copy()
                value.Units = units
            except ValueError:
                raise ValueError(
                    f"Units {units!r} are not equivalent to "
                    f"query condition units {value_units!r}"
                )

        self._value = value
```
We could do with some more units test for this, too ...

davidhassell (changes requested):
Hi Sadie, I've made a suggestion for the changes to `Query` (and a request for some more tests), otherwise all good. A small amount of code caused a lot of thought, here!

cf/data/data.py
sadielbartholomew: Good spot!

sadielbartholomew:
> The key is that the contents of a "set", "wi" or "wo" iterable are not necessarily Data objects (or rather objects with Units.

Aha, that makes so much sense now you have worked it out and summarised it! And the change block you suggest follows on sensibly from that. I'll get the equivalent code updated in line with your suggestion. Thanks!

> We could do with some more units test for this, too ...

Agreed. Right-o, I'll increase the coverage in this respect as part of this PR. I'll update it shortly and tag you to let you know we're ready for a re-review.

cf/query.py
sadielbartholomew: I've just added your code and then consolidated it with an inner helper function. I'll add some new testing first thing tomorrow.

sadielbartholomew:
Sorry @davidhassell I think I forgot to confirm that this is ready for re-review now after I updated this in line with your suggestion and made some consolidations too. Please let me know what you think.

cf/query.py
```diff
@@ -857,32 +857,47 @@ def set_condition_units(self, units):
...
             for r in compound:
                 r.set_condition_units(units)
-
             return
 
         value = self._value
         if value is None:
             return
 
         value_units = getattr(value, "Units", None)
```
davidhassell, line 889: This line is no longer needed.

cf/query.py
```diff
@@ -857,32 +857,47 @@ def set_condition_units(self, units):
...
-        if value_units is None:
-            # Value has no units
-            value = Data(value, units=units)
+        if self.operator in ("wi", "wo", "set"):
+            # Value is a sequence of things that may or may not
+            # already have units
```
davidhassell, line 892: There's a potential performance issue here, in that if a large set is stored as a `Data` object, as opposed to a python built-in like a list, then iterating through it could be slow. For the `Data` object case we could just change the units without iterating through.

My inclination is to resist the overwhelming urge for premature optimsation and do nothing for now, as I confident that folk don't/won't be using huge sets, but I'll leave that decision to you :)

davidhassell (commented):
Hi Sadie - A nice solution, thanks. I've made a couple of comments, but go ahead and merge when you're ready.

cf/query.py
davidhassell, line 892: There's a potential performance issue here, in that if a large set is stored as a Data object, as opposed to a python built-in like a list, then iterating through it could be slow. For the Data object case we could just change the units without iterating through.

My inclination is to resist the overwhelming urge for premature optimsation and do nothing for now, as I confident that folk don't/won't be using huge sets, but I'll leave that decision to you :)

cf/query.py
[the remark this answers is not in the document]
sadielbartholomew: Good spot. It looks like there are a small number of other linting issues present too, after I checked with `$ pre-commit run --all-files` so I'll get those sorted in a batch.

cf/query.py
sadielbartholomew: > There's a potential performance issue here...

Well raised, this could be important in certain cases.

> My inclination is to resist the overwhelming urge for premature optimsation and do nothing for now

It might be wise to manage this somehow, e.g. as your suggest, but I agree that we should worry about it a bit later down the line of the migration, when we start doing profiling and such. So let's be 'lazy in a good way' like Dask and indeed do nothing, and I'll create an Issue quoting your concern we can tackle at the appropriate time!

davidhassell: Sounds like a plan.

sadielbartholomew:
Thanks for all of your very useful feedback on this @davidhassell. Final round of feedback addressed in my latest commit, so I'll merge as advised.

cf/query.py
sadielbartholomew: Issue created (linked into the PR below), so marking this as resolved.

PR #431, opened 2022-08-05
LAMA to Dask migration: branch tidy

sadielbartholomew:
Tidies the `lama-to-dask` branch in the following ways:

* addresses https://github.com/NCAS-CMS/cf-python/pull/425#discussion_r930734685;
* addresses https://github.com/NCAS-CMS/cf-python/pull/425#discussion_r931689805;
* makes formatting tweaks from running `pre-commit run --all-files` for a pass, to apply all required (not much, in this case, but some) auto-formatting across the codebase;
* adds appropriate docstrings to all of the `test_Data` methods (`docformatter` now checks these); 
* un-skips three `test_Data` tests which pass due to other migration progress, but hadn't yet been un-skipped (three skips remain across the module, which we will consider separately to this PR). 

See the separate commits which each focus on one of the five above (should be obvious which).

@davidhassell nothing here is controversial, and indeed the first two points were your own suggestions from a previous code review, and I am aware you are about to go on leave also, so I think it is most suitable to will merge without asking for a review.

### DIFF
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -96,8 +96,6 @@ class DataTest(unittest.TestCase):
     mw = mw
     ones = ones
 
-    test_only = []
-
     def setUp(self):
         # Suppress the warning output for some specific warnings which are
         # expected due to the nature of the tests being performed.
@@ -114,9 +112,6 @@ class DataTest(unittest.TestCase):
             )
 
     def test_Data_equals(self):
-        if self.test_only and inspect.stack()[0][3] not in self.test_only:
-            return
-
         shape = 3, 4
         chunksize = 2, 6
         a = np.arange(12).reshape(*shape)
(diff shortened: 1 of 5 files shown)


cf/test/test_Data.py
```diff
@@ -913,9 +901,7 @@ def test_Data_flatten(self):
 
     @unittest.skipIf(TEST_DASKIFIED_ONLY, "no attribute 'partitions'")
```
sadielbartholomew, line 902: @davidhassell I suspect this test can be removed, as it is LAMA-centered, but I am not 100% sure?

davidhassell: Yes, this can go.

davidhassell (commented):
All good, thanks!

cf/test/test_Data.py
sadielbartholomew: Thanks for clarifying, I will remove it in a stand-alone commit.

PR #432, opened 2022-08-10
LAMA to Dask migration: `Data.stats`

sadielbartholomew:
Migrates the `Data.stats` method towards https://github.com/NCAS-CMS/cf-python/issues/182.

Since `stats` is a compound method, in that it is in essence just reporting the outputs from various other `cf` stats/collapsing methods, each of which have already been 'daskified' hence utilise dask under-the-hood, the simplest way to migrate with full performance (bar working out and making use of any sub-calculation operations which might be shared between any of the statistics or something like that which would surely be over the top for our purposes?) is (I believe) to run each statistic calculation in parallel, which in Dask world for this context is done with `delayed` functions and a final `compute`.

This PR implements this. As indicated by the resultant Dask task graph shown below, all statistics are calculated separately, not in serial, so `stats` should take only as long as the most intensive calculation rather than the sum of all calculation times.

(I hope I haven't misunderstood the assignment here, so to speak, by taking such an approach centered on Dask's `delayed`.)

#### Graph

With the code as-is on the branch/PR here and now, but with a little tweak so we can access and save the Dask task graph, namely I did:

```diff
diff --git a/cf/data/data.py b/cf/data/data.py
index e91d6df03..2da53c70f 100644
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -15,6 +15,7 @@ from dask.array.core import normalize_chunks
 from dask.base import is_dask_collection, tokenize
 from dask.core import flatten
 from dask.highlevelgraph import HighLevelGraph
+from dask import delayed, compute, visualize
 
 from ..cfdatetime import dt as cf_dt
 from ..constants import masked as cf_masked
@@ -9376,16 +9377,20 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
-        return compute(out)[0]
+        return out
```

the task graph generated interactively via:

```python
In [1]: import cf
   ...: import numpy as np
   ...: import dask

In [2]: a = cf.example_field(1)

In [3]: b = a.data.stats()

In [4]: dask.visualize(b, filename='stats-graph-final.png')
Out[4]: <IPython.core.display.Image object>
```

is:

stats-graph-final

which looks right to me. The task graph here doesn't indicate but obviously all results get pulled together into the `dict` output at compute-time at the end via querying the `compute` output tuple).

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -9317,55 +9317,55 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         [[0  1  2]
          [3 --  5]]
         >>> d.stats()
-        {'minimum': <CF Data(): 0>,
-         'mean': <CF Data(): 2.2>,
-         'median': <CF Data(): 2.0>,
-         'maximum': <CF Data(): 5>,
-         'range': <CF Data(): 5>,
-         'mid_range': <CF Data(): 2.5>,
-         'standard_deviation': <CF Data(): 1.7204650534085253>,
-         'root_mean_square': <CF Data(): 2.792848008753788>,
+        {'minimum': 0,
+         'mean': 2.2,
+         'median': 2.0,
+         'maximum': 5,
+         'range': 5,
+         'mid_range': 2.5,
+         'standard_deviation': 1.7204650534085255,
+         'root_mean_square': 2.792848008753788,
          'sample_size': 5}
         >>> d.stats(all=True)
+        {'minimum': 0,
+         'mean': 2.2,
+         'median': 2.0,
+         'maximum': 5,
+         'range': 5,
+         'mid_range': 2.5,
+         'standard_deviation': 1.7204650534085255,
+         'root_mean_square': 2.792848008753788,
+         'minimum_absolute_value': 0,
+         'maximum_absolute_value': 5,
+         'mean_absolute_value': 2.2,
+         'mean_of_upper_decile': 5.0,
+         'sum': 11,
+         'sum_of_squares': 39,
+         'variance': 2.9600000000000004,
+         'sample_size': 5}
+        >>> d.stats(mean_of_upper_decile=True, range=False)
+        {'minimum': 0,
+         'mean': 2.2,
+         'median': 2.0,
+         'maximum': 5,
+         'mid_range': 2.5,
+         'standard_deviation': 1.7204650534085255,
+         'root_mean_square': 2.792848008753788,
+         'mean_of_upper_decile': 5.0,
+         'sample_size': 5}
+
+        To ask for delayed operations instead of computed values:
+
+        >>> d.stats(compute=False)
         {'minimum': <CF Data(): 0>,
          'mean': <CF Data(): 2.2>,
          'median': <CF Data(): 2.0>,
          'maximum': <CF Data(): 5>,
          'range': <CF Data(): 5>,
          'mid_range': <CF Data(): 2.5>,
-         'standard_deviation': <CF Data(): 1.7204650534085253>,
+         'standard_deviation': <CF Data(): 1.7204650534085255>,
          'root_mean_square': <CF Data(): 2.792848008753788>,
-         'minimum_absolute_value': <CF Data(): 0>,
-         'maximum_absolute_value': <CF Data(): 5>,
-         'mean_absolute_value': <CF Data(): 2.2>,
-         'mean_of_upper_decile': <CF Data(): 5.0>,
-         'sum': <CF Data(): 11>,
-         'sum_of_squares': <CF Data(): 39>,
-         'variance': <CF Data(): 2.96>,
-         'sample_size': 5}
-        >>> d.stats(mean_of_upper_decile=True, range=False)
-        {'minimum': <CF Data(): 0>,
-         'mean': <CF Data(): 2.2>,
-         'median': <CF Data(): 2.0>,
-         'maximum': <CF Data(): 5>,
-         'mid_range': <CF Data(): 2.5>,
-         'standard_deviation': <CF Data(): 1.7204650534085253>,
-         'root_mean_square': <CF Data(): 2.792848008753788>,
-         'mean_of_upper_decile': <CF Data(): 5.0>,
-         'sample_size': 5}
-
-        Ask for delayed operations instead of computed values for the stats:
-
-        >>> d.stats(compute=False)
-        {'minimum': Delayed('minimum-0161ad64-aa89-43a7-b651-85a939822f7d'),
-         'mean': Delayed('mean-cef11472-c0b2-4f92-b19a-41295758a870'),
-         'median': Delayed('median-22330dad-f9ff-4337-ae66-03545a529021'),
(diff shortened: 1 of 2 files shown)


cf/data/data.py
```diff
@@ -9376,16 +9378,16 @@ def stats(
...
+                    value = delayed(func)(squeeze=True)
                 else:
-                    value = func(squeeze=True, weights=weights)
+                    value = delayed(func)(squeeze=True, weights=weights)
 
                 out[stat] = value
 
         if all or sample_size:
-            out["sample_size"] = int(self.sample_size())
+            out["sample_size"] = delayed(lambda: int(self.sample_size()))()
```
sadielbartholomew, line 9388: I know lambdas are frowned upon somewhat, but it seems intuitive to use one here instead of defining a trivial function, I think.

davidhassell:
Hi Sadie,

This is a really nice idea. Perhaps we could go one step further and potentially avoid having to read the data from disk each time. If we defined simple wrappers for the stats functions, they could share the same underlying data. E.g.

```python
import dask.array as da
from dask import delayed, compute

def max(x):
    return x.max()

def min(x):
    return x.min()

dx = da.arange(100, chunks=50)

stats = (delayed(min, pure=True)(dx), delayed(max, pure=True)(dx))
compute(*stats)
# (0, 99)
```

It would involve writing a little function for each stat (or lambdas - but I don't know how that would play with the `weights` kwarg), but they would be contained within the `stats` method ... What do you think?

(I'm sure you can visualise the above graph, but I couldn't work out how. The docs imply that the `dx` will be shared between functions, though, I think)

sadielbartholomew:
Thanks David, this:

> Perhaps we could go one step further and potentially avoid having to read the data from disk each time. If we defined simple wrappers for the stats functions, they could share the same underlying data.

is a great idea. Once I've tied up some loose ends elsewhere I will have a crack at getting it implemented in the way you suggest, which also sounds nice to me.

sadielbartholomew:
Hi @davidhassell, I got back onto this one this week (intermittently) as per our priority list. I've updated it to incorporate your idea in https://github.com/NCAS-CMS/cf-python/pull/432#issuecomment-1214809276. Let me know if I've implemented that as you intended, correctly, and if there is any further feedback.

sadielbartholomew:
Regarding the fact that the data type of the scalar values themselves is not consistent, at least for our testing case with a `stats` input of `cf.Data[1, 1]` and `weights=1` set, namely that the values are in some cases `float` and in some cases `int` (see the `test_Data_stats` dict compared output now, noting that this is the form when the test passes, with the `stats` method code as-is), I think that is fine because it makes sense in the context of the different operations calculations, .e.g a maximum of `1` and `1` is `1` not `1.0`, but the mean is `1.0`.

sadielbartholomew:
Updating now in line with an offline discussion with @davidhassell, in which we realised that the approach in https://github.com/NCAS-CMS/cf-python/pull/432#issuecomment-1214809276 wasn't worthwhile. In the meantime, David, what are your thoughts regarding the output value format and `ndim` as I outline in https://github.com/NCAS-CMS/cf-python/pull/432#issuecomment-1300535058, in case we want to tweak the output format?

davidhassell:
Hi Sadie - thanks for the nudge. This is a bit tricky, as the results haven't actually been computed, now! Perhaps we could have a `compute` keyword parameter, that defaults to `True`, which determines whether or not computed `<scalar>` values are returned, or else non-computed `Data` objects?

sadielbartholomew:
> Perhaps we could have a compute keyword parameter, that defaults to True, which determines whether or not computed <scalar> values are returned, or else non-computed Data objects?

Aha, I like that idea! Thanks, let's do that. I'll add it in today. I was also updating the test to cover a representative range of possible input parameters since it was too basic as it was. I'll push the commits up shortly.

sadielbartholomew:
All feedback incorporated so this is now ready for re-review @davidhassell. Thanks.

cf/data/data.py
```diff
@@ -9342,6 +9354,18 @@ def stats(
          'mean_of_upper_decile': <CF Data(): 5.0>,
          'sample_size': 5}
 
+        # Ask for delayed operations instead of computed values for the stats:
```
davidhassell, line 9357:
```suggestion
        Ask for delayed operations instead of computed values for the stats:
        
```

cf/data/data.py
```diff
@@ -9376,16 +9400,26 @@ def stats(
...
+                    value = delayed(func)(squeeze=True, weights=weights)
 
                 out[stat] = value
 
         if all or sample_size:
-            out["sample_size"] = int(self.sample_size())
+            out["sample_size"] = delayed(lambda: self.sample_size())()
 
-        return out
+        if compute:
```
davidhassell, line 9412: Hi Sadie, I think we should return the actual `Data` objects when `compute=False`, rather than `dask.delayed` objects. The former are still lazy, but much more user-friendly.

davidhassell (changes requested):

cf/data/data.py
sadielbartholomew: Oh, I see! I misunderstood what you were implying, sorry. Fair enough, that makes sense. I'll update this shortly to the above.

sadielbartholomew:
@davidhassell have my last two commits amended this to what you were thinking, namely:

> I think we should return the actual Data objects when compute=False, rather than dask.delayed objects.

I assume the `Data` from the `compute=False` outputs are showing data values when printed because `can_compute` is set to return `True` for development purposes at the moment? Or because a representation operation forces a compute, at least right now? Unless I have missed something?

davidhassell:
This look perfect to me. I tried:
```python
In [1]: import cf

In [2]: f = cf.example_field(0)

In [3]: f.data.stats()
Out[3]: 
{'minimum': 0.003,
 'mean': 0.046075,
 'median': 0.036,
 'maximum': 0.146,
 'range': 0.143,
 'mid_range': 0.0745,
 'standard_deviation': 0.03613612285511549,
 'root_mean_square': 0.05855531572795078,
 'sample_size': 40}

In [4]: f.data.stats(compute=False)
Out[4]: 
{'minimum': <CF Data(): 0.003 1>,
 'mean': <CF Data(): 0.046075 1>,
 'median': <CF Data(): 0.036 1>,
 'maximum': <CF Data(): 0.146 1>,
 'range': <CF Data(): 0.143 1>,
 'mid_range': <CF Data(): 0.0745 1>,
 'standard_deviation': <CF Data(): 0.03613612285511549 1>,
 'root_mean_square': <CF Data(): 0.05855531572795078 1>,
 'sample_size': <CF Data(1, 1): [[40]]>}

In [5]: 
```
The values are appearing in the `Data` `repr` for exactly the reason you say (can_compute=True), but if we did `stats = f.data.stats(compute=False)` then we'd still be lazy, of course.

davidhassell (approved):
Hi @sadielbartholomew, thanks for working through all of these subtleties. Please merge when you're ready.

sadielbartholomew:
> The values are appearing in the Data repr for exactly the reason you say (can_compute=True), but if we did stats = f.data.stats(compute=False) then we'd still be lazy, of course.

Brilliant, thanks for clarifying. Merging now.

Issue #435, opened 2022-08-12
Dubious type casting for augmented arithmetic operators


sadielbartholomew:
We currently allow, and validate via testing in `test_Data_BINARY_AND_UNARY_OPERATORS`, some behaviour relating to input and output data types for [augmented arithmetic assignment operators](https://docs.python.org/3/reference/datamodel.html?highlight=iadd#object.__iadd__) that is not allowed by NumPy, and we should consider whether this is suitable or not. I am inclined to say we should make appropriate changes to adopt the NumPy behaviour.

### Specifics

Namely, when an augmented assignment is performed using inputs with data types which lead to a change in data type for the output, e.g. for a simplified scalar case something like `a = 1; a += 1.0`, we permit an in-place change of array `dtype`. As a minimal example, note how NumPy raises a type casting error for the equivalent operation below, whereas we go ahead and produce an output with a changed data type, the same type that the operation not in-place would produce:

```python
>>> import cf
>>> import numpy as np
>>> 
>>> # Setup equivalent arrays
>>> i_np = np.array([1, 2, 3])
>>> i_cf = cf.Data(i_np)
>>> 
>>> # NumPy raises a type casting error:
>>> i_np + 1.0  # operation not in-place is fine
array([2., 3., 4.])
>>> i_np += 1.0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
numpy.core._exceptions.UFuncTypeError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
>>> 
>>> # ... whereas cf performs the operation to give the same result data type
>>> # as the non in-place operation would:
>>> i_cf + 1.0
<CF Data(3): [2.0, 3.0, 4.0]>
>>> i_cf += 1.0
>>> i_cf
<CF Data(3): [2.0, 3.0, 4.0]>

```

and the equivalent behaviour occurs for the various `__i<operator>__` operators.

### Relevant cases in test suite

For reference, the tests in `test_Data_BINARY_AND_UNARY_OPERATORS` which were checking for this (dubious) behaviour, which remain as such from before the LAMA to Dask migration, are:

https://github.com/NCAS-CMS/cf-python/blob/003374315b37568f8594db0bbbab3b5d5c31e0b8/cf/test/test_Data.py#L2062-L2154

PR #436, opened 2022-08-12
LAMA to Dask: `_combined_units` & further branch tidy

sadielbartholomew:
Ultimately reinstate a commented-out test case, failing due to a minor issue in `_combined_units`, which turned out to simply be a use of the property `Data._size` which has straightforwardly been replaced by `Data.size`. I also updated all other remaining usage of `_size` in the Data module, which turned out to be two extra cases overall.

There were in fact two test cases due to be reinstated from the above fix, but one (that concerns `__ipow__`) is failing due to data type differences summarised by #435, so I have updated the messages provided by the commented-out case.


davidhassell (approved):
All good. Sorry it took me so long!

Issue #433, opened 2022-08-11
LAMA to Dask: address `TODODASK` items


sadielbartholomew:
Whilst migrating the Data object itself from LAMA to Dask (notably migrating the methods in #295), we left various `TODODASK` markers. This is a bumper issue to record that these should be addressed for us to have considered #182 closable (addressing may in some cases mean converting to a standard `TODO` if that feels appropriate).

At the time of commenting, we have ~250 cases to resolve:

```console
$ pwd
/home/sadie/cf-python/cf
$ git grep "TODODASK" | wc -l
253
```

These have been classified as one of the following via a `TODODASK<optional marker>` searchable form (copied from https://github.com/NCAS-CMS/cf-python/pull/437#issue-1339605356):

* `TODODASKVER` for cases of placeholders for the final version for the release with the Dask migration which can now be sorted easily in a targeted way via a single sed or similar command;
* `TODODASKDOCS` for required docstring or documentation updates;
* `TODODASKMSG` for use-facing messages e.g. error messages;
* `TODODASKAPI` for API changes or considerations (potentially to bump to the release after Dask as per recent versioning discussions);
* `TODODASKDEPR` for anything deprecation related;
* anything else presently left as-is with `TODODASK` (miscellaneous etc.), these probably being the most meatly/difficult TODOs of all.

(Note that cases of one particular item can be found via something like (without the final pipe to `wc -l`):

```console
$ git grep "TODODASK" | grep -v "TODODASKVER" | grep -v "TODODASKDOCS" | grep -v "TODODASKMSG" | wc -l
45
```
)

Please add any notes and discussion points regarding these to this thread.

davidhassell:
These items now dealt with, apart from the ones we want to leave there for later. Closing.

PR #437, opened 2022-08-15
LAMA to Dask: TODO tag cataloguing & further tidy

sadielbartholomew:
Tidies the `lama-to-dask` branch further in the following ways:

...

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -1768,7 +1768,7 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
 
         >>> a = np.ma.arange(12).reshape(4, 3)
         >>> d = cf.Data(a, 'K')
-        >>> d[1, 1] = np.ma.masked
+        >>> d[1, 1] = cf.masked
         >>> print(d.array)
         [[0 1 2])
          [3 -- 5]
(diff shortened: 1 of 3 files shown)


davidhassell (approved):
Thanks, Sadie. This'll make things easier, and good spot on the `Exception` -> `ValueError` item.

sadielbartholomew:
Thanks for reviewing David. I am about to push a few more 'TODO' tag identifier updates (sorry I should have mentioned that it might be best to review at the end, keeping this in draft form) but they are all quite trivial as with the rest of this PR, so if you don't think it necessary there's no need to review the few extra commits I'll push in a moment.

davidhassell:
Ah - sorry for being slightly precipitous. Good idea to merge whenever you're ready.

sadielbartholomew:
> Ah - sorry for being slightly precipitous.

My fault, I opened this in non-draft form by mistake so you were notified pre-maturely.

> Good idea to merge whenever you're ready.

Thanks, will do.

cf/data/data.py
```diff
@@ -2256,6 +2253,9 @@ def can_compute(self, functions=None, log_levels=None, override=False):
                 False.
 
         """
+        # TODODASKAPI - this method is premature - needs thinking about as part
```
sadielbartholomew, line 2256: These methods marked `TODODASKAPI` are API-related and therefore can probably be bumped to be considered instead as part of work to do after the Dask migration release, i.e. tranform to plain `TODO`s that we can keep track of via a dedicated Issue. I'll note that on the main Issue #433.

PR #440, opened 2022-08-24
dask: `cf.Field.varray`

davidhassell:


cf/field.py
```diff
@@ -4139,8 +4143,16 @@ def varray(self):
         <CF Data(5): [999, ... 4] kg m-1 s-2>
 
         """
-        self.uncompress(inplace=True)
-        return super().varray
+#        self.uncompress(inplace=True)
+#        return super().varray
+        _DEPRECATION_ERROR_ATTRIBUTE(
```
sadielbartholomew, line 4148: I get `NameError: name '_DEPRECATION_ERROR_ATTRIBUTE' is not defined` when I try `<a field>.varray()` and checking the module, it indeed has not been imported. (But it works fine, with the message being appropriate, after adding in the import, so that's the only amendment required here.)

sadielbartholomew (changes requested):
There's a missing import as reported in-line, but otherwise this is all good so please merge.

cf/field.py
davidhassell: Thanks - import added (https://github.com/NCAS-CMS/cf-python/pull/440/commits/5dcfa93addc96b33c6d5d0a9f746e568090e87e0)

PR #443, opened 2022-09-01
dask: Test for `Data.__getitem__` auxiliary masks

davidhassell:


cf/test/test_Data.py
```diff
@@ -1291,10 +1291,14 @@ def test_Data__getitem__(self):
         self.assertEqual(d[[0, 1], [0, 2]].shape, (2, 2))
         self.assertEqual(d[[0, 1], [2]].shape, (2, 1))
 
-        # Ancillary masks
-        #
-        # TODODASK: Test __getitem__ with ancillary masks. Can only do
-        #           this when cf.Data.where has been daskified
+        # Auxilliary masks
```
sadielbartholomew, line 1294:
```suggestion
        # Auxiliary masks
```

or keep the double-l for `ancillary` as per the TODO note? Mixed intention typo I suppose!

sadielbartholomew (approved):
One typo otherwise ready to go, so please merge.

cf/test/test_Data.py
davidhassell: On reflection (and having consulted a dictionary!) I think ancillary is probably a better fit - I'll replace a few occurrences in `data.py`, too.

PR #445, opened 2022-09-06
dask: `Data.Units` @Units.setter

davidhassell:
Allow `Data.Units` assignment when the data units have no been set.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -4169,7 +4169,7 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         except KeyError:
             pass
         else:
-            if not old_units.equivalent(value):
+            if old_units and not old_units.equivalent(value):
                 raise ValueError(
                     f"Can't set Units to {value!r} that are not "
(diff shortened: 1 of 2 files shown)


sadielbartholomew (approved):
All good, fixing and capturing via new testing a specific bug in the present `lama-to-dask` which I confirmed and is clear from the logic of the code pre-fix here, namely:

```pycon
>>> d = cf.Data(1.5)
>>> d.Units = cf.Units("s")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-06b287b500df> in <module>
----> 1 d.Units = cf.Units("s")

~/cf-python/cf/data/data.py in Units(self, value)
   4171         else:
   4172             if not old_units.equivalent(value):
-> 4173                 raise ValueError(
   4174                     f"Can't set Units to {value!r} that are not "
   4175                     f"equivalent to the current units {old_units!r}. "

ValueError: Can't set Units to <Units: s> that are not equivalent to the current units <Units: >. Consider using the override_units method instead.
```

which now works. Please merge.

PR #449, opened 2022-09-09
Original filenames methods

### LINKED ISSUE #448: New "original filenames" methods
The conversation at #365 discussed how to record the names of files that contained some or all of the data at instantiation time - typically (but not necessarily) from a read call.

I propose a new suite of `original_filenames` methods that allow us to define, report, and edit these "original" files - those that contained some or all of the data at instantiation time - and are independent of the `get_filenames` methods. `get_filenames` reports on any files that are used by Data objects _at the time of asking_, which might reported different (typically fewer) files than were being used _at the time of data instantiation_.

It is also the case that the daskification of the cf-python library confuses functionality of the `get_filenames` methods - due to the new lazy execution paradigm, the notion of files in use at the time of asking is a little fuzzy. For this reason, the `get_filenames` methods are likely to be deprecated. _Edit: These have now been deprecated in the PR_

The implementation of this will be in `cfdm` and entirely inherited into cf-python. See https://github.com/NCAS-CMS/cfdm/issues/215 for the implementation details, which were initially proposed as (pending review):

```python
>>> f = cfdm.read("parent_file.nc")[0]
>>> f.original_filenames()
{"parent_file.nc"}

>>> f = cfdm.read("parent_file.nc", external="external_file.nc")[0]
>>> f.original_filenames()
{"parent_file.nc", "external_file.nc"}
>>> f.original_filenames(update="file.nc")
>>> f.original_filenames()
{"parent_file.nc", "external_file.nc", "file.nc"}
>>> f.original_filenames(clear=True)
{"parent_file.nc", "external_file.nc", "file.nc"}
>>> f.original_filenames()
set()
```
You might want to edit the original file names if you are creating constructs ab initio (without `cf.read`, which will set them for you automatically), or if you are creating a Field (or Data or any other construct) that is some combination of others - in this case you might want to store the superset of original files.

Many thanks to @ThibHlln for first raising this the issue in #365.

davidhassell:
Fixes #448

Needs https://github.com/NCAS-CMS/cfdm/pull/216 to work.


davidhassell:
Needs https://github.com/NCAS-CMS/cfdm/pull/216 to work.

davidhassell:
OK - finished adding the bits I realised were missing!

davidhassell:
Note - some of the `.rst` files that have been modified in this PR had been, for some reason, reformatted into some sort of sphinx autodoc format, so these few (~6) have also been reinstated to the structured form. e.g. `cf.CellMeasure.rst`.

cf/data/mixin/deprecations.py
```diff
@@ -256,8 +256,8 @@ def expand_dims(self, position=0, i=False):
         )  # pragma: no cover
 
     def files(self):
-        """Deprecated at version 3.4.0, use method `get_filenames`
-        instead."""
+        """Deprecated at version 3.4.0, consider usin method
```
sadielbartholomew, line 259:
```suggestion
        """Deprecated at version 3.4.0, consider using method
```

sadielbartholomew (approved):
All good with regards to resolving #448 on top of the upstream implementation in place for `cfdm`, though I raised one minor comment in-line and **I do have some questions regarding the approach so please see https://github.com/NCAS-CMS/cf-python/issues/448#issuecomment-1263893466**.

(I do wonder if it might have been better to raise this against `lama-to-dask`, but since both branches are fairly syncronised at the moment it should be OK on `master` for checking it works as we've done now, and when we merge the former into the latter the merge conflicts should be OK, assuming you don't see a reason why they might not be that I don't!)

Please merge when ready. Thanks.

davidhassell:
Implementation now inherited from `cfdm` - see https://github.com/NCAS-CMS/cfdm/pull/216#issuecomment-1268566563 for details.

sadielbartholomew:
(Just resolved the trivial change log merge conflicts so I can re-review more quickly, which I am about to do now...)

cf/data/mixin/deprecations.py
```diff
@@ -313,9 +313,11 @@ def get_filenames(self):
         """Return the names of files containing parts of the data array.
 
         Deprecated at version TODODASKVER and and is no longer
-        available. Consider using the `original_filenames` method
+        available. Consider using the `get_original_filenames` method
         instead.
 
+        .. note:: Might get re-instated in a later version.
```
sadielbartholomew, line 319: :laughing:

sadielbartholomew (approved):
All good, re-review accounting for the change of approach covered in https://github.com/NCAS-CMS/cf-python/pull/449#issuecomment-1270711089 and a quick sanity check on the overall PR state to boot. Thanks, please merge!

PR #450, opened 2022-09-22
dask: `Field.indices`

davidhassell:


cf/query.py
```diff
@@ -266,6 +271,10 @@ def __ne__(self, x):
     def __and__(self, other):
         """The binary bitwise operation ``&``
 
+        Combine two queries with a logical And operation. If the
+        `!value` of both queries is the same then it will retained on
+        the compuind query.
```
sadielbartholomew, line 276:
```suggestion
        `!value` of both queries is the same then it will be retained on
        the compound query.
```

cf/query.py
```diff
@@ -292,6 +310,10 @@ def __iand__(self, other):
     def __or__(self, other):
         """The binary bitwise operation ``|``
 
+        Combine two queries with a logical Or operation. If the
+        `!value` of both queries is the same then it will retained on
+        the compuind query.
```
sadielbartholomew, line 315:
```suggestion
        `!value` of both queries is the same then it will be retained on
        the compound query.
```

sadielbartholomew (changes requested):
A few thoughts...

...

666 lines removed from the `field` module :thinking: !

sadielbartholomew (approved):
Apologies, clicked through with the wrong option on my review just now, meant to put 'Approve' naturally in light of the comments!

PR #457, opened 2022-10-07
dask: `Field.convolution_filter` and `Field.moving_window`

davidhassell:
I'm not entirely _why_ these weren't working under dask, but they look like they are now!


sadielbartholomew (approved):
All good, with both tests in the title now passing, as per the PR motivation. Good test cases added. I checked that "mirror" mode no longer was relevant (turns out that it has been deprecated).

Thanks, please merge.

PR #459, opened 2022-10-09
Inherit construct access from cfdm

davidhassell:
Inherit construct access from cfdm.

Since cfdm v1.10.0.0 much construct access was back-ported to cfdm. This PR essentially deletes code in cf that can now be inherited.

Also allows `test_Field_creation_commands` to pass.


cf/field.py
```diff
@@ -12257,23 +12212,47 @@ def convert(
...
+        Dimension coords: longitude(8) = [22.5, ..., 337.5] degrees_east
+        >>> print(x.array)
+        [ 22.5  67.5 112.5 157.5 202.5 247.5 292.5 337.5]
+        >>> cs = f.convert('X', cellsize=True)
+        >>> print(cs)
+        Field: longitude (ncvar%lon)
+        ----------------------------
+        Data            : longitude(longitude(8)) degrees_east
+        Dimension coords: longitude(8) = [22.5, ..., 337.5] degrees_east
+        >> print(cs.array)
```
sadielbartholomew, line 12238:
```suggestion
        >>> print(cs.array)
```

sadielbartholomew (approved):
Always lovely to see superfluous code cut! All good bar one typo. Please merge when addressed.

PR #464, opened 2022-10-12
dask: `Field.where`

davidhassell:
A bit of a tidy, and a change to `Data.where` to allow leading size one dimensions. Deleted the `_broadcast` function as it is not used.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -9565,7 +9565,7 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
 
         The array and the *condition*, *x* and *y* parameters must all
         be broadcastable across the original array, such that the size
-        of the result is identical to the orginal size of the
+        of the result is identical to the original size of the
         array. Leading size 1 dimensions of these parameters are
         ignored, thereby also ensuring that the shape of the result is
         identical to the orginal shape of the array.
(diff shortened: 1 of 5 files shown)


cf/data/data.py
```diff
@@ -9562,8 +9564,11 @@ def where(
         **Broadcasting**
 
         The array and the *condition*, *x* and *y* parameters must all
-        be broadcastable to each other, such that the shape of the
-        result is identical to the orginal shape of the array.
+        be broadcastable across the original array, such that the size
+        of the result is identical to the orginal size of the
```
sadielbartholomew, line 9568:
```suggestion
        of the result is identical to the original size of the
```

cf/field.py
```diff
@@ -14458,163 +14458,197 @@ def where(
...
+                  ``(3, 5)``, which does not broadcast to the shape of
+                  the ``f``.
 
             construct: `str`, optional
-                Define the condition by applying the *construct* parameter
-                to the given metadata construct's data, rather then the
-                data of the field construct. Must be
+                Define the condition by applying the *construct*
+                parameter to the given metadata construct's data,
+                rather then the data of the field construct. Must be
```
sadielbartholomew, line 14635: A previous typo but we might as well correct it now:
```suggestion
                rather than the data of the field construct. Must be
```

sadielbartholomew (approved):
All good, thanks. I've raised a few typos but other than that this is good to merge.

PR #466, opened 2022-10-13
dask: `Field.percentile`

davidhassell:
Mainly setting the `_axes` attribute correctly in `Data.percentile`.

### DIFF
--- a/cf/test/test_Field.py
+++ b/cf/test/test_Field.py
@@ -2238,7 +2238,7 @@ class FieldTest(unittest.TestCase):
 
         # Percentiles taken across *all axes*
         ranks = ([30, 60, 90], [20], 80)  # include valid singular form
-        
+
         for rank in ranks:
             # Note: Currently in cf the default is squeeze=False, but
             #       numpy has an inverse parameter called keepdims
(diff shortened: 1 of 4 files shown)


cf/test/test_Field.py
```diff
@@ -2235,23 +2235,23 @@ def test_Field_del_domain_axis(self):
...
-                # So first cases (n1, n1) are both squeezed, (n2, n2) are not:
-                a1 = numpy.percentile(f, rank)  # has keepdims=False default
-                b1 = f.percentile(rank, squeeze=True)
-                self.assertTrue(b1.allclose(a1, rtol=1e-05, atol=1e-08))
-                a2 = numpy.percentile(f, rank, keepdims=True)
-                b2 = f.percentile(rank)  # has squeeze=False default
-                self.assertTrue(b2.shape, a2.shape)
-                self.assertTrue(b2.allclose(a2, rtol=1e-05, atol=1e-08))
+
+        # Percentiles taken across *all axes*
```
sadielbartholomew, line 2239: This is making me think that it might be good to blitz all of the `for chunksize in self.chunk_sizes:` remaining in the test suite in one go:`git grep "for chunksize" | wc -l` says `6`, which isn't many now (most were in `test_Data` pre-migration, obviously) but enough to perhaps warrant a dedicated PR)?

sadielbartholomew (approved):
Excellent. Notably `test_Field_percentile` now passes and `test_Data_percentile` still passes. The new logic is sensible. I've raised one passing comment but this is all good and ready to merge, so please go ahead. Thanks.

cf/test/test_Field.py
davidhassell: Sounds good to me - would you like to do it?

sadielbartholomew: Sure, I'll add it to my list (medium-to-low priority unless you say otherwise).

PR #469, opened 2022-10-15
dask: `Data.array` return type (and test housekeeping)

davidhassell:
Sometime dask's `compute` returns a numpy int/float type rather than a scalar array, which doesn't work well with downstream methods (like `Data.__array__`), so `Data.array` make sure we trap those cases and cast it as a proper scalar array.

Also some housekeeping in `test_TimeDuration.py` (no substantive changes).


sadielbartholomew (approved):
Sensible and justified changes. All good, please merge.

PR #472, opened 2022-10-17
dask: General tidy/deprecation/removal of LAMA specific fuctions/classes/functionality

davidhassell:
Particularly related to file and memory management  - which is now delegated to `dask`.


davidhassell:
Hi Sadie - for some horrific reason, my merge of 3.13.1 ended up here! I think we'll get away with it if you just ignore everything in `docs` and its subdirectories. Thanks.

cf/read_write/read.py
```diff
@@ -466,16 +466,40 @@ def read(
                                           the upper bound of the top
                                           model level. By default the
                                           height at top model is taken
-                                          from the top level's upper
+                                          from the top level’s upper
```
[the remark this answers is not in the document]
sadielbartholomew: ... one of those six other cases being just below for `... can’t be determined ...`, hence my suggestion below for consistency with most of the rest of the codebase.

cf/read_write/read.py
```diff
@@ -466,16 +466,40 @@ def read(
...
                                           bound defined by BRSVD1 in
                                           the lookup header. If the
-                                          height at top model can not
-                                          be determined from the
-                                          header and is not provided
-                                          then no
-                                          "atmosphere_hybrid_height_coordinate"
-                                          dimension coordinate
-                                          construct will be created.
+                                          height can’t be determined
```
sadielbartholomew, line 472:
```suggestion
                                          height can't be determined
```

cf/read_write/um/umread.py
```diff
@@ -1203,12 +1202,37 @@ def atmosphere_hybrid_height_coordinate(self, axiscode):
...
+                    )
+                    for rec in self.z_recs
+                ]
+            )
+            if pseudolevels:
+                # Pseudolevels and atmosphere hybrid height
+                # coordinates are both present => can't reliably infer
+                # height. This is due to a current limitation in the C
+                # library that means it can ony create Z-T
+                # aggregations, rather than the requited Z-T-P
```
sadielbartholomew, line 1221:
```suggestion
                # aggregations, rather than the required Z-T-P
```

cf/read_write/um/umread.py
```diff
@@ -1203,12 +1202,37 @@ def atmosphere_hybrid_height_coordinate(self, axiscode):
             field, ac, axes=[_axis["z"]], copy=False
         )
 
-        # atmosphere_hybrid_height_coordinate dimension coordinate
-        TOA_height = bounds1.max()
-        if TOA_height <= 0:
-            TOA_height = self.height_at_top_of_model
+        # Height at top of atmosphere
+        TOA_height = self.height_at_top_of_model
```
sadielbartholomew, line 1206: Not specifically related to your changes, though in the spirit of tidying as per this PR, can we convert `TOA_height` to `toa_height` or some other similar fully non-capitalised name, because it is somewhat confusing and off the standard Python naming style to have the initial letters capitalised like an enumeration or constant name, then switching to all lower-case?

sadielbartholomew (commented):
> I think we'll get away with it if you just ignore everything in docs and its sub-directories

I took your advice and did this instead of reviewing all 7012 files! Mostly all good, but I have raised some minor comments in-line and have one general question, namely:

* wasn't the plan to deprecate `_numpy_allclose` in favour of `data.dask_utils._da_ma_allclose`? In which case there is no need to convert the internals of it?

Let's see where we are after you have responded to these comments, but overall we're probably close to being able to merge in my view.

cf/read_write/um/umread.py
davidhassell: Good idea. Fixed in https://github.com/NCAS-CMS/cf-python/pull/472/commits/df58afbcec71a6554ed40ed849d0bb9c9a52a7fa

davidhassell:
> wasn't the plan to deprecate _numpy_allclose in favour of data.dask_utils._da_ma_allclose? In which case there is no need to convert the internals of it?

Hi Sadie, I'm not sure it will be that simple, unfortunately. Mainly because `data.dask_utils._da_ma_allclose` doesn't, by design, compare masks as `_numpy_allclose` does (or, indeed, as `Data.equals` does). `_numpy_allclose` is only used once in the main code, and a few time in the tests, so I'm sure we can do something here, but perhaps not in this PR.

sadielbartholomew:
> I'm not sure it will be that simple ... so I'm sure we can do something here, but perhaps not in this PR.

Thanks for clarifying (it's all coming back to me now having written the `_da_ma_allclose` if I recall correctly), and fair enough! The updates you made to `_numpy_allclose` all look good and sensible to me and seem to work as they should, so consider the question closed, though maybe we should note down or intentions to 'do something' eventually.

Am I good to re-review or are you still making some changes after feedback?

sadielbartholomew (approved):
Previous feedback addressed and question answered, thanks. Final sanity check of updates complete on my end. All good, please merge.

PR #481, opened 2022-11-01
Fix edge case for cell methods when there are no coordinates

### LINKED ISSUE #480: Edge case for  cell methods when there are no coordinates
In a netCDF file, when there are [cell methods when there are no coordinates](https://cfconventions.org/cf-conventions/cf-conventions.html#cell-methods-no-coordinates), if there is also a 1-d coordinate with the same standard name then the cell method construct will erroneously get assigned to its axis. 

This happens because the `set_construct` method tries to match it up by inspection, which is fine "on the command line", but not so in the rarified conditions of a file read. The answer is to disable this conforming activity in `cf.read`. PR to follow.

davidhassell:
Fixes #480

Requires https://github.com/NCAS-CMS/cfdm/pull/226


davidhassell:
PR to wrong branch! closing to be be reopened.

sadielbartholomew:
Duly ignoring!

PR #483, opened 2022-11-04
Fix edge case for cell methods when there are no coordinates

### LINKED ISSUE #480: Edge case for  cell methods when there are no coordinates

davidhassell:
Fixes #480

Requires https://github.com/NCAS-CMS/cfdm/pull/226 - _edit: which has now been merged._


cf/mixin/fielddomain.py
```diff
@@ -2112,6 +2113,19 @@ def set_construct(
 
                 .. versionadded:: 3.9.0
 
+            conform: `bool`, optional
+                If True (the default), then attempt to replace
+                placeholder identities in *construct* with existing
+                construct identifiers. Specifically, cell method
+                construct axis specifiers (such as ``'T'`)` are mapped
```
sadielbartholomew, line 2120:
```suggestion
                construct axis specifiers (such as ``'T'``) are mapped
```

sadielbartholomew (approved):
Confirmed this as fixing the issue at hand. Implementation seems sensible and the new testing is sufficient to cover the case. All good apart from one typo raised so please merge once that's corrected.

PR #484, opened 2022-11-04
Minor changes to allow test_docstring.py to pass

davidhassell:
A few changes to help `test_docstring.py` pass. In particular, the `@daskified` decorators weren't playing nicely with the test, so I've removed them. They have been very useful, but I think they are no longer needed given how far along we are.

### DIFF
--- a/cf/domain.py
+++ b/cf/domain.py
@@ -576,7 +576,7 @@ class Domain(mixin.FieldDomain, mixin.Properties, cfdm.Domain):
 
         # Get the indices for every domain axis in the domain, without
         # any auxiliary masks.
-        domain_indices = self._indices(mode, None, False, **kwargs)
+        domain_indices = self._indices(mode, None, False, kwargs)
 
         return domain_indices["indices"]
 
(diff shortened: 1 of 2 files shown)


cf/functions.py
```diff
@@ -1729,6 +1731,126 @@ def _numpy_allclose(a, b, rtol=None, atol=None, verbose=None):
...
+                    size = index.size
+
+            if not keepdims and not index.ndim:
+                # Scalar array
+                continue
+        elif isinstance(index, list):
+            size = len(index)
+            if size:
+                i = index[0]
+                if i is False or i is True:
```
sadielbartholomew, line 1839:
```suggestion
                if isinstance(i, bool):
```

is equivalent (I am fairly sure?) but a bit more explicit given the unusual nature of this particular compound logical test?

sadielbartholomew (approved):
Very nice. `test_Domain` now passes for me locally, after sensible changes all suitably tested, when it raises two errors on `lama-to-dask`. I've made one minor comment but otherwise all is good, so please merge when ready.

PR #485, opened 2022-11-07
dask: Changes to allow test_docstring.py to pass

davidhassell:
A few changes to help `test_docstring.py` pass. In particular, the `@daskified` decorators weren't playing nicely with the test, so I've removed them. They have been very useful, but I think they are no longer needed given how far along we are.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -62,55 +62,8 @@ from .utils import (  # is_small,; is_very_small,
     scalar_masked_array,
 )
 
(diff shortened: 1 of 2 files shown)


sadielbartholomew (approved):
> They have been very useful, but I think they are no longer needed given how far along we are.

The time is right, I agree. They served us well but equally good riddance!

Please merge.

sadielbartholomew:
(There was one merge conflict but I have resolved it since it was related to something I recently added. Now good to merge!)

PR #486, opened 2022-11-07
dask: Set `Data._axes` in reshape and binary arithmetic

davidhassell:


cf/data/data.py
```diff
@@ -3431,13 +3431,31 @@ def _binary_operation(self, other, method):
...
+        axes = None
+        ndim0 = dx0.ndim
+        if not ndim0:
+            axes = other._axes
+        else:
+            diff = dx1.ndim - ndim0
+            if diff > 0:
+                axes = list(self._axes)
+                for _ in range(diff):
+                    axes.insert(0, new_axis_identifier(tuple(axes)))
```
sadielbartholomew, line 3444: This block is obviously very similar logically to the block with the same purpose added to `reshape`, but since there are only two cases I think, as you probably also thought when writing it up, that it's okay to have that duplication. If there are any more methods we want to add equivalent logic to, however, I do think we should create a little internal helper function to consolidate it, though.

sadielbartholomew (approved):
Good spot to include this logic on the methods in question! One comment concerning the future, but this is all good. 

Please merge.

cf/data/data.py
davidhassell: Agreed. With only two cases that are quite the same what have is OK. We'll see if more similar cases arise and decide whether to act at that time.

PR #488, opened 2022-11-08
Fix for UM stash codes that are area means returning incorrect cell methods

### LINKED ISSUE #487: UM stash codes that are area means return incorrect cell methods
When reading from disk a UM stash code that is implicitly an area mean (such as 2203: NET DN LW RAD FLUX:OPEN SEA:SEA MEAN), the resulting CF cell method is `None: mean` as opposed to the correct `area: mean`.

davidhassell:
Fixes #487

Also tidy up changelog

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,4 +1,3 @@
-<<<<<<< HEAD
 version 3.14.0
 --------------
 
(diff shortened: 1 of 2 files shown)


sadielbartholomew (approved):
All good, fixing the issue at hand. Please merge.

PR #489, opened 2022-11-08
dask: Remove deprecated tests from `test_general.py`

davidhassell:


sadielbartholomew (approved):
Sensible updates and the test in question now passes locally for me.

All good, please merge - and I've updated #428 with a listing detailed passing and failing tests across the suite, so it would be good to update that as we merge anything relevant, like this. Thanks.

PR #491, opened 2022-11-09
Raise `ValueError` upon attempt to create data-less `Data`

### LINKED ISSUE #378: LAMA to Dask: managing default(-argument) `Data` instance
(Opening as a general issue since `__init__` isn't listed under our table of methods in #295, so I am not sure of the migration status of `Data` initialisation.) Whilst I was trying to create diverse cases of instances of `Data` (in the context of https://github.com/NCAS-CMS/cf-python/pull/376#discussion_r847319179), I noticed `cf.Data()` initialised as-is, i.e. with no arguments, which is seemingly documented as being valid (initial argument `array=None` with `array: optional` stated), runs into an error:

```pycon
>>> cf.Data()
  File "/home/sadie/cf-python/cf/data/data.py", line 964, in __repr__
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/cfdm/data/data.py", line 214, in __repr__
    shape = self.shape
  File "/home/sadie/cf-python/cf/data/data.py", line 158, in wrapper
  File "/home/sadie/cf-python/cf/data/data.py", line 6288, in shape
    dx = self._get_dask()
  File "/home/sadie/cf-python/cf/data/data.py", line 1387, in _get_dask
    return self._custom["dask"]
KeyError: 'dask'
```

The fix is probably a quite simple adaptation, though I don't have time to look into it now; once I have finished reviewing my backlog of Dask migration PRs I will sort this, if you haven't already tackled it @davidhassell.

sadielbartholomew:
Closes #378.

### DIFF
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -114,6 +114,18 @@ class DataTest(unittest.TestCase):
                 "ignore", category=RuntimeWarning, message=expected_warning
             )
 
+    def test_Data__init__basic(self):
+        """Test basic `__init__` cases for Data."""
+        # Most __init__ parameters are covered by the various other tests,
+        # so this is mainly to check trivial cases and especially the edge
+        # case of 'default' Data i.e. if no parameters are specified.
+        cf.Data(0, "s")
+        cf.Data(array=np.arange(5))
+        cf.Data(source=self.filename)
+
+        with self.assertRaises(ValueError):
+            cf.Data()
+
     def test_Data_equals(self):
         """Test the equality-testing Data method."""
         shape = 3, 4
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -381,6 +381,12 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         >>> d = cf.Data(tuple('fly'))
 
         """
+        if array is None and source is None:  # don't create no/empty Data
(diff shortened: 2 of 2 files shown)


cf/test/test_Data.py
```diff
@@ -114,6 +114,18 @@ def setUp(self):
...
             )
 
+    def test_Data__init__basic(self):
+        """Test basic `__init__` cases for Data."""
+        # Most __init__ parameters are covered by the various other tests,
+        # so this is mainly to check trivial cases and especially the edge
+        # case of 'default' Data i.e. if no parameters are specified.
+        cf.Data(0, "s")
+        cf.Data(array=np.arange(5))
+        cf.Data(source=self.filename)
```
davidhassell, line 124: An interesting choice of `source` (i.e. a string), but it certainly proves the test :)

sadielbartholomew: Thanks for clarifying David. That all makes sense.

> ... so we should perhaps consider improving this docstring in another PR ...

Good idea. I'll raise an Issue linking to this thread, update the line in question to specify something normal and sensible for the `source` parameter, and merge the PR.

sadielbartholomew:
(One of the CI jobs has failed due to multiple `OSError: Can't read non-existent file` type errors, which seems unrelated so I will look into separately.)

PR #494, opened 2022-11-11
dask: Cached first, second and last data values

### LINKED ISSUE #493: Discussion: Managing `Data.__str__` with `dask`
This issue is for discussing ways in which `Data.__str__` can be made to perform nicely when its data is stored in a `dask` array.

#### Exposition

The string representation of a `Data` object is currently inherited from cfdm, and looks like:
```python
>>> import cf
>>> d = cf.example_field(0).data
>>> str(d)
[[0.007, ..., 0.013]] 1
```
I.e. it prints the first and last elements (and the second element if there only 3 of them).

With `dask` representing the data, and using the code inherited from cfdm with no changes, printing these elements could

1. trigger an expensive and slow computation
2. require the reading from disk of an entire dask chunk _per element printed_. If each chunk has the default size of 128 MiB, then that could entail reading 256 MiB from disk just to print two numbers.

davidhassell:
Partially fixes #493


davidhassell:
New changes done - ready for review. Thanks.

cf/test/test_Data.py
```diff
@@ -4316,6 +4316,36 @@ def test_Data__init__datetime(self):
         self.assertTrue((q == d).array.all())
         self.assertTrue((d == q).array.all())
 
+    def test_Data__str__(self):
+        """Test `Data.__str__`"""
+        elements0 = ("first_element", "last_element", "second_element")
+        for array in ([1], [1, 2], [1, 2, 3]):
```
sadielbartholomew, line 4322: Ideally we could extend this test method to also include an example array input for `Data` with size larger than three to test that in such cases the `second_element` isn't included in `_custom` in this way and that an `Ellipsis` replaces it in the `str` representation.

Obviously it isn't as simple as adding it to this list to iterate over, but it it might be simple enough to adjust this block with some conditionals to allow that to work with some tweaks, else write a separate block to cover it?

sadielbartholomew (approved):
Achieves the intended feature as described in https://github.com/NCAS-CMS/cf-python/issues/493#issuecomment-1311675048, with a neat implementation and sufficient testing. I've made one suggestion that may be taken or left (not a blocker). Please merge when ready!

cf/test/test_Data.py
davidhassell: Excellent idea: https://github.com/NCAS-CMS/cf-python/pull/494/commits/71a8bf705e2927396aa1a32e56f146936cc3d524

PR #496, opened 2022-11-14
Write omit data

### LINKED ISSUE #477: Allow data to be omitted from netCDF files during `cf.write`
(Copied from https://github.com/NCAS-CMS/cfdm/issues/221)

It would be useful to be able to not write to disk the data from selected construct types.

This would not affect the amount of netCDF variables and dimensions that are written to the file, nor the netCDF variables' attributes, but would not create data on disk for the requested variables thus resulting in a smaller file than it otherwise would have been.

When the new file is read then the data of these variables will be represented by an array of all missing data.

This will be useful for cases when there is a need to store a file's metadata structure for inspection, but the data values are immaterial.

----

This is implemented in `cfdm v1.10.0.1`, and just needs the `omit_data` keyword to be brought into `cf.write`.

davidhassell:
Fixes #477


sadielbartholomew (approved):
Perfect, please merge.

(Note that we can, in certain cases, observe the pair of issues raised over on `cfdm` downstream here when files are re-read after being written without data using this new keyword, namely NCAS-CMS/cfdm#224 and NCAS-CMS/cfdm#225, but when those are fixed in `cfdm` the fix should pass through here to cf).

PR #497, opened 2022-11-14
Move test to `ppTest`

davidhassell:
Fix up the PP read test from `test_read_write.py` and move it to its spiritual home in `test_pp.py`.


sadielbartholomew (approved):
Fixes via changing the `"endian": "big"` value to `"little"`, with both affected tests passing locally for me and the move of the test method making good sense. All good, please merge.

PR #502, opened 2022-11-17
dask: Subsampled compression

davidhassell:
Small fixes to allow `test_subsampling.py` to pass, that essentially just forward port a few things from cfdm.


cf/test/test_subsampling.py
```diff
@@ -0,0 +1,249 @@
...
+                lat0, ignore_data_type=True, rtol=0, atol=9e-5, verbose=-1
+            )
+        )
+        self.assertTrue(
+            lon.equals(
+                lon0, ignore_data_type=True, rtol=0, atol=9e-5, verbose=-1
+            )
+        )
+
+    @unittest.skipIf(True, "TODO: awaiting test file")
```
sadielbartholomew, line 180: Any news on when this test file might become available? I seem to recall (for the `cfdm` case, I guess the same file will be used for both libraries) you asked someone to provide something for this specific case and were waiting on that (though that could be my memory going wrong)?

cf/test/test_subsampling.py
```diff
@@ -0,0 +1,249 @@
...
+            os.remove(f)
+        except OSError:
+            pass
+
+
+atexit.register(_remove_tmpfiles)
+
+
+class SubsampledTest(unittest.TestCase):
+    """Test management of constructs with underlying subsampled arrays."""
```
sadielbartholomew, line 33: The linting checks use a tool (`docformatter`, as listed on the output report) that checks the docstring summary line length, and this one is just too long (probably my fault from adding docstrings to the tests), so for a full linting pass if we can just tweak this to shorten it a bit. How about:

```suggestion
    """Test management of underlying subsampled arrays."""
```

?

sadielbartholomew (approved):
I've raised one question and a few minor comments, so please check those out, but this is all good. Please merge when ready.

cf/test/test_subsampling.py
davidhassell: Ah yes - still waiting, although I have been in touch. I'll chase it up again ...

sadielbartholomew: Thanks for clarifying. Obviously it would be good to not leave it too long else it might fall off our radar and then never get tested overall...

PR #506, opened 2022-11-22
dask: Fix for cached str-type data elements

davidhassell:
Minor bug fix for the new element caching for `Data.__str__`.


sadielbartholomew (approved):
A justified change, with apologies for missing the subtlety in question during code review for #494. All good, please merge!

To note, I'm noticing the occasional segmentation fault on the `test_read_write` test (it's hard to estimate at what frequency as, all in the same environment e.g. desktop or laptop, they tend to happen often over short periods and then not at all for some reason). I did check by running it on the `lama-to-dask` branch too that it isn't something related to this change, and indeed I see the seg faults intermittently on that too. (I'm not sure if I was seeing those on there recently so far as I remember, but don't think I was, so perhaps a recent change has led to this? Are you seeing seg faults at all?)

PR #507, opened 2022-12-01
Dask subarray graph structure - give subarrays/fragment arrays their own keys

davidhassell:
When creating dask graphs ab initio with `da.Array`, it is important that any subarrays/fragments have their own dask key, so that they can be located by `get_filenames`.


sadielbartholomew (approved):
The motivation makes sense and this is a sensible approach to that. Tested as working (again we can ignore the CI job failures which are erroring for unrelated reasons). All good, please merge.

PR #510, opened 2022-12-06
dask: Fix for cached elements from size-0 arrays

davidhassell:
During `cf.read`, prevents cached data elements from being gleaned from size-0 arrays. Allows `read_writeTest.test_read_broken_bounds` to pass.


sadielbartholomew (approved):
Justified change, with a nice clear comment to broadcast as such, confirmed as working for me locally (`test_read_write.py` now passes fully so we can mark that as passing now on #428!). Lovely, please merge.

PR #517, opened 2022-12-08
dask: Correctly read VLEN string variables

davidhassell:
Fix for incorrect reading of VLEN string variables from netCDF4 files. I don't know why this was failing, but new change reuses code from cfdm, rather than cutting and pasting, which has to be better! This allows `test_groups.py` to pass.

Need https://github.com/NCAS-CMS/cfdm/pull/235 to be merged first. _Edit: which has now happened_


sadielbartholomew (approved):
> but new change reuses code from cfdm, rather than cutting and pasting, which has to be better! 

Definitely better!

> This allows test_groups.py to pass.

This passes locally for me. All the changes look sensible. I've raised a few minor comments but otherwise this is good to merge in my eyes so please go ahead when ready.

PR #518, opened 2022-12-12
dask: Remove `FileArrayMixin._dask_lock`

davidhassell:
Remove `FileArrayMixin._dask_lock`. The property was returning `True`, which created a lock on chunk reads in `data.creation.to_dask`, preventing concurrent reads from data on disk and effectively creating turning off parallelism by forcing sequential behaviour.

In the absence of a `_dask_lock`, the lock passed to `dask.array.from_array` is `False`, which allows concurrent reads.

### DIFF
--- a/cf/data/array/mixin/filearraymixin.py
+++ b/cf/data/array/mixin/filearraymixin.py
@@ -9,21 +9,15 @@ class FileArrayMixin:
     """
 
     @property
-    def _dask_lock(self):
-        """TODODASKDOCS.
-
-        Concurrent reads are assumed to be not supported.
-
-        .. versionadded:: TODODASKVER
-
-        """
-        return True
-
-    @property
     def _dask_meta(self):
-        """TODODASKDOCS.
+        """The metadata for the containing dask array.
+
+        This is the kind of array that will result from slicing the
+        file array.
 
         .. versionadded:: TODODASKVER
+
+        .. seealso:: `dask.array.from_array`
 
         """
         return np.array((), dtype=self.dtype)


sadielbartholomew:
Nicely spotted. I'll review this immediately since investigating the reason for the much slower than anticipated performance is pretty critical!

sadielbartholomew (approved):
> Remove FileArrayMixin._dask_lock. The property was returning True

I used some `print` statements to inspect the `lock` status on the `lama-to-dask` branch and on this one, to compare, and didn't see any evidence of the `lock` being `True` across any of the tests in either case, notably the former case -  not sure if I am missing something here? FYI I was checking the above using:

```diff
diff --git a/cf/data/creation.py b/cf/data/creation.py
index eae01b907..37066329b 100644
--- a/cf/data/creation.py
+++ b/cf/data/creation.py
@@ -126,10 +126,12 @@ def to_dask(array, chunks, default_chunks=False, **from_array_options):
     kwargs.setdefault("meta", getattr(array, "_dask_meta", None))
 
     try:
+        print(kwargs["lock"], kwargs["meta"])
         return da.from_array(array, chunks=chunks, **kwargs)
     except NotImplementedError:
         # Try again with 'chunks=-1', in case the failure was due to
         # not being able to use auto rechunking with object dtype.
+        print("EXCEPT BLOCK", kwargs["lock"], kwargs["meta"])
         return da.from_array(array, chunks=-1, **kwargs)
```

Regardless, this seems like a sensible change, so please merge if you are happy (I'm just not convinced much if anything has changed after this changeset, from what I can see...).

davidhassell:
Hi Sadie, Try putting `print('lock=',lock, repr(array))` just after the line `lock = getattr(array, "_dask_lock", False)` - does that show some "True"s in lama-to-dask?

sadielbartholomew:
> does that show some "True"s in lama-to-dask?

Ah yes, I do see some `True` values coming through now in `test_Field`, though they are rarer than I was anticipating, so maybe that was the problem at hand (selective blindness)! Thanks for clarifying. All good (and understood too now!), please merge.

davidhassell:
Glad you see them too! They should only have been appearing when the dask array is built on a `NetCDFArray` instance, which is indeed less common in the test suite. Merging now.

PR #523, opened 2022-12-13
Fix `DimensionCoordinate.create_bounds` for cyclic coordinates & improve docs

### LINKED ISSUE #514: Incorrect Voronoi bounds from `DimensionCoordinate.create_bounds` for cyclic coordinates
At version 3.13.1, incorrect Voronoi bounds are created from `DimensionCoordinate.create_bounds` for cyclic and period coordinates. 
```python
>>> import cf
>>> cf.__version__
'3.13.1'
>>> d = cf.DimensionCoordinate(data=[180, 90, 40, 0.0])   
>>> d.period(360)
>>> d.cyclic(0)
set()
>>> b = d.create_bounds()
>>> print(b.array) 
[[225. 135.]   # Wrong
 [135.  65.]
 [ 65.  20.]
 [ 20. -20.]]  # Wrong
```
when the result should be
```python
[[270. 135.]   # Correct
 [135.  65.]
 [ 65.  20.]
 [ 20. -90.]]  # Correct
```
Also, the docstring for the method is lacking in detail.

davidhassell:
Fixes #514

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -13,7 +13,7 @@ version 3.14.0
 * Fixed bug that raised an exception when using the ``equal`` or
   ``exist`` keyword of `cf.aggregate`
   (https://github.com/NCAS-CMS/cf-python/issues/499)
-* Fixed bug the created incorrect cyclic Voronoi bounds from
+* Fixed bug that created incorrect cyclic Voronoi bounds from
   `DimensionCoordinate.create_bounds`
   (https://github.com/NCAS-CMS/cf-python/issues/514)
 * Changed dependency: ``1.10.0.1<=cfdm<1.10.1.0``
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -11,6 +11,9 @@ version 3.14.0
 * Fixed bug that raised an exception when using the ``equal`` or
   ``exist`` keyword of `cf.aggregate`
   (https://github.com/NCAS-CMS/cf-python/issues/499)
+* Fixed bug the created incorrect cyclic Voronoi bounds from
+  `DimensionCoordinate.create_bounds`
+  (https://github.com/NCAS-CMS/cf-python/issues/514)
 * Changed dependency: ``1.10.0.1<=cfdm<1.10.1.0``
 * New dependency: ``dask>=2022.6.0``
 
--- a/cf/dimensioncoordinate.py
+++ b/cf/dimensioncoordinate.py
@@ -339,140 +339,77 @@ class DimensionCoordinate(
     ):
         """Create cell bounds.
 
-        The returned `Bounds` instance can always be added to the
-        dimension coordinate construct with its `set_bounds` method,
-        for instance:
+        Creates new cell bounds, irrespective of whether the cells
+        already have cell bounds. The new bounds are not set on the
+        dimension coordinate construct, but if that is desired they
+        may always be added with the `set_bounds` method, for
+        instance:
 
         >>> b = d.create_bounds()
         >>> d.set_bounds(b)
 
         By default, Voronoi cells are created by defining cell bounds
         that are half way between adjacent coordinate values. For
-        non-cyclic coordinates, extrapolation is used to ensure that
-        the first and last cells have their coordinates in their
-        centre.
+        non-cyclic coordinates, extrapolation is used so that the
+        first and last cells have their coordinates in their centres.
 
         For instance, the Voronoi cells created for non-cyclic
         coordinates ``0, 40, 90, 180`` have bounds ``(-20, 20), (20,
-        65), (65, 135), (135, 225)``
+        65), (65, 135), (135, 225)``. If those coordinates were
+        instead cyclic with a period of ``360``, then the bounds would
+        be ``(-90, 20), (20, 65), (65, 135), (135, 270)``.
 
-        If those coordinates were instead cyclic with a period of
-        ``360``, then the bounds would be ``(-90, 20), (20, 65), (65,
-        135), (135, 270)``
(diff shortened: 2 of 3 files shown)


cf/dimensioncoordinate.py
```diff
@@ -339,155 +339,223 @@ def create_bounds(
...
+                *Parameter example:*
+                  For coordinates ``1, 2, 10``, setting ``bound=0.5``
+                  will result in bounds of ``(0.5, 1.5), (1.5, 2.5),
+                  (2.5, 17.5)``.
 
                 *Parameter example:*
-                    For coordinates ``1, 2, 10``, setting ``cellsize=1,
-                    flt=0.5`` will result in bounds of ``(0.5, 1.5), (1.5,
-                    2.5), (9.5, 10.5)``.
+                  Beware that if *bound* is too large (small), then
```
sadielbartholomew, line 380:
```suggestion
                  Beware that if *bound* is too large or small, then
```

since in this case there is no matched component in parentheses later in the sentence, so it is somewhat confusing?

cf/dimensioncoordinate.py
```diff
@@ -339,155 +339,223 @@ def create_bounds(
...
+                contigious. The *cellsize* parameter may be one of:
+
+                * data_like scalar
+
+                  Defines the cell size, either in the same units as
+                  the coordinates or in the units provided. Note that
+                  in this case, the position of each coordinate within
+                  the its cell is determined by the *flt* parameter.
+
+            ..
```
sadielbartholomew, line 398: Quick question: what do you use these `..` for, I've always wondered (or assumed they were added by mistake, but I doubt so now)?

cf/dimensioncoordinate.py
```diff
@@ -339,155 +339,223 @@ def create_bounds(
...
+                  to specify cell size of one calendar month, starting
+                  and ending on the 15th day:
+                  ``cellsize=cf.M(day=15)`` (see `cf.M` for
+                  details). Note that the *flt* parameter is ignored
+                  in this case.
+
+            flt: array_like scalar, optional
+                When creating cells with sizes specified by the
+                *cellsize* parameter, define the fraction of the each
+                cell which is less its coordinate value. By default
```
sadielbartholomew, line 415:
```suggestion
                *cellsize* parameter, define the fraction of each
                cell which is less than its coordinate value. By default
```

cf/dimensioncoordinate.py
```diff
@@ -339,155 +339,223 @@ def create_bounds(
...
+                *flt* is 0.5, so that each cell has its coordinate at
+                it's centre. Ignored if *cellsize* is not set.
```
sadielbartholomew, line 417:
```suggestion
                its centre. Ignored if *cellsize* is not set.
```

cf/dimensioncoordinate.py
```diff
@@ -578,17 +647,34 @@ def create_bounds(
...
-                        bounds_1d[-1] = max
-                    else:
-                        bounds_1d[0] = max
-                if min is not None:
+                if self.cyclic():
+                    period = self.period()
+                    if period is None:
+                        raise ValueError(
+                            "Can't create Voronoi bounds for cyclic "
+                            "corodinates with no period"
```
sadielbartholomew, line 655:
```suggestion
                            "coordinates with no period"
```

cf/dimensioncoordinate.py
```diff
@@ -339,155 +339,223 @@ def create_bounds(
...
+        >>> b = d.create_bounds()
+        >>> print(b.array)
+        [[270. 135.]
+         [135.  65.]
+         [ 65.  20.]
+         [ 20. -90.]]
+
+        Cellsize units must be equivalent to the coordinate units, or
+        if the cellhas has no units then they are assumed to be the
+        same the coordinates:
```
sadielbartholomew, line 483:
```suggestion
        Cellsize units must be equivalent to the coordinate units, or
        if the cell has no units then they are assumed to be the
        same as the coordinates:
```

cf/test/test_DimensionCoordinate.py
```diff
@@ -484,14 +484,118 @@ def test_DimensionCoordinate_set_bounds(self):
...
+                        [65.0, 20.0],
+                        [20.0, -90.0],
+                    ]
+                )
+            ).all()
+        )
+
+        # Cellsize units must be equivalent to the coordinate units,
+        # or if the cellhas has no units then they are assumed to be
+        # the same the coordinates:
```
sadielbartholomew, line 549:
```suggestion
        # Cellsize units must be equivalent to the coordinate units,
        # or if the cell has no units then they are assumed to be
        # the same as the coordinates:
```

(equivalent to the changes made in the docstring, if that saves you work checking...)

sadielbartholomew (approved):
Confirmed as a logical fix for the bug with a very thorough and clear update to the docstring :100:. I've highlighted some typos (and asked one random question) but other than those this is good to merge in my eyes. Thanks.

Changelog.rst
```diff
@@ -13,6 +13,9 @@ version 3.14.0
 * Fixed bug that raised an exception when using the ``equal`` or
   ``exist`` keyword of `cf.aggregate`
   (https://github.com/NCAS-CMS/cf-python/issues/499)
+* Fixed bug the created incorrect cyclic Voronoi bounds from
```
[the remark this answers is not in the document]
davidhassell: I'll check the Change log - as that's where I probably copied it from!

cf/dimensioncoordinate.py
sadielbartholomew: Aha, that makes sense. Thanks for explaining.

PR #524, opened 2022-12-13
Set module in `unittest.main` for all test modules

sadielbartholomew:
A quick PR to allow wrapping of any, or all, of our test modules by profiling tools etc. (which I realise isn't the best way to gauge performance but still can be useful as an investigation tool) by explicitly setting [the `module` argument to `unittest.main`](https://docs.python.org/3/library/unittest.html#unittest.main) which produces no difference in behaviour when run in the standard/direct way, but enables all of the test methods, rather than none due to the module not being recognised, to be run when the script is wrapped.

To avoid having to set all of the filenames individually to get e.g. `module="test_Data"`, I did a `sed` replacement via `$ find . -type f | xargs sed -i 's/unittest\.main(/unittest\.main(module=__file__\.split("\.")\[0\], /g'` under the `test` directory.

@davidhassell: requesting a review despite this being quite a trivial PR, please confirm whether or not this is going to cause any problems that I might not have foreseen. Thanks.

### DIFF
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -4415,4 +4415,4 @@ if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
--- a/cf/test/create_test_files.py
+++ b/cf/test/create_test_files.py
@@ -5299,4 +5299,4 @@ if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
--- a/cf/test/setup_create_field.py
+++ b/cf/test/setup_create_field.py
@@ -209,4 +209,4 @@ if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
--- a/cf/test/test_AuxiliaryCoordinate.py
+++ b/cf/test/test_AuxiliaryCoordinate.py
@@ -339,4 +339,4 @@ if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
--- a/cf/test/test_CellMeasure.py
+++ b/cf/test/test_CellMeasure.py
@@ -40,4 +40,4 @@ if __name__ == "__main__":
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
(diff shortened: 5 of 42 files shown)


davidhassell:
Hi Sadie, thanks. The changes look benign to me (although I don't know quite what they do!). I presume that this PR doesn't do any profiling, right? (and hence slow down the tests?). How might you wrap them to run them with profiling?

cf/test/test_Data.py
```diff
@@ -4415,4 +4415,4 @@ def test_Data__str__(self):
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(verbosity=2)
+    unittest.main(module=__file__.split(".")[0], verbosity=2)
```
sadielbartholomew, line 4418: Hi David, sorry for not being clear. As an example of using `cProfile` for this test module, with the change to the line here we see 134 tests run as expected, but without the `module` specified it just shows this, for the same command:

```console
$ python -m cProfile -o datatest.prof test_Data.py 
Run date: 2022-12-13 14:58:43.777883
Platform: Linux-4.15.0-54-generic-x86_64-with-glibc2.10 
HDF5 library: 1.10.6 
netcdf library: 4.8.0 
udunits2 library: /home/sadie/anaconda3/envs/cf-env/lib/libudunits2.so.0 
Python: 3.8.10 /home/sadie/anaconda3/envs/cf-env/bin/python
netCDF4: 1.5.6 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/netCDF4/__init__.py
cftime: 1.6.0 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/cftime/__init__.py
numpy: 1.22.2 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/numpy/__init__.py
psutil: 5.9.0 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/psutil/__init__.py
scipy: 1.8.0 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/scipy/__init__.py
matplotlib: 3.4.3 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/matplotlib/__init__.py
ESMF: 8.1.1 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/ESMF/__init__.py
cfdm: 1.10.0.1 /home/sadie/cfdm/cfdm/__init__.py
cfunits: 3.3.5 /home/sadie/cfunits/cfunits/__init__.py
cfplot: 3.1.18 /home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/cfplot/__init__.py
cf: 3.14.0b0 /home/sadie/cf-python/cf/__init__.py


Ran 0 tests in 0.000s

OK
```

basically because `unittest.main` looks for tests in the default `module='__main__'`. So when it gets run via, say, `cProfile`, the `module` name isn't the same and so it doesn;t pick up on the test methods in the test module in question. Same when wrapping the full `run_tests.py`. Does that make sense?

sadielbartholomew: > I presume that this PR doesn't do any profiling, right? (and hence slow down the tests?)

And yes, no profiling is done! These changes just allow us to do some profiling via the command line, if we wish. Otherwise it makes no difference to anything, unless you can see any hidden issues!

davidhassell: Got it - thanks for the explanations. This is all good for me, so please merge when you're ready.

PR #528, opened 2022-12-16
dask: Force from_array lock=False for nested arrays in compressed chunks

davidhassell:
The recent change to the default dask file lock (#518) causes problems with compressed data: It's currently set up so that the each dask chunk has to compute another dask array. It does this nested compute with the synchronous scheduler - which it seemed to like - but now that `lock=False` it doesn't like it at all. 

The solution (well, hack) I have come up with involves redefining the other dask array with `da.from_array(..., lock=True)` and then all is well.

This is not great, and all of these compressed `to_dask_array` methods should really, if possible, integrate the other dask array into a single dask graph. That's one for later (post 3.14.0).

### DIFF
--- a/cf/data/array/gatheredarray.py
+++ b/cf/data/array/gatheredarray.py
@@ -1,9 +1,9 @@
 import cfdm
 
-from .mixin import ArrayMixin
+from .mixin import ArrayMixin, CompressedArrayMixin
 
 
-class GatheredArray(ArrayMixin, cfdm.GatheredArray):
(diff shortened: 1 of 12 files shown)


sadielbartholomew (approved):
Well-spotted RE the underlying issue. You've acknowledged that the fix in this PR is not a robust long-term one and that has been justified so I'm happy too. I'll open a quick issue to reference your intentions to refactor eventually. All of the tests that you have touched here pass fully for me locally.

So all is good! Please go ahead and merge.

PR #531, opened 2022-12-19
dask: Reinstate netCDF lock

davidhassell:
It turns out that removing the concurrency lock for opening netCDF files is not (yet) safe in the netCDF-C library, so we should put it back for now. I had thought that separate opens/closes would be OK, but not so. This may be fixed in future version of netCDF-C. See, e.g. https://github.com/Unidata/netcdf4-python/issues/844 and https://github.com/pydata/xarray/issues/4100.

_Edit:_ The `test_Field_collapse` are still segfaulting intermittently, but less frequently with this. I think it may be due to the fact that each `from_array` call is creating its own lock, so its still possible for concurrent accesses that originate from the same file, e.g. `f + f.copy()`. The answer could be to create a common lock object and coordinate all access to a particular file around that. 

_Edit:_ Following this change`test_Field_collapse` (with the concatenate fixes from another PR) didn't segfault 18 times in row before I stopped counting - record!

_Edit:_ I deleted the three functions from `creation.py` because they're not ever used, and were only there because I thought they would be useful over a year ago, when I didn't really know what I was doing.

### DIFF
--- a/cf/data/array/netcdfarray.py
+++ b/cf/data/array/netcdfarray.py
@@ -1,5 +1,7 @@
 import cfdm
 
+from dask.utils import SerializableLock
+
 from .mixin import FileArrayMixin
 
 
@@ -18,10 +20,18 @@ class NetCDFArray(FileArrayMixin, cfdm.NetCDFArray):
     def _dask_lock(self):
         """Set the lock for use in `dask.array.from_array`.
 
-        Returns `True` because concurrent reads are not currently
-        supported by the netCDF-C library.
+        Returns a lock object (unless no file name has been set)
+        because concurrent reads are not currently supported by the
+        netCDF-C library. The lock object will be the same for all
+        `NetCDFArray` instances with this file name, which means that
+        all file access coordinates around the same lock.
 
         .. versionadded:: TODODASKVER
 
         """
-        return True
+        filename = array.get_filename(None)
+        if filename is None:
+            return False
+        
+        return SerializableLock(filename)
+        
--- a/cf/data/array/netcdfarray.py
+++ b/cf/data/array/netcdfarray.py
@@ -18,7 +18,7 @@ class NetCDFArray(FileArrayMixin, cfdm.NetCDFArray):
     def _dask_lock(self):
         """Set the lock for use in `dask.array.from_array`.
 
-        Returns `true` because concurrent reads are not currently
+        Returns `True` because concurrent reads are not currently
         supported by the netCDF-C library.
 
         .. versionadded:: TODODASKVER
--- a/cf/data/array/netcdfarray.py
+++ b/cf/data/array/netcdfarray.py
@@ -23,7 +23,7 @@ class NetCDFArray(FileArrayMixin, cfdm.NetCDFArray):
         which case `False` is returned) because concurrent reads are
         not currently supported by the netCDF-C library. The lock
         object will be the same for all `NetCDFArray` instances with
-        this file name, which means that all file access coordinates
+        this file name, which means that all files access coordinates
         around the same lock.
 
         .. versionadded:: TODODASKVER
--- a/cf/data/array/netcdfarray.py
+++ b/cf/data/array/netcdfarray.py
@@ -1,5 +1,4 @@
 import cfdm
-
 from dask.utils import SerializableLock
 
 from .mixin import FileArrayMixin
@@ -20,18 +19,18 @@ class NetCDFArray(FileArrayMixin, cfdm.NetCDFArray):
     def _dask_lock(self):
         """Set the lock for use in `dask.array.from_array`.
 
-        Returns a lock object (unless no file name has been set)
-        because concurrent reads are not currently supported by the
-        netCDF-C library. The lock object will be the same for all
-        `NetCDFArray` instances with this file name, which means that
-        all file access coordinates around the same lock.
+        Returns a lock object (unless no file name has been set, in
+        which case `False` is returned) because concurrent reads are
+        not currently supported by the netCDF-C library. The lock
+        object will be the same for all `NetCDFArray` instances with
+        this file name, which means that all file access coordinates
+        around the same lock.
 
         .. versionadded:: TODODASKVER
 
         """
-        filename = array.get_filename(None)
+        filename = self.get_filename(None)
         if filename is None:
             return False
-        
+
         return SerializableLock(filename)
-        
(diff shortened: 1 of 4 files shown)


sadielbartholomew:
@davidhassell am I good to start reviewing or might you still make some (pre-feedback) tweaks?

cf/data/array/netcdfarray.py
```diff
@@ -13,3 +14,23 @@ def __repr__(self):
...
+
+    @property
+    def _dask_lock(self):
+        """Set the lock for use in `dask.array.from_array`.
+
+        Returns a lock object (unless no file name has been set, in
+        which case `False` is returned) because concurrent reads are
+        not currently supported by the netCDF-C library. The lock
+        object will be the same for all `NetCDFArray` instances with
+        this file name, which means that all file access coordinates
```
sadielbartholomew, line 26:
```suggestion
        this file name, which means that all files access coordinates
```

cf/data/creation.py
```diff
@@ -3,81 +3,47 @@
...
+        from_array_options: `dict`, optional
             Keyword arguments to be passed to `dask.array.from_array`.
 
+            If *from_array_options* has no ``'lock'`` key then the
+            `lock` keyword is set to the `_dask_lock` attribute of
+            *array* or, if there is no such attribute, `False`.
+
+            If *from_array_options* has no ``'meta'`` key then the
+            `meta` keyword is set to the `_dask_meta` attribute of
+            *array* or, if there is no such attribute, `None`.
```
sadielbartholomew, line 41: Is there a particular reason we have the former becoming `False` whereas the latter becoming `None` "if there is no such attribute" - I guess it stems from the [different defaults in input and motivation for the `meta` and `lock` keywords to `da.from_array`](https://docs.dask.org/en/stable/generated/dask.array.from_array.html#dask-array-from-array)), in which case maybe we can briefly state that to be clear that this is the reason we default to those too? Something like:

```suggestion
            If *from_array_options* has no ``'lock'`` key then the
            `lock` keyword is set to the `_dask_lock` attribute of
            *array* or, if there is no such attribute, `False`,
            the value it is assigned to by default in
            `dask.array.from_array`.

            If *from_array_options* has no ``'meta'`` key then the
            `meta` keyword is set to the `_dask_meta` attribute of
            *array* or, if there is no such attribute, `None`,
            the value it is assigned to by default in
            `dask.array.from_array`.
```

cf/test/test_read_write.py
```diff
@@ -883,4 +883,4 @@ def test_write_omit_data(self):
     print("Run date:", datetime.datetime.now())
     cf.environment()
     print()
-    unittest.main(module=__file__.split(".")[0], verbosity=2)
+    unittest.main(verbosity=2)
```
sadielbartholomew, line 886: Sorry I forgot to fix all of these cases last Friday. I'll put up a PR to sort it today!

sadielbartholomew (approved):
It's difficult for me to test the aspect of reducing the observed segmentation faults since I haven't seen any for a few weeks (the darn things, only showing up when they are not wanted and never when they are!) but the tests we know should pass still do, with no noticeable speed compromise (no need to profile here), and the lock restoration and other changes made are all well-justified, so I am happy. I've made a few minor comments; please merge when ready.

cf/data/creation.py
sadielbartholomew: In hindsight, maybe this is over the top. Let me know what you think...

davidhassell:
Hi @sadielbartholomew  - I was thinking that maybe one lock _per file_ might not be safe enough, and that one lock _for all files_ is what's needed. I'll run a few tests that with the code that's already here that combines two different netCDF files in the same dask graph to see if they have a propensity to seg fault ....

cf/data/creation.py
sadielbartholomew: Fair enough, marking as resolved!

sadielbartholomew:
> I'll run a few tests that with the code that's already here that combines two different netCDF files in the same dask graph to see if they have a propensity to seg fault ....

Right-o. Makes sense (I think). If you go for that in the end, maybe you can attack that in a separate PR to separate out the steps here, as this aspect is done and we can keep the ext bit self-contained?

davidhassell:
Hi Sadie - If we need a global lock, this will change code here, rather than add to it, so maybe I'll run the tests before merging ...

davidhassell:
Hi Sadie, I've tested with multiple files and the one-lock-per-file and it seems safe, i.e. no seg faults.

However, I've been looking at how xarray deals with locks (https://github.com/pydata/xarray/blob/main/xarray/backends/netCDF4_.py, https://github.com/pydata/xarray/blob/main/xarray/backends/locks.py) and they appear to do a stricter job - applying a global lock for all files, as far as I can tell after a short look.

I'd like to investigate a bit more how xarray does things before settling this.

sadielbartholomew:
> I'd like to investigate a bit more how xarray does things before settling this

Sounds wise, if they do it that way there is probably a good reason so! I'll keep an eye out for any further comments here, although give me a shout if there's anything I can do RE aspects relating to this PR otherwise.

davidhassell:
So, `xarray` certainly applies a global lock across all netCDF files, and this is the right thing to do because the HDF5 library modifies global data structures that are _independent of a particular HDF5 dataset or HDF5 file_ (https://stackoverflow.com/questions/34906652/does-hdf5-support-concurrent-reads-or-writes-to-different-files).

That same stack overflow article also mentions that it is possible to compile HDF5 in with MPI support but no thread safety, or with thread safety but no MPI support. However the default (like most folk get from conda) is neither. 

So, I think that we we probably do need a global lock for netCDF files, and this is an easy change: https://github.com/NCAS-CMS/cf-python/pull/531/commits/77ea4fad8f5c07d468617fb87cc802b3f9719269

sadielbartholomew:
> So, I think that we we probably do need a global lock for netCDF files, and this is an easy change: https://github.com/NCAS-CMS/cf-python/commit/77ea4fad8f5c07d468617fb87cc802b3f9719269

Nice, with thanks for clarifying on the reasons! So this is good for re-review now?

davidhassell:
Thanks, Sadie - yes, good to review

sadielbartholomew:
> That same stack overflow article...

Excellent researching by the way :laughing: (joking about the standalone phrase above which amused me, though actually that one is well-linked to canonical sources so it's reliable enough as I guess you thought too)

sadielbartholomew (approved):
Justified and working update to instead use a global lock to cover all netCDF files. The one typo as feedback from the first review has re-appeared, however, but if you can blitz that this is all good and ready to merge. (May we never see another seg fault in the wild again! :crossed_fingers: )

PR #536, opened 2022-12-20
dask: Remove `is_small` and `is_very_small`

davidhassell:
Removes functions `is_small` and `is_very_small` from `data/utils.py`, which were envisaged right at the start , but never implemented, and we now know are not  needed


sadielbartholomew (approved):
> which were envisaged right at the start , but never implemented, and we now know are not needed

I can't actually remember or work out from the docstrings why there were put there in the first place!

Always good to trim some fat from the codebase. All good, please merge.

PR #537, opened 2022-12-21
dask: Remove `Data.force_compute` and `Data.can_compute`

davidhassell:
Removes methods `Data.force_compute` and `Data.can_compute`, which were envisaged right at the start but are now needed.


sadielbartholomew (approved):
Perfect, please merge (assuming you meant "now *not* needed" with:

> which were envisaged right at the start but are now needed

!)

davidhassell:
_not_ need, indeed :)

PR #538, opened 2022-12-21
LAMA to Dask: update `cf.environment` with new dependencies

sadielbartholomew:
Upgrade `cf.environment` to be ready and correct for `3.14.0`, by:

* adding the two new dependency libraries (i.e. relative additions to `requirements.txt`) to the output, which notably includes `dask`, the package information on which might prove very useful with inevitable bug reports that will come in for the `.0` release!
* reordering a few libraries in the output listing, such that any C/Fortran underlying libraries are given first, then the NCAS CF Tools libraries at the very end (see comments on the corresponding dictionary summarising this).

### DIFF
--- a/cf/functions.py
+++ b/cf/functions.py
@@ -2984,38 +2984,44 @@ def environment(display=True, paths=True):
     **Examples**
 
     >>> cf.environment()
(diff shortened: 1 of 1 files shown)


davidhassell (approved):
Perfect. Please merge!

PR #539, opened 2022-12-22
dask: Deal with TODODASKMSG placeholders

davidhassell:
Deal with TODODASKMSG placeholders.

### DIFF
--- a/cf/data/mixin/deprecations.py
+++ b/cf/data/mixin/deprecations.py
@@ -312,6 +312,9 @@ class DataClassDeprecationsMixin:
     def chunk(self, chunksize=None, total=None, omit_axes=None, pmshape=None):
         """Partition the data array.
 
+        Deprecated at version TODODASKVER. Use the `rechunk` method
+        instead.
+
         :Parameters:
 
             chunksize: `int`, optional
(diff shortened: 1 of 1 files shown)


cf/data/mixin/deprecations.py
```diff
@@ -731,7 +748,11 @@ def partition_boundaries(self):
 
         """
         _DEPRECATION_ERROR_METHOD(
-            "TODODASKMSG - consider using 'chunks' instead"
+            self,
+            "partition_boundaries",
+            message="Consider using the 'chunks' attribute instead.",
```
sadielbartholomew, line 753: Perhaps, to use the same tone/format as (most, codebase-wide) of the other messages, just:

```suggestion
            message="Use the 'chunks' attribute instead.",
```

?

sadielbartholomew (approved):
One minor suggestion raised, but all good so please merge when ready.

cf/data/mixin/deprecations.py
davidhassell: Hi Sadie, I see what you mean, but I try to use "consider" when the suggested alternative doesn't quite do what the deprecated method did, i.e. you can't just change the name and carry on!

sadielbartholomew: Ah, that makes sense! Noted for future. Thanks, marking as resolved.

PR #550, opened 2023-01-05
dask: Remove CachedArray class

davidhassell:
Remove the `CachedArray` class that was only used by LAMA. With dask we are not in so much control of cached files (which is good!), but if we were to veer into this territory again we almost certainly won't use the CachedArray approach.


sadielbartholomew (approved):
Perfect, please merge.

PR #551, opened 2023-01-05
dask: `Data.__init__` with no arguments

### LINKED ISSUE #378: LAMA to Dask: managing default(-argument) `Data` instance

davidhassell:
(Re-) Fixes #378. See that issue for details.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -336,12 +336,6 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         >>> d = cf.Data(tuple('fly'))
 
         """
-        if array is None and source is None:  # don't create no/empty Data
-            raise ValueError(
-                "Can't create empty data: some input data or datum must be "
-                "provided via the 'source' or 'array' parameters."
-            )
-
         if source is None and isinstance(array, self.__class__):
             source = array
 
@@ -387,6 +381,9 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         self.hardmask = hardmask
 
         if array is None:
+            # Initialise the dask array as `None` to allow basic
+            # inspection without failing
+            self._set_dask(array)
             return
 
         try:
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -381,9 +381,7 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         self.hardmask = hardmask
 
         if array is None:
-            # Initialise the dask array as `None` to allow basic
-            # inspection without failing
-            self._set_dask(array)
+            # No data has been set
             return
 
         try:
@@ -8315,13 +8313,16 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         dask.array<cf_soften_mask, shape=(4,), dtype=int64, chunksize=(4,), chunktype=numpy.ndarray>
 
         """
-        if apply_mask_hardness:
+        if apply_mask_hardness and "dask" in self._custom:
             if self.hardmask:
                 self.harden_mask()
             else:
                 self.soften_mask()
 
-        return self._custom["dask"]
+        try:
+            return self._custom["dask"]
+        except KeyError:
+            raise ValueError(f"{self.__class__.__name__} object has no data")
 
     def datum(self, *index):
         """Return an element of the data array as a standard Python
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -151,15 +151,12 @@ class DataTest(unittest.TestCase):
 
     def test_Data__init__basic(self):
         """Test basic `__init__` cases for Data."""
-        # Most __init__ parameters are covered by the various other tests,
-        # so this is mainly to check trivial cases and especially the edge
-        # case of 'default' Data i.e. if no parameters are specified.
+        # Most __init__ parameters are covered by the various other
+        # tests, so this is mainly to check trivial cases.
+        cf.Data()
         cf.Data(0, "s")
         cf.Data(array=np.arange(5))
         cf.Data(source=self.filename)
-
-        with self.assertRaises(ValueError):
-            cf.Data()
 
     def test_Data_equals(self):
         """Test the equality-testing Data method."""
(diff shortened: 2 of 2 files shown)


sadielbartholomew (commented):
The code changes themselves are fine and good, and with this PR merged *creation* of a plain `cf.Data()` works and is tested to do so, but I am concerned about side-effects of the creation of such an object. For example from seeing the list of methods available on a case via a `__dir__()` call, and trying out some, I hit trivial errors such as those illustrated below. Those are often the kinds of errors you might expect to happen, conceptually, on an array-less instance, or following from making:

> the "dask" key in the _custom dictionary is inintialised to None, thereby allowing basic inspection of the instance.

Ultimately I am not really sure whether allowing this (`Data.__init__` with no arguments) is a good idea after all, because really if we allow it we should go through all of the possible methods for `Data` and ensure they run or (probably for most) error gracefully with that case, and that could be pretty time-consuming. But with the alternative I implemented in #491 we have the complication of the docstring rewriting testing, so either way there is work to be done!

In light of that, what do you think we should do from here? Could we add some fancy logic, perhaps, to stop methods being called on a `Data` instance created with no data? Or (less preferred, in my eyes) warn users upon initialisation of data-less `Data` that many or most methods won't work properly?

### Method call examples

```python
>>> d = cf.Data()
>>> d.ndim
  File "/home/sadie/cf-python/cf/data/data.py", line 4357, in ndim
    return dx.ndim
AttributeError: 'NoneType' object has no attribute 'ndim'
>>> d.ndim()
  File "/home/sadie/cf-python/cf/data/data.py", line 4357, in ndim
    return dx.ndim
AttributeError: 'NoneType' object has no attribute 'ndim'
>>> d.size()
  File "/home/sadie/cf-python/cf/data/data.py", line 4427, in size
    size = dx.size
AttributeError: 'NoneType' object has no attribute 'size'
>>> d.max()
  File "/home/sadie/cfdm/cfdm/decorators.py", line 42, in inplace_wrapper
    self.INPLACE_ENABLED_PLACEHOLDER = self.copy()
  File "/home/sadie/cfdm/cfdm/data/data.py", line 1520, in copy
    return super().copy(array=array)
  File "/home/sadie/cfdm/cfdm/core/data/data.py", line 322, in copy
    return type(self)(source=self, copy=True, _use_array=array)
  File "/home/sadie/cf-python/cf/data/data.py", line 361, in __init__
    self._set_dask(array, copy=copy, conform=False)
  File "/home/sadie/cf-python/cf/data/data.py", line 1282, in _set_dask
    array = array.copy()
AttributeError: 'NoneType' object has no attribute 'copy'
>>> d.get_filenames()
  File "/home/sadie/cf-python/cf/data/data.py", line 5474, in get_filenames
    dsk = collections_to_dsk((self.to_dask_array(),), optimize_graph=True)
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/dask/base.py", line 368, in collections_to_dsk
    dsk, keys = _extract_graph_and_keys(val)
  File "/home/sadie/anaconda3/envs/cf-env/lib/python3.8/site-packages/dask/base.py", line 394, in _extract_graph_and_keys
    graphs.append(v.__dask_graph__())
AttributeError: 'NoneType' object has no attribute '__dask_graph__'
>>> d.is_masked()
  File "/home/sadie/cf-python/cf/data/data.py", line 4281, in is_masked
    out_ind = tuple(range(dx.ndim))
AttributeError: 'NoneType' object has no attribute 'ndim'

```

davidhassell:
Hi Sadie, OK, how about https://github.com/NCAS-CMS/cf-python/pull/551/commits/59a6e3dddff435c18422e67fa89a77d6c8383fc0, which produces:

```python
>>> import cf
>>> d = cf.Data()
>>> d.ndim
    ...
ValueError: Data object has no data
>>> d.get_filenames()
    ...
ValueError: Data object has no data
>>> 
```

sadielbartholomew:
> OK, how about https://github.com/NCAS-CMS/cf-python/commit/59a6e3dddff435c18422e67fa89a77d6c8383fc0

Ah, perfect! Achieves what I hoped for as best solution, namely:

> go through all of the possible methods for Data and ensure they run or (probably for most) error gracefully with that case

without any of the nuisance I anticipated! Great idea.

I'm just re-running the test suite and then I can approve. Thanks.

sadielbartholomew (approved):
All good now (feedback addressed). Ideally we can add one or two `self.assertRaises(ValueError)` checks to the test to ensure the behaviour implemented by 59a6e3dddff435c18422e67fa89a77d6c8383fc0 is coming through to the methods (which I have tested manually), but regardless please merge when ready.

davidhassell:
Thanks, @sadielbartholomew  - I've add a test (https://github.com/NCAS-CMS/cf-python/pull/551/commits/8e343bba8bb4a5b2e71f6e94308b89c725f0fda1) ... and merging.

PR #552, opened 2023-01-05
dask: `Container` inheritance

davidhassell:
* Add missing `Container` mixin class inheritance to array classes. This is needed to pick up the cf docstring substitutions.
* Allow no-arg init of `CFANetCDFArray`

Necessary (but not sufficient) to get `test_docstring.py` to pass.

### DIFF
--- a/cf/data/array/abstract/array.py
+++ b/cf/data/array/abstract/array.py
@@ -1,9 +1,10 @@
 import cfdm
 
+from ....mixin_container import Container
 from ..mixin import ArrayMixin
 
 
-class Array(ArrayMixin, cfdm.Array):
+class Array(ArrayMixin, Container, cfdm.Array):
     """Abstract base class for a container of an underlying array.
 
     The form of the array is defined by the initialization parameters
--- a/cf/data/array/cfanetcdfarray.py
+++ b/cf/data/array/cfanetcdfarray.py
@@ -134,7 +134,7 @@ class CFANetCDFArray(NetCDFArray):
                 aggregated_data = source.get_aggregated_data(copy=False)
             except AttributeError:
                 aggregated_data = {}
-        else:
+        elif filename is not None:
             from CFAPython import CFAFileFormat
             from CFAPython.CFADataset import CFADataset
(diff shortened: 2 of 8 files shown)


sadielbartholomew (changes requested):
The first point (from your opening comment) has been achieved, but the second point promise of:

> Allow no-arg init of CFANetCDFArray

seems to require some tweaks since I see:

```pytb
>>> import cf
>>> cf.CFANetCDFArray()
  File "/home/sadie/cfdm/cfdm/core/abstract/container.py", line 259, in _get_component
    return self._components[component]
KeyError: 'shape'

During handling of the above exception, another exception occurred:

  File "/home/sadie/cf-python/cf/data/array/netcdfarray.py", line 20, in __repr__
  File "/home/sadie/cfdm/cfdm/data/netcdfarray.py", line 295, in __repr__
    return f"<{self.__class__.__name__}{self.shape}: {self}>"
  File "/home/sadie/cfdm/cfdm/data/netcdfarray.py", line 409, in shape
    return self._get_component("shape")
  File "/home/sadie/cfdm/cfdm/core/abstract/container.py", line 264, in _get_component
    return self._default(
  File "/home/sadie/cfdm/cfdm/core/abstract/container.py", line 141, in _default
    raise default
ValueError: CFANetCDFArray has no 'shape' component
```

Otherwise everything looks good!

davidhassell:
Hi Sadie - I'm happy to let this go, as the no-arg init is working fine ... it's just the `repr` of it that fails, and that's OK :)

davidhassell:
Hi Sadie - couldn't stop thinking about this. You're right that `repr` should _also_ work, but as this class is a placeholder for 4.0.0 and is not used at all by 3.14.0, it's probably best to not develop if further until after 3.14.0, as we'll probably only change what we might do now. Thanks.

sadielbartholomew (approved):
Thanks for the clarification regarding the `repr`. I'm happy given your justification that it doesn't work right now.

Feedback/questions all addressed. Great, please merge.

PR #554, opened 2023-01-06
dask: `cf.size`

davidhassell:
Replace `np.size(...)` with the new `cf.size(...)`, which is a basically a copy of the numpy function.

The reason for doing this is that when dask arrays are passed to `np.size` this happens:

```python
>>> import numpy as np
>>> import dask.array as da
>>> np.size(da.arange(9))
/home/david/miniconda3/lib/python3.9/site-packages/dask/array/core.py:1712:
FutureWarning: The `numpy.size` function is not implemented by Dask array.
You may want to use the da.map_blocks function or something similar to
silence this warning. Your code may stop working in a future release.
  warnings.warn(
9
>>>
```

This is related to how Dask interacts with the numpy `__array_function__` interface: https://github.com/dask/dask/blob/2022.12.1/dask/array/core.py#L1707-L1728. I haven't really looked at what is going on here, but what look I had led me to think that creating our own little function was the path of least resistance :). If this were to be something which could be "fixed" via Dask functionality, then great! But until such a time we are covered.

Took the liberty of deleting some ancient redundant code from `field.py` at the same time.


cf/functions.py
```diff
@@ -3096,6 +3096,43 @@ def default_netCDF_fillvals():
...
+
+    **Examples**
+
+    >>> cf.size(9)
+    1
+    >>> cf.size("foo")
+    1
+    >>> cf.size([9])
+    1
+    >>> cf.size((8, 9)
```
sadielbartholomew, line 3120:
```suggestion
    >>> cf.size((8, 9))
```

cf/functions.py
```diff
@@ -3096,6 +3096,43 @@ def default_netCDF_fillvals():
     return netCDF4.default_fillvals
 
 
+def size(a):
+    """Return the number of elements.
```
sadielbartholomew, line 3100: +1 for documenting this by hand and not going with the blanket "This docstring was copied from numpy.size. Some inconsistencies with the cf version may exist." :smiley:

sadielbartholomew (approved):
Thanks David for the useful context giving the motivations for this. I agree that:

> creating our own little function was the path of least resistance :)

is the sensible way to handle this.

Perfect, please merge once one typo has been fixed.

PR #557, opened 2023-01-09
dask: `Field.argmax`

davidhassell:


sadielbartholomew:
I haven't been assigned to this but assume you want me to review it (after the regridding PR) @davidhassell, so will self-assign. Unassign me if I'm wrong. Thanks.

cf/field.py
```diff
@@ -12458,104 +12458,94 @@ def flip(self, axes=None, inplace=False, i=False, **kwargs):
...
+        >>> i = f.argmax('T')
+        >>> print(i.array)
+        [[31 10  2 27 16  7 21 17]
+         [24 34  1 22 28 17 19 13]
+         [20  7 15 21  6 20  8 18]
+         [24  1  7 18 19  6 11 18]
+         [16  8 12  9 12  2  6 17]]
+
+        Find the coordinates of the global maximum value, showing that
+        it occurs on 1960-11-16 at location 292.5 degrees_east, 45.0
```
sadielbartholomew, line 12527:
```suggestion
        it occurs on 1960-11-16 at location 292.5 degrees east, 45.0
```

(or add an underscore within the line below, to achieve consistency).

cf/field.py
```diff
@@ -12458,104 +12458,94 @@ def flip(self, axes=None, inplace=False, i=False, **kwargs):
...
+            unravel: `bool`, optional
+                If True then when locating the maximum over the whole
+                data, return the location as an index for each axis as
+                a `tuple`. By default an index to the flattened array
+                is returned in this case. Ignored if locating the
+                maxima over a subset of the axes.
+
         :Returns:
 
+            `Data` or `tuple`
```
sadielbartholomew, line 12500:
```suggestion
            `Data` or `tuple` of `Data`
```

to be specific?

sadielbartholomew (approved):
I've raised a few minor comments but overall this is all good. Please merge when ready.

One quick question, also: why do we have an `argmax` but not `argmin` method on both Data and Field? Seems weird to me to have one and not both...

cf/field.py
davidhassell: It's OK, but it should by tuple of `int`: https://github.com/NCAS-CMS/cf-python/pull/557/commits/f8e3b407ff4e60c4e688dccef47ba848a28522a2

davidhassell:
Hi Sadie - there's no `argmin` simply because there wasn't one before, and I didn't want to create extra work right now. We'll put it in for 4.0.0 (I'll open an issue so we don't forget).

PR #563, opened 2023-01-12
dask: Catch `NotImplemented` resulting from arithmetic

davidhassell:
Prevent this:
```python
>>>  import cf
>>> d = cf.Data([1, 2, 3])
>>>  e = d + ('sdf',)
NotImplemented has been set in the place of a dask array.

This could occur if any sort of exception is raised by a function that is run on chunks (via, for instance, da.map_blocks or
dask.array.core.elemwise). Such a function could get run at definition time in order to ascertain suitability (such as
data type casting, broadcasting, etc.). Note that the exception may be difficult to diagnose, as dask will have silently
trapped it and returned NotImplemented (for instance, see dask.array.core.elemwise). Print statements in a local copy
of dask are possibly the way to go if the cause of the error is not obvious.

>>>
```
replacing it with raising a `TypeError`, which is what numpy and Dask do.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -1290,7 +1290,8 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         """
         if array is NotImplemented:
             logger.warning(
-                "NotImplemented has been set in the place of a dask array."
+                "WARNING: NotImplemented has been set in the place of a "
+                "dask array."
                 "\n\n"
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
> replacing it with raising a TypeError, which is what numpy and Dask do

I agree that's the best approach. Perfect, please merge.

PR #564, opened 2023-01-12
dask: Allow weighted sums in `Data.stats`

davidhassell:
Allow weighted sums in `Data.stats`.

### DIFF
--- a/cf/test/test_Data.py
+++ b/cf/test/test_Data.py
@@ -857,9 +857,7 @@ class DataTest(unittest.TestCase):
 
         # NaN values aren't 'equal' to e/o, so check call works and that some
         # representative values are as expected, in this case
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=RuntimeWarning)
-            s5 = cf.Data([[-2, -1, 0], [1, 2, 3]]).stats(all=True, weights=0)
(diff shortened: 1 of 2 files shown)


cf/test/test_Data.py
```diff
@@ -854,12 +854,16 @@ def test_Data_stats(self):
                 "sample_size": 1,
             },
         )
+
         # NaN values aren't 'equal' to e/o, so check call works and that some
         # representative values are as expected, in this case
-        s5 = cf.Data([[-2, -1, 0], [1, 2, 3]]).stats(all=True, weights=0)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            s5 = cf.Data([[-2, -1, 0], [1, 2, 3]]).stats(all=True, weights=0)
```
sadielbartholomew, line 862: These `RuntimeWarning`s are already caught via the `setUp` method so we can simplify to:

```suggestion
        s5 = cf.Data([[-2, -1, 0], [1, 2, 3]]).stats(all=True, weights=0)
```

(I double-checked that it doesn't raise the warnings still locally after this change.)

sadielbartholomew (approved):
One minor change suggested, otherwise good to merge. Thanks.

cf/test/test_Data.py
davidhassell: Great - that's much nicer :)

PR #572, opened 2023-01-27
LAMA to Dask: misc. fixes for a passing `test_aggregate`

sadielbartholomew:
Towards #182, confirms that `cf.aggregate` works under the new "Dask way" with no modifications, with `test_basic_aggregate` failing instead due to specific issues elsewhere, namely:

* an issue with data comparison by `Data.equals`;
* an issue with construct comparison leading to field copies not being recognised as equal.

(Note to tidy up previous investigation/debugging branches, I created a fresh branch and re-applied commits to break down the fixes to specific parts.)

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -7063,10 +7063,19 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
                 atol=float(atol),
             )
         elif not self_is_numeric and not other_is_numeric:
-            # Apply bool() since the output may be numeric and we need a
-            # Boolean result to avoid casting error ("... can't be cast using
-            # casting='same_kind'")
-            data_comparison = bool(da.all(self_dx == other_dx))
+            # If the array (say d) is fully masked, then the output of
+            # np.all(d == d) and therefore da.all(d == d) will be a
+            # np.ma.masked object which has dtype('float64'), and not
+            # a Boolean, causing issues later. To ensure data_comparison
+            # is Boolean, we must do an early compute to check if it is
+            # a masked object and if so, force the desired result (True).
+            #
+            # This early compute won't degrade performance because it
+            # would be performed towards result.compute() below anyway.
+            data_comparison = da.all(self_dx == other_dx).compute()
+            if data_comparison is np.ma.masked:
+                data_comparison = True
+
         else:  # one is numeric and other isn't => not equal (incompat. dtype)
             logger.info(
                 f"{self.__class__.__name__}: Different data types:"
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -7063,7 +7063,10 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
                 atol=float(atol),
             )
         elif not self_is_numeric and not other_is_numeric:
-            data_comparison = da.all(self_dx == other_dx)
+            # Apply bool() since the output may be numeric and we need a
+            # Boolean result to avoid casting error ("... can't be cast using
+            # casting='same_kind'")
+            data_comparison = bool(da.all(self_dx == other_dx))
         else:  # one is numeric and other isn't => not equal (incompat. dtype)
             logger.info(
                 f"{self.__class__.__name__}: Different data types:"
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -146,7 +146,33 @@ class _Meta:
 
             f: `Field` or `Domain`
 
-            {{verbose: `int` or `str` or `None`, optional}}
+            verbose: `int` or `str` or `None`, optional
+                If an integer from ``-1`` to ``3``, or an equivalent
+                string equal ignoring case to one of:
+
+                * ``'DISABLE'`` (``0``)
+                * ``'WARNING'`` (``1``)
+                * ``'INFO'`` (``2``)
+                * ``'DETAIL'`` (``3``)
+                * ``'DEBUG'`` (``-1``)
+
+                set for the duration of the method call only as the
+                minimum cut-off for the verboseness level of displayed
+                output (log) messages, regardless of the
+                globally-configured `cf.log_level`. Note that
+                increasing numerical value corresponds to increasing
+                verbosity, with the exception of ``-1`` as a special
+                case of maximal and extreme verbosity.
+
+                Otherwise, if `None` (the default value), output
+                messages will be shown according to the value of the
+                `cf.log_level` setting.
+
+                Overall, the higher a non-negative integer or
+                equivalent string that is set (up to a maximum of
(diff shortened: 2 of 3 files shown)


sadielbartholomew:
@davidhassell please either hold on the reviewing here until tomorrow morning, or you can partially review under the following context:

* as-is on this branch, `test_aggregate` passes fully, but notably I've commented out a set of related test assertions, which rely on a fix in cfdm for which I have a PR but it is not quite ready so I will put up tomorrow morning;
* once the corresponding cfdm PR is in, I will self-review that everything's working together and then open this for review fully.

If you like you could review the commits here standalone noting the above, with the main one is the initial tweak to fix an aspect of `Data.equals`, [02133f7](https://github.com/NCAS-CMS/cf-python/pull/572/commits/02133f71e987e6dfb5d7efaf7c110c83a9f1e5f8).

cf/data/data.py
```diff
@@ -7063,7 +7063,10 @@ def equals(
                 atol=float(atol),
             )
         elif not self_is_numeric and not other_is_numeric:
-            data_comparison = da.all(self_dx == other_dx)
+            # Apply bool() since the output may be numeric and we need a
+            # Boolean result to avoid casting error ("... can't be cast using
+            # casting='same_kind'")
+            data_comparison = bool(da.all(self_dx == other_dx))
```
sadielbartholomew, line 7069: I'm not convinced this change is the *best* approach to solving the issue of some of these `da.all` calls emerging as non-Boolean types, rather than tweaking something downstream, but this works and is elegant so I think it is good at least for now?

davidhassell: Hi Sadie - in what circumstances is the result of `da.all` of non-boolean datatype?

sadielbartholomew: Hi David, good question! It seems when there are string types under comparison it returns a numeric somehow, as the case in question here. Let me just try to get a minimal example in IPy to show you, one moment...

sadielbartholomew: OK here we are. I am wondering if this is a bug in Dask now I've got the example here (for which my `bool()` wrapper will provide a workaround for now for our release, if so). IN short, for an all-masked array (perhaps only `string` data type), the `compute`d result is of type `float64` and not Boolean. I am not sure if I'm missing something obvious, but I have no idea why that would or should be the case... Have a look at this and please let me know what you think @davidhassell.

```python
In [34]: s = np.array([b'foo', b'bar', b'baz'])

In [35]: sm = np.ma.array([b'foo', b'bar', b'baz'], mask=[True, False, True])

In [36]: s_da = da.from_array(s)

In [37]: s_da
Out[37]: dask.array<array, shape=(3,), dtype=|S3, chunksize=(3,), chunktype=numpy.ndarray>

In [38]: sm_da = da.from_array(sm)

In [39]: sm_da
Out[39]: dask.array<array, shape=(3,), dtype=|S3, chunksize=(3,), chunktype=numpy.MaskedArray>

In [40]: # This all works as expected...

In [41]: s_da.dtype
Out[41]: dtype('S3')

In [42]: sm_da.dtype
Out[42]: dtype('S3')

In [43]: da.all(s_da == s_da).dtype
Out[43]: dtype('bool')

In [44]: da.all(s_da == s_da).compute().dtype
Out[44]: dtype('bool')

In [45]: da.all(sm_da == sm_da).dtype
Out[45]: dtype('bool')

In [46]: da.all(sm_da == sm_da).compute().dtype
Out[46]: dtype('bool')

In [47]: sm = np.ma.array([b'foo', b'bar', b'baz'], mask=[True, True, True])

In [48]: # This is an all masked array, unlike the previous: note the difference

In [49]: sm_da.dtype
Out[49]: dtype('S3')

In [50]: sm2 = np.ma.array([b'foo', b'bar', b'baz'], mask=[True, True, True])

In [51]: sm2_da = da.from_array(sm2)

In [52]: sm2_da.dtype
Out[52]: dtype('S3')

In [53]: da.all(sm2_da == sm2_da).dtype
Out[53]: dtype('bool')

In [54]: da.all(sm2_da == sm2_da).compute().dtype
Out[54]: dtype('float64')

In [55]: # What?

```

sadielbartholomew: I'm on:

```console
dask                      2022.5.0+12.gafe1e06a           dev_0    <develop>
dask-core                 2023.1.0           pyhd8ed1ab_0    conda-forge
```

FYI. SO I should be up to date with regards to the dask version? I'm just having a look at their Issue Tracker to see if there's anything like this registered there...

davidhassell: Weird, indeed - I'll absorb those results ....

davidhassell:
```python
In [14]: np.all(sm2 == sm2)
Out[14]: masked

In [15]: np.all(sm2 == sm2).dtype
Out[15]: dtype('float64')

In [16]: bool(np.all(sm2 == sm2))
Out[16]: False
```
So, the results is `masked`, which is indeed a float, but which is `False` when evaluated in a boolean context.

However, we want it to be `True`! Do we need something like:
```python
        data_comparison = da.all(self_dx == other_dx).compute()
        if data_comparison is np.ma.masked:
            data_comparison = True
```

sadielbartholomew: Aha, the `float64` in question being `masked` makes a lot of sense! And indeed, good point that we want a `True` result instead.

I think your suggestion works better, and is good enough for now (v3.14), however (for an ultimate solution) I feel like fundamentally we shouldn't really be getting a masked point here, since the idea is to compare the underlying data with the `data_comparison` and leave any aspects of the masking for the `mask_comparison` later on, since the whole point is to separate those out in our `Data.equals` approach. Unless I am missing something?

sadielbartholomew:
Thanks for all of your invaluable guidance here @davidhassell. (I manually added you as a co-author on the new fix approach commit, since you essentially wrote that code!)

This is good to review now, I think. FYI, the linting CI job is failing due to steup issues relating to GitLab (I see these on the command line too, so think it will resolve itself in time and will ignore it for now). So we can ignore that.

cf/data/data.py
```diff
@@ -7063,7 +7063,19 @@ def equals(
...
-            data_comparison = da.all(self_dx == other_dx)
+            # If the array (say d) is fully masked, then the output of
+            # np.all(d == d) and therefore da.all(d == d) will be a
+            # np.ma.masked object which has dtype('float64'), and not
+            # a Boolean, causing issues later. To ensure data_comparison
+            # is Boolean, we must do an early compute to check if it is
+            # a masked object and if so, force the desired result (True).
+            #
+            # This early compute won't degrade performance because it
+            # would be performed towards result.compute() below anyway.
```
davidhassell, line 7074: Great comment!

davidhassell (approved):
Looks good. All makes sense following our conversations.

PR #583, opened 2023-01-30
dask: The final removal of LAMA :( :)

davidhassell:
The final removal of LAMA :( :)


davidhassell:
LAMA is dead. Long live Dask.

PR #584, opened 2023-02-02
Added rotated to regular grid recipe

bewithankit:
Added recipe to regrid from a rotated latitude-longitude source domain to a regular latitude-longitude destination domain.


bewithankit:
Will do the PR on Main.

sadielbartholomew:
Thanks @bewithankit. (Closed because I am in the process of changing the default 'NCAS-CMS/cf-python' branch to 'main' instead of 'master' as present, and this was the lone PR raised against 'master', and to be re-opened on 'main'.)

PR #585, opened 2023-02-02
Default branch rename from `master` to `main`

### LINKED ISSUE #190: Rename master branch to 'main'
GitHub has made the switch from calling the default branch of a repository 'main' rather than 'master' (given the slavery connotations as far as I have read) so to keep up with the times we should eventually rename our `master` branches on cf-python, cfdm and cfunits accordingly.

I think a sensible time to make the change would be right after we merge in the `dask` branch ready for a 4.0.0 release of cf-python. We can update our development workflows (including GitHub Actions workflows) and local branches and aliases etc., as well as the developer documentation, then. I don't think it will take very long to sort, but I think best wait until the Dask work is in to make the name change across our repos.

sadielbartholomew:
Closes #190. Trivial configuration and setup changes, so doesn't need a review.


sadielbartholomew:
This PR includes an update to the CI workflows to reference `main` and not `master`, where the present references pre-merge will not be found because `master` has just been deleted. So the CI jobs should fail here, but hopefully work once we merge.

Hence good to merge.

PR #600, opened 2023-02-23
New `remove_vertical_crs` keyword to `cf.Field.collapse`

### LINKED ISSUE #599: `cf.Field.collapse` should delete a vertical CRS that spans a collapse axis
When `cf.Field.collapse` operates along one or more axes spanned by a vertical coordinate reference system (such as `standard_name:atmosphere_hybrid_height_coordinate`), the code currently deletes any domain ancillaries that span the any of the collapse axes, but leaves other domain ancillaries behind, as well as the vertical coordinate reference system.

However, the [CF conventions say](https://cfconventions.org/cf-conventions/cf-conventions.html#parametric-v-coord) that missing terms should be assumed to be zero, and so if the non-parametric vertical coordinates are computed they will likely give inappropriate, but possibly plausible, results - the worst kind of error!

So, removing the CRS and its contained constructs is certainly the safe thing to do by default, but you might want to keep it, perhaps to reinstate a new correct domain ancillary. Therefore, it should be possible to override the removal of the vertical CRS with a keyword parameter to `cf.Field.collapse`. 

PR to follow.

```python
>>> import cf
Platform: Linux-5.15.0-60-generic-x86_64-with-glibc2.35
HDF5 library: 1.12.1
netcdf library: 4.8.1
ESMF: 8.2.0
Python: 3.10.8
dask: 2022.12.1
netCDF4: 1.6.0
psutil: 5.9.0
packaging: 22.0
numpy: 1.23.5
scipy: 1.9.3
matplotlib: 3.6.2
cftime: 1.6.2
cfunits: 3.3.5
cfplot: 3.1.31
cfdm: 1.10.0.2
cf: 3.14.0
```

davidhassell:
Fixes #599

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,14 @@
+version 3.14.1
+--------------
+
+**2023-??-??**
+
+* Fixed bug in `cf.Field.collapse` that always retained invalid
+  vertical coordinate reference systems
+  (https://github.com/NCAS-CMS/cf-python/issues/599)
+
+----
+
 version 3.14.0
 --------------
 
--- a/cf/field.py
+++ b/cf/field.py
@@ -46,6 +46,7 @@ from .functions import (
     _DEPRECATION_ERROR_METHOD,
     DeprecationError,
     _section,
+    flat,
     parse_indices,
 )
 from .functions import relaxed_identities as cf_relaxed_identities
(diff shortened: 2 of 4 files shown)


sadielbartholomew (approved):
Fixes the issue at hand with a nice implementation and adds sufficient testing to confirm that it does. I've raised one minor comment, but otherwise this is all good.

PR #604, opened 2023-02-28
Added Sphinx-Gallery

### LINKED ISSUE #590: Documentation: add gallery to itemise & showcase recipe plots
I am loving our new [recipes section of the docs](https://ncas-cms.github.io/cf-python/recipes.html), especially the lovely cf-plot outputs that are exhibited as outputs of many/most of the recipes. :100: This made me think that it would be great to include a HTML gallery (or more than one, depending on how many sections seem sensible) which shows the main plot produced alongside the name of the recipe at the top of the recipes page, either in place of or below the present bullet-pointed list itemising the recipes.

It might not be clear from that description quite what I mean, so specifically I am talking about the HTML galleries used commonly to display a little table of 'cards' holding output examples in the docs of libraries capable of producing or leading to image outputs of some sort, with good examples as follows (looking at the linked pages should make it clear what I mean and envision, no need to screenshot them here too):

* [matplotlib](https://matplotlib.org/stable/plot_types/index.html);
* [cartopy](https://scitools.org.uk/cartopy/docs/latest/gallery/index.html#);
* [astropy](https://docs.astropy.org/en/stable/generated/examples/index.html);
* etc. (see link below for more examples if desired).

Most of these libraries use a neat library called ['Sphinx-Gallery'](https://sphinx-gallery.github.io/stable/index.html) for the implementation, and indeed there is a nice listing of some notable projects using that tool, with even more examples linked, [here](https://sphinx-gallery.github.io/stable/projects_list.html).

This way, we can showcase the nice plots users can produce, helping to advertise and summarise the recipes. (And it looks cool :sunglasses: .) I imagine there will be at least some cases where recipes won't produce a plot for (one of) the end-result(s), in which case we can use a placeholder image to indicate that there is no associated plot or image. This can be customised under 'Sphinx-Gallery', e.g. [SunPy use a specific image of the sun which looks nice](https://docs.sunpy.org/en/stable/generated/gallery/index.html). Ours could be an image of the Earth or similar.

Note many examples have a hover-over feature which provides extra information on hover, which would be useful to give a one-line summary of the recipe.

***

This is a challenge for you to take on, should you wish to accept it (and have the time too of course!), @bewithankit.

bewithankit:
Closes #590


davidhassell (approved):
Looks good!

PR #606, opened 2023-03-01
Allow `cf.aggregate` to retain external cell measures

### LINKED ISSUE #605: Allow `cf.aggregate` to retain external cell measures
External cell cell measures are always deleted during aggregation. The [code says](https://github.com/NCAS-CMS/cf-python/blob/v3.14.0/cf/aggregate.py#L666-L669): 
```python
            # If the measure is an external variable, remove it because
            # the dimensions are not known so there is no way to tell if the
            # aggregation should have changed it. (This is sufficiently
            # sensible behaviour for now, but will be reviewed in future.)
```
Recent experiences suggest that a different approach is better: keep common external cell measures, and let the checking done by `cf.read` highlight any problems at a later date.

This allows the the cell measures to be retained when they are still valid, and if they are broken trap that before any harm is done.

```python
>>> import cf
Platform: Linux-5.15.0-60-generic-x86_64-with-glibc2.35
HDF5 library: 1.12.2
netcdf library: 4.8.1
ESMF: 8.2.0
Python: 3.10.8
dask: 2023.1.1
netCDF4: 1.6.2
psutil: 5.9.4
packaging: 23.0
numpy: 1.22.3
scipy: 1.8.1
matplotlib: 3.4.3
cftime: 1.6.2
cfunits: 3.3.5
cfplot: 3.1.31
cfdm: 1.10.0.2
cf: 3.14.0
```

davidhassell:
Fixes #605


davidhassell:
Hi @sadielbartholomew  - I think I have to go ahead and merge this for timely "CANARI" purposes. Happy to revisit later ...

PR #607, opened 2023-03-02
Refactor the modification of components when setting a new Dask array with `Data._set_dask`

davidhassell:
There was nothing particularly wrong with the original code, but two upcoming PRs on active storage and CFA are going to need a much much fine-grained approach to conforming components when the Dask array is updated.

That is why the new `_ALL` constant is 15 and not 3 - the aforementioned PRs each need a new constants, with value `4` and `8`.


cf/data/data.py
```diff
@@ -1239,26 +1248,64 @@ def __keepdims_indexing__(self):
...
+        if clear & _ARRAY:
+            # Delete a source array
+            self._del_Array(None)
 
-    def _set_dask(self, array, copy=False, conform=True):
+        if clear & _CACHE:
+            # Delete cached element values
+            self._del_cached_elements()
+
+    def _set_dask(self, array, copy=False, clear=_ALL):
```
sadielbartholomew, line 1308: The 'seealso' admonition for this method still references the old `_conform_after_dask_update` and needs updating (but a `grep` on `cf/` on this branch shows that that one is the last remaining reference to the old name).

sadielbartholomew (commented):
A clever way to achieve finer granularity for setting the components, nice! The implementation looks good, though I'd prefer to see some more testing to reflect the increased flexibility, so could we add a little to the unit tests? Other than that, I've raised some minor comments but everything looks good.

cf/data/data.py
davidhassell: Fixed: https://github.com/NCAS-CMS/cf-python/pull/607/commits/64b028d94ec591a71c90c4efa16b9bfbb570ac55

davidhassell:
I shall concoct some more tests ...

davidhassell:
New test: https://github.com/NCAS-CMS/cf-python/pull/607/commits/7b11017e15d58c0d7fd2a93f0845faeaa47aeb86

PR #608, opened 2023-03-05
Refactor the storage and access of cached elements

davidhassell:
There was nothing particularly wrong with the original code, but the upcoming PR on CFA needs more performant and flexible approach.

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -1397,7 +1397,9 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         :Returns:
 
             `dict`
-                The cached element values
+                The cached element values, where the keys are the element
+                positions within the dask array and the values are the cached
+                values for each position.
 
         **Examples**
 
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -1374,31 +1374,45 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
     def _del_cached_elements(self):
         """Delete any cached element values.
 
-        Updates *data* in-place to remove the cached element values
-        ``'first_element'``, ``'second_element'`` and
-        ``'last_element'``.
-
-        .. note:: By default, `_del_cached_elements` is run whenever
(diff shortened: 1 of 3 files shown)


cf/data/data.py
```diff
@@ -1374,41 +1374,58 @@ def _del_dask(self, default=ValueError(), conform=True):
...
+            `dict`
+                The cached element values
+
+        **Examples**
+
+        >>> d._get_cached_elements()
+        {}
+
+        >>> d._get_cached_elements()
+        {0: 273.15, 1: 274.56, -1: 269.95}
```
sadielbartholomew, line 1408: Can users specify to cache values that aren't the first, second or last somehow (or will we ever open up that possibility)? Because then the negative indices could lead to trouble, for example say the array is size `n` then the equivalent index `n-1` would be stored separately in the dictionary and the two could get out of sync. So possibly we'd need some extra logic to handle that, in which case some updates along those lines might be required?

sadielbartholomew (approved):
The change from `'nth_element'` to index value representing the element's array position is very sensible (and natural, not sure why neither of us picked up on that as a better way in the original implementation!).

I've raised a few questions and minor comments but otherwise everything is good to go, so please consider those and then merge when ready. Thanks.

cf/data/data.py
davidhassell: They can indeed, but I'm not worried about them getting out of sync because we already do this for size 1 arrays (i.e. we store `0` and `-1` separately with the same value; and also all of this is "protected" by underscores, so buyer beware! How does that sound?

sadielbartholomew: Aha, good point (and good illustration RE the size-1 array)! Internals should not be messed with and if the user does this it's their own fault if something goes wrong, I can't argue with that :smile: That's fine by me, then. Thanks for clarifying. Marking as resolved.

sadielbartholomew (approved):
(Confirming that all feedback has been adequately addressed, and this is ready to merge in my eyes.)

PR #622, opened 2023-03-22
Fixes bug when reading UM/PP data that are runlength encoded 

### LINKED ISSUE #621: Reading UM/PP files fails when LBNPT or LBROW is zero
When UM/PP either of the lookup header entries  LBNPT or LBROW are zero, a failure occurs when trying to create the coordinates:
```
>>> import cf
>>> f = cf.read('cm655_cn134_19531029-W_uk-11a.da19531026_00')
<snip>
 File ~/miniconda3/lib/python3.10/site-packages/cf/read_write/um/umread.py:3031, in UMField.xy_coordinate(self, axiscode, axis)
   3028 dc = self.coord_names(dc, axiscode)
   3030 if X and bounds is not None:
-> 3031     autocyclic["cyclic"] = abs(bounds[0, 0] - bounds[-1, -1]) == 360.0
   3032     autocyclic["period"] = Data(360.0, units=units)
   3033     autocyclic["axis"] = axis_key

IndexError: index 0 is out of bounds for axis 0 with size 0
```
Such fields are runlength encoded, and cf-python can't unpack them because it doesn't know what the mask is.

A solution is to simply skip these fields.

```python
>>> import cf
HDF5 library: 1.12.1
netcdf library: 4.8.1
ESMF: 8.2.0
Python: 3.10.9
dask: 2023.3.0
netCDF4: 1.6.0
psutil: 5.9.4
packaging: 23.0
numpy: 1.22.3
scipy: 1.8.1
matplotlib: 3.4.3
cftime: 1.6.2
cfunits: 3.3.5
cfplot: 3.1.31
cfdm: 1.10.0.3
cf: 3.14.1
```

davidhassell:
Fixes #621

### DIFF
--- a/cf/read_write/um/umread.py
+++ b/cf/read_write/um/umread.py
@@ -664,8 +664,9 @@ class UMField:
 
         if not LBROW or not LBNPT:
             logger.warn(
-                f"WARNING: Skipping field derived from STASH code {stash} "
-                f"with LBROW={LBROW} and LBNPT={LBNPT}"
+                f"WARNING: Skipping STASH code {stash} with LBROW={LBROW}, "
+                f"LBNPT={LBNPT}, LBPACK={int_hdr[lbpack]} "
+                "(possibly runlength encoded)"
             )  # pragma: no cover
             self.field = (None,)
             return


cf/read_write/um/umread.py
```diff
@@ -662,6 +662,14 @@ def __init__(
         BDX = real_hdr[bdx]
         BDY = real_hdr[bdy]
 
+        if not LBROW or not LBNPT:
+            logger.warn(
+                f"WARNING: Skipping field derived from STASH code {stash} "
+                f"with LBROW={LBROW} and LBNPT={LBNPT}"
```
sadielbartholomew, line 668: Could we add a brief sentence to justify _why_ the field(s) was (were) skipped to put the line into context? Or should that be obvious to users of UM and PP data (those who know more about the lookup codes than myself - all I can find from a little look-see is some descriptions of those pair of entries in the UM docs e.g. [here](https://artefacts.ceda.ac.uk/badc_datadocs/um/umdp_F3-UMDPF3.pdf) to add to your comment on this Issue:

> Such fields are runlength encoded, and cf-python can't unpack them because it doesn't know what the mask is.

as background)?

sadielbartholomew (commented):
> A solution is to simply skip these fields.

I have to ask, since I lack contextual knowledge about the PP lookup codes and context, and can't seem to find much from looking briefly online to help (see also my in-line comment), but are you taking this approach, namely skipping the field(s) in question, because they simply can't be read in (e.g. from lack of sufficient metadata on the field) and therefore it is only sensible to skip them, or because it is convenient to skip them i.e. too much hassle to process them to get them read in?

It's naturally fine if it's the former, but if the latter it would be good to just get a brief justification why such fields / zero value lookup entries are troublesome (mostly to help my own understanding!). Thanks.

Apart from my questions raised above and in-line, this all looks good and I'll approve after I know more on the background. Thanks.

cf/read_write/um/umread.py
sadielbartholomew: Fair enough. That all makes sense and thanks for clarifying.

In that case, it sounds like indicating it is potentially run-length encoded is indeed a good way forward, so please go ahead with that suggestion or a similar one.

sadielbartholomew: (Discussed offline. In short, all good and new commit implements, so marking as resolved.)

sadielbartholomew (approved):
All questions answered / feedback addressed. Great, please merge.

PR #624, opened 2023-03-23
Fix excessive regrid memory use

### LINKED ISSUE #623: Regridding sometimes fails due to running out of memory
Since the advent of Dask at v3.14.0, regridding sometimes fails due to running out of memory.

This happens because the sparse weight matrix is converted to dense form, which can be huge. E.g. when regridding a 400x400 grid to a 300x300 grid, the dense  weights matrix has shape `(90000, 160000)`, taking up `300*300*400*400*8 bytes = 107.3 GiB`. That is enough to kill any process on my laptop, at least!

The attached PR reformulated the weights creation so that it stays in sparse form. In the above example, the sparse form has size (assuming 4 destination grid weights per source grid point, as is the case for, e.g., linear regridding) `300*300*4*8 bytes = 0.0027 GiB` - essentially negligible!

```python
>>> import cf
HDF5 library: 1.12.1
netcdf library: 4.8.1
ESMF: 8.2.0
Python: 3.10.9
dask: 2023.3.0
netCDF4: 1.6.0
psutil: 5.9.4
packaging: 23.0
numpy: 1.22.3
scipy: 1.8.1
matplotlib: 3.4.3
cftime: 1.6.2
cfunits: 3.3.5
cfplot: 3.1.31
cfdm: 1.10.0.3
cf: 3.14.1
```

davidhassell:
Fixes #623 

#### Notes
* The previous approach loaded both the source and destination grid masks onto the weights, which had to be a dense `numpy` array since `scipy` sparse arrays do not support masks.
* The new approach keeps the weights in a  `scipy` sparse array, but creates a separate boolean mask to applied to the regridded data. 
* The change to `cf/regrid/regrid.py` pre-creates the sparse matrix representation in a returned regrid operator. This makes sense because the purpose of returning the operator is for efficiency in repeated future uses, so it makes sense not to have to do this every time.
* The change from `coo_array` to `csr_array` is because the latter is much more efficient at row slicing.

### DIFF
--- a/cf/regrid/regridoperator.py
+++ b/cf/regrid/regridoperator.py
@@ -499,6 +499,17 @@ class RegridOperator(mixin_Container, Container):
     def tosparse(self):
         """Convert the weights to `scipy` sparse array format in-place.
 
+        The weights are converted to Compressed Sparse Row(CSR) array
+        format, i.e. the `weights` attribute becomes a
+        `scipy.sparse._arrays.csr_array` object. A CSR array used as
+        the most efficient sparse array type given that we expect no
+        changes to the sparsity structure, and any further
+        modification of the weights to account for missing values in
+        the source grid will always involve row-slicing.
+
+        The `dst_mask` attribute is updated to `True` for desitnation
+        grid points for which the weights are all zero.
+
         :Returns:
 
             `None`
--- a/cf/regrid/regridoperator.py
+++ b/cf/regrid/regridoperator.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 
+import numpy as np
 from cfdm import Container
 
 from ..decorators import _display_or_return
@@ -339,10 +340,18 @@ class RegridOperator(mixin_Container, Container):
                 The deep copy.
 
         """
+        row = self.row
+        if row is not None:
+            row = row.copy()
+
+        col = self.col
+        if col is not None:
+            col = col.copy()
+
         return type(self)(
             self.weights.copy(),
-            self.row.copy(),
-            self.col.copy(),
+            row,
+            col,
             method=self.method,
             src_shape=self.src_shape,
             dst_shape=self.dst_shape,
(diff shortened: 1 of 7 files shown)


sadielbartholomew:
FYI I just added one minimal commit to resolve the lone trivial merge conflict (new Changelog entries to combine) so I can review without it being in the way.

cf/regrid/regridoperator.py
```diff
@@ -487,54 +496,25 @@ def parameters(self):
...
 
         """
+        row = self.row
+        if row is None:
+            return
+
         from math import prod
 
-        from scipy.sparse import coo_array
+        from scipy.sparse import csr_array
```
sadielbartholomew, line 513: Regarding this change and your related comment:

> The change from coo_array to csr_array is because the latter is much more efficient at row slicing.

I notice the SciPy docs say that the CSR method has such an advantage but [also disadvantages of](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.html):

> slow column slicing operations (consider CSC)
> changes to the sparsity structure are expensive (consider LIL or DOK)

but I believe those disadvantages aren't relevant in this use context, unless I am missing some case where the latter may be detrimental, and if so if you have considered those alternatives suggested for suitability?

sadielbartholomew (approved):
Clever approach which I agree with as covered in the notes of the opening comment. The implementation is sound and doesn't call for any new testing. The performance improves very significantly in test cases I played around with and did basic profiling on, so passes with flying colours in that respect.

Therefore overall this is all good, though I've raised some minor comments and a question. Please consider those and then merge when ready. Thanks.

cf/regrid/regridoperator.py
davidhassell: That's it - not going to change the sparsity, and will always do at least as many row slices as column slices. I'll make a note: d94e5cf5124e8b23cb517675b5b58bfcf6f2dce4

PR #627, opened 2023-03-29
Fix bounds properties on computed vertical coordinates

### LINKED ISSUE #626: Bounds of computed vertical coordinates retain incorrect properties
When computing non-parametric vertical coordinates with `cf.Field.compute_vertical_coordinates`, the computed bounds can inherit properties from the parametric coordinates, which is incorrect. Here's an example of incorrect retained bounds that shows the bounds having an inconsistent standard name: 

```python
    Auxiliary coordinate: altitude
        long_name = 'Computed from parametric
                     atmosphere_hybrid_height_coordinate vertical coordinates'
        standard_name = 'altitude'
        units = 'm'
        Data(50, 201, 200) = [[[20.0, ..., 18220.9921875]]] m
        Bounds:formula = 'z = a + b*orog'
        Bounds:standard_name = 'atmosphere_hybrid_height_coordinate'
        Bounds:units = 'm'
        Bounds:Data(50, 201, 200, 2) = [[[[0.0, ..., 18681.5124544]]]] m
```

The solution is simply to clear these bounds properties.

```python
>>> import cf
HDF5 library: 1.12.1
netcdf library: 4.8.1
ESMF: 8.2.0
Python: 3.10.9
dask: 2023.3.0
netCDF4: 1.6.0
psutil: 5.9.4
packaging: 23.0
numpy: 1.22.3
scipy: 1.8.1
matplotlib: 3.4.3
cftime: 1.6.2
cfunits: 3.3.5
cfplot: 3.1.31
cfdm: 1.10.0.3
cf: 3.14.1
```

davidhassell:
Fixes #626

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -5,6 +5,9 @@ version 3.14.2
 
 * Fixed bug in `cf.Field.read` when reading UM/PP data that are
   runlength encoded (https://github.com/NCAS-CMS/cf-python/issues/621)
+* Fixed bug in `cf.Field.compute_vertical_coordinates` that retained
+  incorrect bounds properties
+  (https://github.com/NCAS-CMS/cf-python/issues/626)
 * Removed benign UserWarning from `cf.Field.percentile`
   (https://github.com/NCAS-CMS/cf-python/issues/619)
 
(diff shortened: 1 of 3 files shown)


sadielbartholomew (commented):
I agree with the approach to fix the underlying issue by clearing the inappropriately-inherited properties on the bounds. Fix tested as working locally and via the CI (the few failing jobs are failing on setup issues I shall address elsewhere soon).

I would therefore approve this, but feel I must encourage us to add a test to capture the bug before supporting a merge! If there's good enough reason to skip some new testing along those lines, then fair enough, but do provide a little bit of justification. Then, either way, I'll pop the approval up. Thanks.

PR #632, opened 2023-04-12
Fix `Field.set_data` checking

davidhassell:
Tightens up `Field.set_data` checking, preventing some cases that could cause inconsistencies downstream failures - better to trap the failures in set_data.


sadielbartholomew (approved):
Perfect. Please merge!

PR #643, opened 2023-04-28
Deprecate `cf.relative_vorticity`

davidhassell:
Deprecates `cf.relative_vorticity`, and allows `cf.Field.laplacian_xy`, `cf.Field.grad_xy`, `cf.curl_xy` and `cf.div_xy` to work in cases when the dimension coordinates are missing standard names.


sadielbartholomew:
Out of interest, why is `cf.relative_vorticity` being deprecated?

davidhassell:
Hi Sadie - (following our chat), we decided to not fix it nor alias it to `curl_xy` because it's spherical polar coordinate fomula was wrong (well, it was an approximation rather than an exact formula), and its API was different to `curl_xy`, so aliasing it was not straight forward.

bewithankit (approved):
Hi David, Thank you for the changes! Looks all good, merge when ready!

PR #646, opened 2023-05-03
Correct `units` test in `propertiesdata.set_data`

davidhassell:
The original tested `if units is not None:` with `units` being a `cf.Units` object. This is nonsense, so the test should be `if units:`.

The new test is for the use case that brought this to light.


sadielbartholomew (approved):
Good spot. Sensible fix. The new test does indeed fail on the present `main`, so good idea to bring it into the test suite to capture any future issues.

Perfect, please merge.

PR #653, opened 2023-05-05
Added recipe to calculate and plot Relative Vorticity

bewithankit:
Added recipe to to calculate and plot Relative Vorticity. The deployment of the same could be previewed at https://bewithankit.github.io/cf-python/recipes/plot_10_recipe.html.


davidhassell (changes requested):
Some wording suggestions. Please merge when you've condisered them (and applied the select_field and later subspacing as discussed offline)

PR #654, opened 2023-05-10
Fix the return type of `cf.example_fields`

davidhassell:
Fix the return type of `cf.example_fields`.


bewithankit:
Thanks David! Looks good, merge when ready!

PR #657, opened 2023-05-23
Added recipe to plot Climate Stripes

bewithankit:
Added recipe to plot Climate Stripes. The deployment of the same could be previewed at https://bewithankit.github.io/cf-python/recipes/plot_11_recipe.html.


docs/source/recipes/plot_11_recipe.py
```diff
@@ -0,0 +1,89 @@
+"""
+Plotting the Warming Stripes
+============================
+
+In this recipe, we will plot the Warming Stripes (Climate Stripes) created by
+Professor Ed Hawkins at NCAS, University of Reading. Here we will use the
+`HadCRUT.5.0.1.0 dataset
+<https://www.metoffice.gov.uk/hadobs/hadcrut5/data/current/download.html>`_ for
```
sadielbartholomew, line 8: Good idea linking to the source, but I think it would be useful to also indicate which dataset(s) specifically are used, so that people can follow along and there is quite a big listing of various datasets available from that page.

sadielbartholomew (commented):
Fantastic idea :smiley: And I am glad (but not surprised :wink: ) that we can make the plot so easily with cf-python!

It looks all good overall. I've raised some minor comments (mostly to do with the numbering needing a tweak for a duplicate), many of which will need addressing in other equivalent places should you agree with, and make, any changes which I raised on `docs/source/recipes/plot_11_recipe.py` only.

Please merge when you feel you've addressed my feedback (or if you want a final re-review the just let me know). Thanks.

bewithankit:
Thank you for the suggestions Sadie. And yes! I was pleasantly surprised as to how easy it was with cf-python!

sadielbartholomew (approved):
All feedback addressed.

One final comment, namely that upon generating the final figure I notice there is some barely-visible, but just about, whitespace between the 'stripes', which is clearer when you zoom in a bit, e.g. as shown below:

Screenshot at 2023-05-24 18-35-04

Ideally we can add some logic that removes those, assuming it would only be a line or two at most so as not to ruin the simplicity of the script otherwise. Do you think that is possible?

(Sorry for being pernickety, but it would mean that any zooming in the user wants to do will show subsets of the stripes in the standard formatting, rather than barcode-like stripes with white inbetween.)

Please do merge though if you don't think such a tweak is necessary, or it can't be done in a few lines. Thanks.

bewithankit:
Hi Sadie, it could be achieved using [matplotlib.axes.Axes.fill_between](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.fill_between.html#matplotlib.axes.Axes.fill_between), but then the number of parameters increase while plotting:
```python
fig_global, ax_global = plt.subplots(figsize=(10, 2))

for i in range(global_avg_temp.shape[0]):
    ax_global.fill_between(
        [i, i + 1],
        0,
        1,
        color=cmap(norm_global(global_avg_temp[i])),
        linewidth=0,
    )

ax_global.axis("off")

plt.show()
```
Figure 1_001

Do you think we should use [matplotlib.axes.Axes.axvline](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axvline.html) with less parameters or [matplotlib.axes.Axes.fill_between](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.fill_between.html#matplotlib.axes.Axes.fill_between)?

davidhassell (approved):
Brilliant, Ankit. I've nothing to add from Sadie's review, other than perhaps to link to the wiki page that describes the stripes. Please merge when you and Sadie are happy.

docs/source/recipes/plot_11_recipe.py
```diff
@@ -0,0 +1,90 @@
...
+
+# %%
+# 9. Create the figure and the axes for the global plot. Loop over the selected 
+# years, plot a colored vertical stripe for each and remove the axes:
+fig_global, ax_global = plt.subplots(figsize=(10, 2))
+
+for i in range(global_avg_temp.shape[0]):
+    ax_global.axvline(
+        x=i, color=cmap(norm_global(global_avg_temp[i])), linewidth=4
+    )
```
sadielbartholomew, line 66: > Do you think we should use [matplotlib.axes.Axes.axvline](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axvline.html) with less parameters or [matplotlib.axes.Axes.fill_between](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.fill_between.html#matplotlib.axes.Axes.fill_between)?

I had a little play around with the options, and actually I think neither! From looking in the docs to see what these do, I noticed an alternative,[`axvspan`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axvspan.html), which is perfect for what we need, because it can be set to start at the end of the last stripe, horizontally considered, and end at the start of the next one, and so on:

```suggestion
for i in range(global_avg_temp.shape[0]):
    ax_global.axvspan(
        xmin=i-0.5,
        xmax=i+0.5,
        color=cmap(norm_global(global_avg_temp[i]))
    )
```

This means that regardless of the zoom level applied, there won't be any gaps between the stripes, e.g:

cs_zoom_Figure_1

Whereas `linewidth` varies according to zoom level so gaps will always be liable to appear. So this way, i.e. that in the suggestion, seems most robust to me. And the code stays simple, especially when the arguments are listed out as `xmin=`, `xmax=` to make it clear what we're plotting.

Does it seem OK to you as a solution? See also the regional stripes where I am making the equivalent suggestion.

docs/source/recipes/plot_11_recipe.py
```diff
@@ -0,0 +1,90 @@
...
+uk_avg_temperature = uk_temperature.collapse("X: Y: mean")
+uk_avg_temp = uk_avg_temperature.array.squeeze()
+norm_uk = plt.Normalize(uk_avg_temp.min(), uk_avg_temp.max())
+
+# %%
+
+fig_uk, ax_uk = plt.subplots(figsize=(10, 2))
+
+for i in range(uk_avg_temp.shape[0]):
+    ax_uk.axvline(x=i, color=cmap(norm_uk(uk_avg_temp[i])), linewidth=4)
```
sadielbartholomew, line 86:
```suggestion
    ax_uk.axvspan(xmin=i-0.5, xmax=i+0.5, color=cmap(norm_uk(uk_avg_temp[i])))
```

(See comment above.)

sadielbartholomew (approved):
In regards to your latest comments...

After this is addressed I believe we are ready to merge, hence I'll put in a formal 'approval'.

docs/source/recipes/plot_11_recipe.py
bewithankit: Ah nice! [`axvspan`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.axvspan.html), is the perfect one that we need! Thanks Sadie!

PR #659, opened 2023-05-30
Scrollable Navigation Menu (Sidebar)

### LINKED ISSUE #658: Docs: immobile navigation menu can be cut off
Equivalent to NCAS-CMS/cfdm#50, but more urgent for cf-python due to that menu being significantly longer than the cfdm docs one as of late, due to various new sections being added over the last few years.

As a solution, I have a strong preference for us just adding a scrollbar to the sidebar. Any disagreement on that then please do shout! But the scrollbar seems like a natural solution that is (probably) easy enough to add.

bewithankit:
Might resolve #658.

### DIFF
--- a/docs/source/_static/customise-alabaster.css
+++ b/docs/source/_static/customise-alabaster.css
@@ -7,6 +7,8 @@ div.sphinxsidebar {
     width: {{ theme_sidebar_width }};
     font-size: '13pt';
     line-height: 1.5;
+    max-height: 100%;
+    overflow-y: auto;
 }
 
 /* -- body styles ----------------------------------------------------------- */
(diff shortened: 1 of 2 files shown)


bewithankit:
Hi Sadie, yes it is ready for review.

sadielbartholomew (approved):
A simple solution that resolves the issue at hand. Great stuff!

Admittedly, because re-building the documentation takes a fair bit of time nowadays and I was short on it today, I tested this via adding the lines added for the PR in the appropriate place in the CSS directly using the browser developer console. I can confirm from this, that the scroll bar appears and enables scrolling of the sidebar/left area only, which is exactly what we need:

scroll-cf-one-crop

And I agree, as we discussed last week, that removing the 'Related Topics' section is a good idea, so thanks for including that.

Perfect, please merge.

bewithankit:
Thank you for the review Sadie!

PR #660, opened 2023-06-02
Automate Recipe filtering

bewithankit:
The old way to create recipes with filters is actually [a cumbersome process](https://github.com/NCAS-CMS/cf-python/blob/b6a7c7eca7001b92609726e30816356bb6a57892/DOCUMENTATION.md) which involves generating the documentation first, then copying new generated lines (also the new toctree mentions) from `/docs/_sources/recipes/index.rst.txt` to `docs/source/recipes/filter.rst` which acts like a template where the Javascript, CSS and some of the HTML elements for the filters reside. The documentation then has to be regenerated to get `filter.html` which has to be renamed to `index.html` while deleting the previous `index.html`. The new idea gets rid of the whole editing `.rst` files as they have to be generated twice and focuses on getting the filters working directly in the` index.html` page. Although it gets rid of a lot of previous copy-pasting steps it is still 'manually automated' which makes use of the stream editor `sed` to replace the lines like:
```html
<div class="sphx-glr-thumbcontainer" tooltip="Vorticity, the microscopic measure of rotation in a fluid, is a vector field defined as the cur...">
```
with
```html
<div class="sphx-glr-thumbcontainer subspace contourmap maths" tooltip=" Subspace, Contourmap, Mathematical Operations">
```
Putting this as draft as I would like to somehow get the previous lines starting with `<div class="sphx-glr..` for every recipe to get replaced, so that it goes from something like:
```html
<div class="sphx-glr-thumbcontainer" tooltip="Vorticity, the microscopic measure of rotation in a fluid, is a vector field defined as the cur..."><img alt="Calculating and plotting the relative vorticity" src="../_images/sphx_glr_plot_10_recipe_thumb.png" />
<p><a class="reference internal" href="plot_10_recipe.html#sphx-glr-recipes-plot-10-recipe-py"><span class="std std-ref">Calculating and plotting the relative vorticity</span></a></p>
  <div class="sphx-glr-thumbnail-title">Calculating and plotting the relative vorticity</div>
</div>
```
to
```html
<div class="sphx-glr-thumbcontainer subspace contourmap maths" tooltip=" Subspace, Contourmap, Mathematical Operations"><img alt="Calculating and plotting the relative vorticity" src="../_images/sphx_glr_plot_10_recipe_thumb.png" />
<p><a class="reference internal" href="plot_10_recipe.html#sphx-glr-recipes-plot-10-recipe-py"><span class="std std-ref">Calculating and plotting the relative vorticity</span></a></p>
  <div class="sphx-glr-thumbnail-title">Calculating and plotting the relative vorticity</div>
</div
```
The whole process would read from something like a `recipe_list.txt`:
```txt
plot_1_recipe.html#sphx-glr-recipes-plot-1-recipe-py
<div class="sphx-glr-thumbcontainer collapse lineplot" tooltip="Collapse, Lineplot">
plot_2_recipe.html#sphx-glr-recipes-plot-2-recipe-py
<div class="sphx-glr-thumbcontainer collapse lineplot subspace" tooltip="Collapse, Lineplot, Subspace">
plot_3_recipe.html#sphx-glr-recipes-plot-3-recipe-py
<div class="sphx-glr-thumbcontainer collapse contourmap" tooltip="Collapse, Contourmap">
...
```
which could be edited with each added recipe to do the changes. @davidhassell and @sadielbartholomew if you have any ideas do let me know! Thanks!
___
UPDATE: Using `awk` to see if it could be accomplished.


davidhassell (approved):
Difficult to test in the absence of a new recipe - but it looks good! Let's merge it and get someone other than you to test it when the next recipe comes along.

PR #662, opened 2023-06-07
Added PDF for Cheat Sheet

bewithankit:
Added a printable PDF for the cf-python Cheat Sheet. Used Canva to create it keeping in mind the colours and fonts of our website and the ease of use. To keep it updated with any API changes (as suggested by @sadielbartholomew), I am thinking of sharing the editable  [link](https://www.canva.com/design/DAFk9_BVfNY/gmQHycBiV_YbTIWMqYxK1g/edit) somewhere in the README so that anyone would be able to make changes in the future.


sadielbartholomew (commented):
Looks fantastic, nice job!

My only thought in feedback is that perhaps it would be better to drop the `>>>` interactive prompt on all lines, since we don't seem to be (unless I've missed something) showing any outputs here, just the calls themselves (I agree this is best with the limited space), and the prompt therefore can be assumed and otherwise just adds noise to the code snippets, in my opinion. And without those, multiple code lines and any blocks can be copied directly (obviously a printed sheet is the main use case here, but folk might still check it on a screen). What do you think?

bewithankit:
I think you're right Sadie. There is no need for the `>>>` prompt and it'll be easier to copy. Thanks!

RELEASE.md
```diff
@@ -47,6 +47,12 @@
 * **Follow all of the steps outlined externally in `DOCUMENTATION.md`**,
   notably so that the correct Sphinx-related environment is prepared for
   documentation building.
+  
+* Ensure that the [PDF for Cheat Sheet](docs/_downloads/cheatsheet.pdf) 
+  is updated to include any API changes. The PDF is created using Canva 
+  keeping in mind the colours and fonts of the website. The same could 
+  be edited using this 
+  [link](https://www.canva.com/design/DAFk9_BVfNY/gmQHycBiV_YbTIWMqYxK1g/edit).
```
sadielbartholomew, line 55: Nicely clarified here.

Though I am wondering, how would we access the current version of the A4/printable cheat sheet before it is exported as PDF? Is there an option on Canva to save the cheat sheet in editable/build form so that it can be edited and re-exported and saved as PDF when we want to do that, and if so please can we save that also to the repo? Better to have it stored on GitHub than having to possibly pester you for such a resource anytime it needed an update, or re-write the cheat sheet from scratch each time for lack of one, if that makes sense!

sadielbartholomew (approved):
Thanks for addressing my previous item of feedback. A few new minor comments, but once those are addressed, I am happy to merge. Namely (see also one in-line):

* Nice touch adding the links at the bottom right, though the 'Tutorial' one mistakenly links to the API Reference at present;
* We can also remove the `...` continuation marker left in a couple of places, as that is also an interactive prompt thing and not needed (and indeed a bit confusing and preventing copying of lines) now the prompts are not included in the code snippets.

RELEASE.md
bewithankit: Hi Sadie, Sorry, I didn't get what you are trying to explain! The given link itself contains the Cheat Sheet in an editable form which can be exported as a PDF, PNG or any format of choice.

PR #663, opened 2023-06-09
Remove archived documentation from repo, point to new location

sadielbartholomew:
Removes the built archived documentation (essentially that contained in the `docs/3.*` directories) which has moved to its new home in the [NCAS-CMS/cf-python-docs repository](https://github.com/NCAS-CMS/cf-python-docs) and updates the links to such archived docs so that the latest documentation points to the new location.

~Leaving as a draft until we can work out why the styling is not being applied correctly on the shifted docs over in the new repo home (I for one am a bit stumped as to why the CSS and other assets aren't being picked up from there, but maybe we just need to wait for those to be registered?)...~


sadielbartholomew:
12 million lines down, that's quite something! :sunglasses:

sadielbartholomew:
Approved offline by @davidhassell, thanks David. Merging...

sadielbartholomew:
CI jobs aren't relevant, so stopping them.

PR #669, opened 2023-06-13
Added logo

### LINKED ISSUE #508: Set of related logos for the NCAS CF Python libraries
We've noted previously that it would be good to get an official logo for each of the NCAS CF Python libraries. It would be good to get the ball rolling with thinking about designs and prototyping, so this issue can serve as a record of thoughts and progress.

By 'logo', I suggest we create, in line with most software logo resources, either an image and some text for the library name that is incorporated into the image, or separate to it but usually displayed by it.

I think given the links between these libraries, including the dependency for some pairs, it would be best (and also most efficient) to create a set of logos together that are inter-related in some aspect. to highlight the grouping, but also with slight differences across the set to distinguish.

Assuming agreement and ultimate permission from the original developers, I think the following five would be a good group to include as the set, but with an underlying idea for a theme to relate the logos, it should be quite simple to readily create a new design that fits with the rest should one or more new NCAS CF Python tools ermege (so we can keep in mind extensibility):
* cfdm;
* cf-python;
* cfunits;
* cf-plot;
* cf-checker.

If there is consensus for the above, then I believe we have to consider the following:
* what a good logo design would be (image, font, relative dimensions and placement, etc.);
* how we can vary the design so it has differences across the set of libraries (other than the library name text);
* what we should keep the same across the set to tie the logos together.

### General thoughts

My thoughts in brief are:

1. Given the CF Conventions are at the heart of the libraries it seems natural that this is highlighted in the logo. The 'CF' initials feature across the library names so it would be ideal to get those in the logo image rather than just the library name text.

2. Given the conventions are based on the Earth Sciences, having a sketch or plot of the all or part of the earth under some projection as the image seems sensible (at least, I can't think of other good ideas for what the image should be).

3. The earth is a bit difficult to draw out so it seems like a good idea to plot it for the image as a map projection using a tool, naturally cf-plot!

4. ...but a logo shouldn't be too detailed as it usually will be shown with small dimensions so any map projection plot used for the image should be quite simplistic as to not look too fussy or not work in fairly low-resolution.

5. A nice idea given the above would be to put the library name text over some earth map projection, but we might be straying too close to [the Cartopy logo](https://scitools.org.uk/cartopy/docs/latest/_static/cartopy.png) with that and it is best if we don't have a logo that resembles any of similar tools in use in our domain in case it looks like we copied or at least weren't very original...

### A proposal, with rough sketches

I thought about this earlier in the year and my favourite idea that I could come up with based on the above thoughts, which I will propose here as a candidate for the shared basis of the logos, is to decompose some elliptical-ish projection of the earth using some major displayed gridding lines into a 'C' and 'F' letter.

The following sketch tries to illustrate the idea, if you can forgive the crudeness (I used a crude paint tool app to draw on a plotted cartopy projection as it was the easiest to capture the idea):

orig_white

Because the two letters might not be clear, I thought the other parts should be in a different colour, and made slightly smaller on the left and larger on the right so as to imply a forward trajectory, as if it were showing snapshots in time going forward in the plane, where lines as shown could help to emphasise that:

final_logo_idea

My thinking is that the colours would be best coming from some simple [see (4)], probably blockfill, plot on the chosen projection, plotted using cf-plot. Then the grey divisions would be the same or a similar dataset also in blockfill but in a colour scale that is very different and more subtle, perhaps greyscale. They could be cut into placement as shown in the sketches manually using a photo editing tool.

#### Questions

It's not clear with the proposal as-is:

* where the library name text should go in each case;
* how to vary the design across the set;
* data and colour palettes, etc., to plot for each earth section,

bewithankit:
Resolves #508. 
A quick preview:
image


sadielbartholomew (commented):
The logo design and execution are both fantastic. Brilliant work :100:

***

_TLDR: I think we should decide the (related) logos for all other NCAS Data Tools libraries before committing to this very specific design for cf-python, so should get agreement on those after some input from Andy and Ros (and more from David too), maybe even the whole of CMS, before merging._

But before we merge this to christen the final logo, I am just wondering what the plan might be for logos for the other NCAS Data Tools libraries? Whilst obviously cf is of paprticular concern here, I think it would be wise to work out what logo we want across all such libraries before committing to one for a given library. From my opening comment on #508 (quoting myself!): 

> I think given the links between these libraries, including the dependency for some pairs, it would be best (and also most efficient) to create a set of logos together that are inter-related in some aspect. to highlight the grouping, but also with slight differences across the set to distinguish.
> 
> Assuming agreement and ultimate permission from the original developers, I think the following five would be a good group to include as the set, but with an underlying idea for a theme to relate the logos, it should be quite simple to readily create a new design that fits with the rest should one or more new NCAS CF Python tools ermege (so we can keep in mind extensibility):
> 
> cfdm;
> cf-python;
> cfunits;
> cf-plot;
> cf-checker.

(We can now add cf-view to the list!) Of course, we'd want input and permission from Andy and Ros as creators of some of those as to what they might want.

My main thought would be that we can use the precise logo shape and form as added here, but vary the colours a bit for the six different libraries to distinguish them, and also add some small text curving around the globe to type the name of the library, i.e. add '-python', ''dm', etc. in small letters in each case. But that's just one idea. I definitely think folk would be keen to use a variant of this particular logo, because it's great IMO, but we should ask to be sure (I don't think it will time-consuming or controversial, hopefully just a few quick discussions)...

What do you think Ankit and @davidhassell? (In particular Ankit are you happy to create several variants on the logo we can use across other libraries, and check the library originators are happy to use them?)

bewithankit:
That's true Sadie, we should give it a thought because new NCAS CF Data Tools will keep growing (George's upcoming cf-store?) and it'll be great if they are all cohesive. Also, (not the main point but) this will make it easier to highlight the bunch of tools together when it comes to the next board meeting and stuff. We need to involve Bryan in this discussion too.

davidhassell (approved):
Looks good, let's go!

PR #670, opened 2023-06-14
Docs: remove or replace cases of some subjective filler terms

sadielbartholomew:
Remove or replace the use of the words across the documentation that are considered bad practice (see, for example, [the Google dev docs style guide list of words to not use or to avoid](https://developers.google.com/style/word-list)) given the implication that something is easy when a user might not find it so, and that they are also effectively filler words so unnecessary, namely (see the case-by-case commits which tackle all cases of use of each related term):

* "simple" and "simply";
* "easy" and "easily";
* "just" (converting to "only" in the case that this is what is meant, being unambiguous).

This PR also fixes some sentences with minor typos that I have noticed in the process (see final commit).

### DIFF
--- a/docs/source/introduction.rst
+++ b/docs/source/introduction.rst
@@ -113,8 +113,7 @@ manipulation and can:
 **Visualisation**
 -----------------
 
-
-Powerful and flexible visualisations of field
+Powerful, flexible, and user-friendly visualisations of field
 constructs are available with the `cf-plot` package that is installed
 separately to `cf` (see http://ajheaps.github.io/cf-plot for details).
 
--- a/docs/source/introduction.rst
+++ b/docs/source/introduction.rst
@@ -38,7 +38,7 @@ to new datasets.This is so that datasets that are partially conformant
 may nonetheless be modified in memory.
 
 .. code-block:: python
-   :caption: *A simple example of reading a field construct from a
+   :caption: *A basic example of reading a field construct from a
              file and inspecting it.*
 
    >>> import cf
@@ -114,8 +114,8 @@ manipulation and can:
 -----------------
 
 
-Powerful, flexible, and very simple to produce visualisations of field
-constructs are available with the `cfplot` package, that is installed
+Powerful and flexible visualisations of field
+constructs are available with the `cfplot` package that is installed
 separately to `cf` (see http://ajheaps.github.io/cf-plot for details).
 
 See the `cfplot gallery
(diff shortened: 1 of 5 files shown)


docs/source/introduction.rst
```diff
@@ -114,13 +114,13 @@ manipulation and can:
 -----------------
 
 
-Powerful, flexible, and very simple to produce visualisations of field
-constructs are available with the `cfplot` package, that is installed
+Powerful and flexible visualisations of field
+constructs are available with the `cfplot` package that is installed
```
davidhassell, line 118: Could we somehow retain the notion that cf-plot is very easy to use (e.g. not all plotting libraries allow you to produce a plot with a one-liner)

docs/source/tutorial.rst
```diff
@@ -529,7 +529,7 @@ files <External-variables-with-cfa>`.
 **Visualisation**
 -----------------
 
-Powerful, flexible, and very simple to produce visualisations of field
+Powerful and flexible visualisations of field
```
davidhassell, line 532: Could we somehow retain the notion that cf-plot is very easy to use (e.g. not all plotting libraries allow you to produce a plot with a one-liner) (same as earlier)

davidhassell (commented):
Hi Sadie - Thanks, I really like these changes. Just one question/suggestion (asked twice).

docs/source/introduction.rst
sadielbartholomew: Indeed, I was wondering this myself - I guess we want to advertise that cf-plot is easy to use without appearing to be making bold and/or unsubstantiated claims such as 'very easy to use'.

> e.g. not all plotting libraries allow you to produce a plot with a one-liner

True, maybe this is the crux to focus in on. Can we say something along the lines of "_user-friendly_" or (something to lead into this) "with minimal setup" (e.g:

_"The `cf-plot` package, designed to produce visualisations from field constructs in as few lines/calls as necessary and in a powerful and flexible manner, ..."_

) which might highlight that, indirectly (former) or directly (latter)?

Just realising also we should correct it to `cf-plot` (avec hyphen), the actual library name! I'll do that now...

davidhassell: I think I prefer `Powerful, flexible, and user-friendly visualisations of field ...`. Your call.

sadielbartholomew: I am more than happy to go with 'user-friendly', a more concise choice :slightly_smiling_face: Great, I'll add that into both equivalent cases raised here, then mark this as resolved.

davidhassell (approved):
All good. Please merge!

PR #671, opened 2023-06-19
In `cf.read`, skip caching elements from compression index variables.

davidhassell:
Requires `cfdm 1.10.1.1`, specifically https://github.com/NCAS-CMS/cfdm/issues/260 (_Edit 2023-06-20: now merged into main_), from which it utilises the new `compression_index` keyword parameter of `NetCDF._create_data`.


cf/read_write/netcdf/netcdfread.py
```diff
@@ -181,6 +182,12 @@ def _create_data(
 
                 .. versionadded:: 3.15.0
 
+           compression_index: `str`, optional
```
sadielbartholomew, line 185: Sorry I missed this on the corresponding cfdm PR (where we have merged the same description), but a Boolean seems to be expected?

```suggestion
           compression_index: `bool`, optional
```

sadielbartholomew (approved):
One minor comment, but works and incorporates the speedup successfully. Please merge when ready.

PR #673, opened 2023-06-22
Update create_regular methods to remove cellsize divisibility check

bewithankit:
The previous iteration of `cf.Domain.create_regular` and `cf.DimensionCoordinate.create_regular` checked explicitly for the range of the dimension to be divisible by the cellsize which is removed here and a note about the same is added in the docstring.


davidhassell (approved):
Great - all good!

bewithankit:
Thanks David!

PR #680, opened 2023-07-07
Add recipe to plot AOD using masking

bewithankit:
Add recipe to plot AOD using masking. The deployment of the same could be previewed at https://bewithankit.github.io/cf-python/recipes/plot_12_recipe.html#sphx-glr-recipes-plot-12-recipe-py


docs/_downloads/plot_12_recipe.py
```diff
@@ -0,0 +1,116 @@
...
+# `plt.suptitle <https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.suptitle.html>`_
+# is used to add a title for the whole figure; 
+# the subplots for plotting are selected using 
+# `cfplot.gpos <https://ajheaps.github.io/cf-plot/gpos.html>`_ after which 
+# `cfplot.mapset <https://ajheaps.github.io/cf-plot/mapset.html>`_ is used to 
+# set the map limits and resolution for the subplots;  
+# as cf-plot stores the plot in a plot object with the name 
+# ``cfp.plotvars.plot``, country borders are added using normal 
+# `Cartopy operations <https://scitools.org.uk/cartopy/docs/latest/reference/index.html>`_ 
+# on the ``cfp.plotvars.mymap`` object:
```
sadielbartholomew, line 85: Really great walkthrough of the plotting code :100:.

Since it's quite a long compound summary, split by semi-colon (which is the best approach otherwise and still works), I think it might be more readable to instead use a bullet point list to split up the various statements. Is that possible in this workflow where the comments in the scripts, like this file, seem to be converted to the rendered page text? If it's not possible and too much of a pain to achieve, then no worries, since the semi-colon separated list is perfectly fine, just a bit harder to pick out each statement from.

sadielbartholomew (approved):
Excellent new recipe. Thanks Ankit, great stuff.

I've raised some very minor comments in-line. One other, final, one is that the first plot has quite a lot of vertical whitespace around it as-is which ideally could be trimmed a bit (via whatever cf-plot or matplotlib calls achieve this in a simple way) so it isn't as detached from the text for that step, and the next step's text:

...
scr_whitespace

Nothing I've mentioned is a blocker though, so feel free to merge this once you feel ready.

bewithankit:
> Excellent new recipe. Thanks Ankit, great stuff.
> 
> I've raised some very minor comments in-line. One other, final, one is that the first plot has quite a lot of vertical whitespace around it as-is which ideally could be trimmed a bit (via whatever cf-plot or matplotlib calls achieve this in a simple way) so it isn't as detached from the text for that step, and the next step's text:
> 
> ... scr_whitespace
> 
> Nothing I've mentioned is a blocker though, so feel free to merge this once you feel ready.

Hi Sadie, thank you for all the suggestions! :raised_hands:

The thing is when I try to define the plot area using `cfp.gopen(top=0.5)`, it works normally and gets rid of the whitespace:

Selection_045

But Sphinx-Gallery behaves in a different way when it tries to generate the plot and renders it with whitespace around it:

ImageMagick: cfplot png_001

PR #688, opened 2023-08-18
Add recipe to resample land use flags to a coarser grid

bewithankit:
Dhirendra had some land use data at a resolution of 100 m, with several flags defined with numbers representing the type of land use. Regridding the data to a corser resolution of approximately 25 km would have incorrectly represented the flags on the new grids. This recipe resamples land use flags to a coarser grid (while showcasing the use of [`cf.Field.digitize`](https://ncas-cms.github.io/cf-python/method/cf.Field.digitize.html), [`cf.histogram`](https://ncas-cms.github.io/cf-python/function/cf.histogram.html) and [`cartopy.io.shapereader.natural_earth`](https://scitools.org.uk/cartopy/docs/v0.15/tutorials/using_the_shapereader.html)). The deployment of the same could be previewed at https://bewithankit.github.io/cf-python/recipes/plot_15_recipe.html#sphx-glr-recipes-plot-15-recipe-py.


docs/source/recipes/plot_15_recipe.py
```diff
@@ -0,0 +1,153 @@
...
+#   along with the digitized data.
+# - The `np.linspace <https://numpy.org/doc/stable/reference/generated/numpy.linspace.html>`_
+#   function is used to create an array of evenly spaced bin edges from 0 to 50,
+#   with 51 total values. This creates bins of width 1.
+# - The ``digitized`` variable contains the bin indices for each data point, 
+#   while the bins variable contains the actual bin values.
+# - The `cf.histogram <https://ncas-cms.github.io/cf-python/function/cf.histogram.html>`_ 
+#   function is called on the digitized data to create a histogram. This 
+#   function returns a field object with the histogram data.
+# - The ``h.array.squeeze()`` function extracts the histogram data as a NumPy
```
sadielbartholomew, line 89:
```suggestion
# - The `squeeze <https://ncas-cms.github.io/cf-python/method/cf.Field.digitize.html>`_`
# function applied to the histogram ``array`` extracts the histogram data as a NumPy
```

would be better to include a link and to the relevant part of our API docs, I think (obviously no need to link to external library API docs for their functions etc., e.g. the matplotlib docs, though often they get linked in the code snippets anyway which is nice).

docs/source/recipes/plot_15_recipe.py
```diff
@@ -0,0 +1,153 @@
...
+countries = ["Ireland", "Belgium", "Switzerland"]
+
+# %%
+# 6. Set up the figure and axes for plotting the histograms:
+#
+# - The ``plt.subplots`` function is called to set up a figure with three 
+#   subplots, with a figure size of 8 inches by 10 inches.
+# - A loop iterates over each country in the countries list and for each 
+#   country, the ``extract_data`` function is called to extract the land use 
+#   data for that country.
```
sadielbartholomew, line 139:
```suggestion
#   country, the ``extract_data`` function is called to extract its land use 
#   data.
```

just to avoid using the word 'country' four times which feels quite repetitive!

docs/source/recipes/plot_15_recipe.py
```diff
@@ -0,0 +1,153 @@
...
+# - The ``plot_histogram`` function is then called to plot the histogram of land
+#   use distribution on the corresponding subplot.
+# - The ``plt.tight_layout`` function is called to ensure that the subplots are 
+#   properly spaced within the figure and finally, the ``plt.show function``
```
sadielbartholomew, line 143:
```suggestion
#   properly spaced within the figure and finally, the ``plt.show`` function
```

sadielbartholomew (approved):
Excellent recipe, nice job. Also, good country choice to display the land use data for, Charlotte and Andrea will be interested! :smile:

I've raised some minor comments, so do consider those, but this is all good overall, so please merge when ready.

docs/source/recipes/plot_15_recipe.py
bewithankit: Good one, I think you meant https://ncas-cms.github.io/cf-python/method/cf.Field.squeeze.html right?

bewithankit:
Thank you for the suggestions Sadie! :raised_hands:

PR #694, opened 2023-09-19
Fix bug in `cf.Field.match_by_construct`

### LINKED ISSUE #691: `cf.Field.match_by_construct` always returns `True` for 1-d constructs whose axis is not in the data
At v3.15.3, `cf.Field.match_by_construct` always returns `True` when matching against a size 1 construct that whose domain axis is not spanned by the field's data array, regardless of the condition:

```python
>>> import cf
>>> f = cf.example_field(0)
>>> print(f)
Cell methods    : area: mean

>>> f.match_by_construct(time=cf.dt('2019-01-01'))
True
>>> f.match_by_construct(time=cf.dt('1959-12-16'))
True
```

davidhassell:
Fixes #691

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,13 @@
+version 3.15.4
+--------------
+
+**2023-??-??**
+
+* Fix bug `cf.Field.match_by_construct` that always returned True for
+  1-d constructs whose axis is not in the data, even when the
+  criterion was not matched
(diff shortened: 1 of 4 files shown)


sadielbartholomew (approved):
Good spot on the bug. I can confirm the PR fixes it and provides new testing to expose it pre-fix. All good, please merge.

Issue #628, opened 2023-03-29
2D lat-lon coordinates calculation for CF-supported projections


sadielbartholomew:
Add, as a new feature, functionality which creates corresponding 2D latitude and longitude coordinates (as outputs) from 1D projection coordinates and coordinate reference parameters (as inputs), considering bounds if they are relevant.

This is a useful and important feature, above all, because regridding explicitly requires knowledge/input of the 2D lat-lon coordinates when considering fields with coordinates defined by a projection, even though they are superfluous and hence optional for projection cases, so the feature would allow us to support the full set of CF-supported projections in regridding operations.

### Implementation plans and notes

* Achieve the above by making use of the pyproj library (see below) as a new dependency to cf-python, assuming it is capable of effectively and efficiently enabling us to do so (determine this first ideally, or if necessary during the process!). It will be needed to convert from 1D projection coordinates to the 2D lat-lon coordinates they correspond to for the given projection.
* Should only need (as sufficient) to cover the projections supported by the CF Conventions (see reference below RE Appendix F).
* Probably this will ultimately be (user-facing wise) in the form of a new function called `create_2D_lats_and_lons` or similar, analogous to the already-existing `create_bounds` method, which will be linked up to the regridding methods to ensure they utilise the created coordinates to regrid where applicable.
* As a subsidiary issue, there is current functionality that manages as a special case of this problem the conversion from rotated poles to regular lats and lons, which we will want to subsume into this new feature to avoid duplication of effectively the same functionality for the special case. This could be done after the main work, or as part of it. 

### Relevant standards and libraries

* **PROJ** (previously called PROJ.4), a cartographic projections and coordinate transformations library/standard: see e.g. the [documentation](https://proj.org/index.html) here;
* **pyproj** library, a Python interface to PROJ: see e.g. it on [PyPI](https://pypi.org/project/pyproj/), [GitHub](https://github.com/pyproj4/pyproj);
* (obviously!) the **CF Conventions**, with relevant chapters being:
    * [5.6, ‘Horizontal Coordinate Reference Systems, Grid Mappings, and Projections’](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#grid-mappings-and-projections);
[...]

PR #695, opened 2023-09-20
Projection lat-lon coordinates part 1: `GridMapping` objects

sadielbartholomew:
Part 1 of 2 towards #628. This first part sets up the Grid Mapping class and helper function infrastructure towards interfacing with PROJ using pyproj, so that in the second part (nearly ready, but with some issues left to resolve) we can use those to invoke PROJ `cs2cs` functionality to transform from one coordinate reference system to another, else be informed that the transformation isn't valid or possible, which allows us to get the lat-lon coordinates for a given projection with corresponding CRS parameters.

CF-supported Grid Mappings have been modelled explicitly as objects in this first step in order to, for example:
* allow us to define all supported Grid Mappings to be a paradigm for those and the CF map parameter arguments listed there;
* allow us to interface with them using our own `cf.Units` instead of having to convert to and work with PROJ formats for inputs (which are strings with a unit as a suffix e.g. "30D", "0.5R", etc.), where functions to convert to and from PROJ data and units are provided;
* to facilitate future cf-plot work and integration into the cf-python codebase.

### Diagrams of new class hierarchy

TODO

### DIFF
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -323,6 +323,31 @@ from .read_write import read, write
 
 from .regrid import RegridOperator
 
+from .gridmappings import (
+    GridMapping,
+    AzimuthalGridMapping,
+    ConicGridMapping,
+    CylindricalGridMapping,
+    LatLonGridMapping,
+    PerspectiveGridMapping,
+    AlbersEqualArea,
+    AzimuthalEquidistant,
+    Geostationary,
+    LambertAzimuthalEqualArea,
+    LambertConformalConic,
+    LambertCylindricalEqualArea,
+    Mercator,
+    ObliqueMercator,
+    Orthographic,
+    PolarStereographic,
+    RotatedLatitudeLongitude,
+    LatitudeLongitude,
+    Sinusoidal,
+    Stereographic,
+    TransverseMercator,
+    VerticalPerspective,
+)
+
 
 # Set up basic logging for the full project with a root logger
 import logging
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -346,6 +346,8 @@ from .gridmappings import (
     Stereographic,
     TransverseMercator,
     VerticalPerspective,
+    _all_abstract_grid_mappings,
+    _all_concrete_grid_mappings,
 )
 
 
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -348,6 +348,7 @@ from .gridmappings import (
     VerticalPerspective,
     _all_abstract_grid_mappings,
     _all_concrete_grid_mappings,
+    _get_cf_grid_mapping_from_name,
 )
 
 
--- a/cf/__init__.py
+++ b/cf/__init__.py
@@ -346,10 +346,10 @@ from .gridmappings import (
     Stereographic,
     TransverseMercator,
     VerticalPerspective,
+    convert_proj_angular_data_to_cf,
     _all_abstract_grid_mappings,
     _all_concrete_grid_mappings,
     _get_cf_grid_mapping_from_name,
-    _convert_units_proj_to_cf,
 )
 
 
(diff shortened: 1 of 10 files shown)


sadielbartholomew:
Something has gone weird with all the commit timestamps, they are all pointing to today after my rebase. I might need to force push a fix. One moment...

sadielbartholomew:
And the order is messed up too :roll_eyes:

sadielbartholomew:
I'm going to wait an hour or so for my break whilst I see if it's a sync'ing thing and GitHub fixes it by then, if notk I might open a new PR since I'm worried something has gone badly wrong with the rebase...

sadielbartholomew:
OK timestamps are now all sensible and accurate (at least, under the 'Commits' pane now, not here on the discussion thread at least as far as I see now). Sorry about that - gits fault, not mine! The PR itself is all as intended, though I want to add a few updates here and there (some from part 2 that are better placed in part 1, and some docstring TODO filling out). Nothing that will change it in any way that makes it inaccurate to look at as-is, though!

I have went down late-night advanced git learning rabbit hole. I'm not sure why I haven't noticed and been annoyed by this before (probably I have, but have since forgotten), but when you do things such as a rebase, in my case to tidy some messy commits, some timestamps will be altered to the date of the rebase. Which can mess up the ordering and the dates shown for commits. Hence why `git log` was showing different stamps to what the listing here showed, which made me think something was dodgy with the code I'd pushed relative to the local copy before I'd tidied by rebase.

Turns out, if you use `git log --pretty=fuller` you can see the `AuthorDate` and `CommitDate` separately and when e.g. a rebase puts these out of sync you can use `git filter-branch -f --env-filter 'export GIT_COMMITTER_DATE="$GIT_AUTHOR_DATE"' <hash of commit previous to one of concern e.g. first in PR>..HEAD` to make those identical, and that fixed up the stamps here. I mention this all here partly as a reminder to myself, since I imagine I'll want to use this in future!

Anyway, all good now RE the commits here.

sadielbartholomew:
@davidhassell thanks for the very useful review discussion on Friday. See below for a brief summary of my understanding of the main points of the conversation (the side points are scribbled in my notes). Please let me know if that all sounds (about) right to you. I'll be updating the PR in line with the agreed changes (starting now, since I can always tweak further later if you do query something below).

### Summary

Note I'll tick these off with specific (sets of) commits, as I address them.

- [ ] **Design Pattern Change**: The current approach for creation of the Grid Mapping objects, for purposes mostly of validation of the map parameter inputs, is using an Abstract Factory design pattern, with use of the `ABC` metaclass and `abstractmethod` decorator, etc., for the base `GridMapping` class, but we want to convert to something more akin to using `__new__` as a factory method to generate the specific concrete classes, so using the Factory design pattern instead (as indicated by our whiteboard notes). Benefits being:
    - this avoids the clunky `_get_cf_grid_mapping_from_name` to match CF Grid Mapping class to a 'grid_mapping_name', doing this upon instance creation;
    - by avoiding `ABC` we can avoid use of a metaclass which would forbid use of our docstring re-writer, which would be really useful here given the amount of map parameter documenting required (the bulk of the lines changes here at present, probably) - it was on the TODO list before you pointed out it wouldn't be possible with the present approach.
    - nicer interface to the validation process, custom to each concrete GM, with a `validate` method.
- [ ] **Movement of code**: the `coordinatereference` module does not have access to the coordinates themselves, so don't touch that and instead move the main function for this feature, `create_2d_lats_and_lons`, to the `fielddomain` object.
- [ ] **House style**: since we tend to have only one class at most per module, for consistency move the separate concrete `GridMapping` classes out into their own appropriately-named modules under a `gridmappings` directory.
- [ ] **Special treatment of LatLon**: since this is our destination CRS usually (though eventually we can/will make this configurable), it can take no `__init__` and we can define a helper method `is_latlon_gm()` (or similar in name) to aid processing of any existing lat-lon coordinates, avoiding re-generation scenarios.

#### Notes

Including these for longevity and so I don't have to keep turning around to view them on my whiteboard :slightly_smiling_face:

davidhassell:
Hi Sadie - thanks for the summary - that's how I remember it, too!

PR #703, opened 2023-10-24
Testing and CI tidying including adding Python 3.12 job

sadielbartholomew:
Minor admin tweaks to:

* add some missing newer Python versions into the CI test workflow, including support for the new latest major version 3.12, which was released about 3 weeks ago which I tested locally to be compatible dependency-wise with cf and that all tests pass as they should with it (:heavy_check_mark:);
* add some test method skips where SciPy is required under-the-hood but not directly in the test, which produces a somewhat confusing error message as an outcome of `NameError: name 'convolve1d' is not defined. Did you mean: 'cf_convolve1d'?` because it can't find the `scipy` `convolve1d` and suggests our own wrapper;
* make two small formatting tweaks so that `test_style` i.e. `pycodestyle` passes.

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -60,7 +60,7 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         auto-update-conda: true
-        miniconda-version: 'latest'
+        miniconda-version: "latest"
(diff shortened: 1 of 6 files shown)


davidhassell (approved):
Thanks, Sadie - all these changes look very good to me. Please merge.

PR #710, opened 2024-02-07
Allow cell area calculation to ignore cell measure constructs

### LINKED ISSUE #709: Allow cell area calculation to ignore cell measure constructs
Sometimes there are cell measure constructs that are unusable, for instance if they are external and so contain no data - in which case cell area calculations with `cf.Field.cell_area` fail.

It would be good to provide an option to not use cell measure when this occurs.

PR to follow.

davidhassell:
FIxes #709

### DIFF
--- a/cf/field.py
+++ b/cf/field.py
@@ -3494,18 +3494,20 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
                 .. versionadded:: 3.3.0
 
             cell_measures: `bool`, optional
+
                 If True, the default, then area and volume cell
                 measure constructs are considered for weights creation
-                when *weights* of `True`, ``'area'``, or ``'volume'``,
-                Ignored for other values of *weights*.
+                when *weights* is `True`, ``'area'``, or
+                ``'volume'``. If False then cell measure constructs
+                are ignored for these *weights*.
 
                 .. versionadded:: 3.17.0
 
             coordinates: `bool`, optional
                 If True, the default, then coordinate constructs are
-                considered for weights creation when *weights* of
-                `True` or ``'area'``. Ignored for other values of
-                *weights*.
+                considered for weights creation for *weights* of
+                `True` or ``'area'``. If False then cell measure
+                constructs are ignored for these *weights*.
 
                 .. versionadded:: 3.17.0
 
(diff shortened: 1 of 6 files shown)


cf/field.py
```diff
@@ -2374,21 +2378,92 @@ def cell_area(
...
+        >>> print(f)
+        Field: specific_humidity (ncvar%q)
+        ----------------------------------
+        Data            : specific_humidity(latitude(5), longitude(8)) 1
+        Cell methods    : area: mean
+        Dimension coords: latitude(5) = [-75.0, ..., 75.0] degrees_north
+                        : longitude(8) = [22.5, ..., 337.5] degrees_east
+                        : time(1) = [2019-01-01 00:00:00]
+        Cell measures   : measure:area(latitude(5), longitude(8)) = [[4271287143027.272, ..., 4271287143027.272]] m2
+        >>> f.cell_area(methods=1)
```
sadielbartholomew, line 2465:
```suggestion
        >>> f.cell_area(methods=True)
```

1 is of course equivalent but could be confusing as an example.

cf/field.py
```diff
@@ -3381,6 +3493,23 @@ def weights(
...
+                when *weights* is `True`, ``'area'``, or
+                ``'volume'``. If False then cell measure constructs
+                are ignored for these *weights*.
+
+                .. versionadded:: 3.17.0
+
+            coordinates: `bool`, optional
+                If True, the default, then coordinate constructs are
+                considered for weights creation for *weights* of
+                `True` or ``'area'``. If False then cell measure
```
sadielbartholomew, line 3508:
```suggestion
                `True` or ``'area'``. If False then coordinate
```

cf/test/test_weights.py
```diff
@@ -291,6 +291,43 @@ def test_weights_line_area_ugrid(self):
         )
         self.assertEqual(w.Units, cf.Units("m"))
 
+    def test_weights_cell_measures_coordinates(self):
+        import cf
```
sadielbartholomew, line 295:
```suggestion
```

Since it is already imported. (With removal of the next line down, the 'suggestion' feature won't let me include that.)

sadielbartholomew (approved):
Adds the feature promised. Nice implementation. Minor comments raised, but overall this is all good, so please merge once those have been considered and included if you agree with my points. Thanks.

PR #711, opened 2024-02-07
Allow data instantiation from a `xarray.DataArray` object 

### LINKED ISSUE #706: Allow data instantiation from a `xarray.DataArray` object
At v3.16.0, a passing an `xarray.DataAarray` object in as the array of a `cf.Data` object works, insofar as it doesn't fail. However, because the xarray object claims (or might claim) to be a Dask collection  (i.e. `dask.base.is_dask_collection(<xarray.DataArray>)` is `True`), the xarray object could be stored internally in place of the expected Dask array.

The solution is to test for a `xarray.Datarray` inputs and convert it to a Dask array as appropriate.

Dask has some prior art on this, e.g. https://github.com/dask/dask/blob/2023.12.1/dask/array/core.py#L4569-L4570

davidhassell:
Fixes #706 

Uses the same technique as Dask.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -3,6 +3,8 @@ version 3.17.0
 
 **2024-??-??**
 
+* Allow `cf.Data` to be initialised with `xarray.DataAarray`
+  (https://github.com/NCAS-CMS/cf-python/issues/706)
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
Sensible approach. All good, please merge.

PR #714, opened 2024-02-19
Fix bug in `cf.Field.subspace` and `cf.Field.__getitem__` for some cyclic subspaces

### LINKED ISSUE #713: Bug in `cf.Field.subspace` and `cf.Field.__getitem__` for some cyclic subspaces
Between v3.14.0 and v3.16.0, there is a bug which incorrectly positions the data for cyclic subspaces that span the edges of the array - the coordinate metadata are correct but the data array is anchored on its origin left-hand edge, rather than the new origin.

For subspaces derived from 2-d coordinates (e.g. the latitude and longitude of tripolar ocean grids) thee is no problem.

For example:
```python
>>> import cf
>>> f = cf.example_field(0)
>>> g = f[:, -3:-5:1]
>>> print(f)
Cell methods    : area: mean
>>> print(g)
Data            : specific_humidity(latitude(5), longitude(6)) 1
Cell methods    : area: mean
                : longitude(6) = [-112.5, ..., 112.5] degrees_east
                : time(1) = [2019-01-01 00:00:00]g = f[:, -3:-5:1]
>>> f[:, :6].data.equals(g[:, :6].data, verbose=2)  # THIS SHOULD BE False
True
>>> f[:, -3:].data.equals(g[:, :3].data, verbose=2)  # THIS SHOULD BE True
False
```

davidhassell:
Fixes #713

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -15,6 +15,9 @@ version 3.17.0
 * Fix bug that caused `cf.Field.del_file_location` to fail when
   updating its metdata constructs
   (https://github.com/NCAS-CMS/cf-python/issues/707)
+* Fix bug that caused incorrect data arrays in some cyclic subspaces
+  created by `cf.Field.subspace` and `cf.Field.__getitem__`
+  (https://github.com/NCAS-CMS/cf-python/issues/713)
 
 version 3.16.0
 --------------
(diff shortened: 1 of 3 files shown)


sadielbartholomew (approved):
Good spot. Fixes the bug and captures it in a test; excellent, please merge.

PR #722, opened 2024-02-22
Fix bug when calculating UGRID cell areas when non-spatial coordinates span the discrete axis

### LINKED ISSUE #721: Fix bug when calculating UGRID cell areas when non-spatial coordinates span the discrete axis
When calculating  UGRID cell areas with non-spatial coordinates span that span the discrete axis, an `IndexError` sometimes occurs:

```python
IndexError: Invalid indices [0, 0, 0] for array with shape (13824, 4)

>>> print (f)
Field: air_pressure_at_mean_sea_level (ncvar%pmsl)
--------------------------------------------------
Data            : air_pressure_at_mean_sea_level(time(2), ncdim%ensemble(3), ncdim%nMesh2d_face(13824)) Pa
Cell methods    : time: mean (interval: 6 h)
Dimension coords: ncvar%ensemble(ncdim%ensemble(3)) = [0.0, 1.0, 2.0]
Auxiliary coords: longitude(ncdim%nMesh2d_face(13824)) = [70.3125, ..., 112.86804962158203] degrees_east
                : latitude(ncdim%nMesh2d_face(13824)) = [0.8827333450317383, ..., -47.359893798828125] degrees_north
                : time(time(2)) = [1978-09-01 12:00:00, 1978-09-02 12:00:00] 360_day
Topologies      : cell:face(ncdim%nMesh2d_face(13824), 4) = [[0, ..., 13755]]
>>> f.cell_area(great_circle=True)
Traceback
    ...
IndexError: Invalid indices [0, 0, 0] for array with shape (13824, 4)
```
Whether or not this happens depends on the arbitrary of the auxiliary coordinates stored inside the field (specifically within the embedded `Constructs` instance).

```python
Platform: Linux-5.15.0-94-generic-x86_64-with-glibc2.35
HDF5 library: 1.14.2
netcdf library: 4.9.2
esmpy/ESMF: 8.4.2
Python: 3.11.4
dask: 2023.12.1
netCDF4: 1.6.4
psutil: 5.9.5
packaging: 23.0
numpy: 1.25.2
scipy: 1.11.3
matplotlib: 3.8.0
cftime: 1.6.2
cfunits: 3.3.6
cfplot: 3.3.0
cfdm: 1.11.0.0
cf: 3.16.0
```

davidhassell:
Fixes #721 

No unit test - due to the lightly random nature, it's a bit tricky to write one. But the fix is pretty simple and it works for the use case described in the issue :)

### DIFF
--- a/cf/weights.py
+++ b/cf/weights.py
@@ -1652,7 +1652,7 @@ class Weights(Container, cfdm.Container):
         )
 
         for key, aux in auxiliary_coordinates_1d.items():
-            if aux.ctype not in "XYZ":
+            if str(aux.ctype) not in "XYZ":
                 continue
 
             aux_axis = f.get_data_axes(key)[0]
--- a/cf/weights.py
+++ b/cf/weights.py
@@ -1652,8 +1652,8 @@ class Weights(Container, cfdm.Container):
         )
 
         for key, aux in auxiliary_coordinates_1d.items():
-            #            if aux.ctype not in "XYZ":
-            #                continue
+            if aux.ctype not in "XYZ":
+                continue
 
             aux_axis = f.get_data_axes(key)[0]
 
--- a/cf/weights.py
+++ b/cf/weights.py
@@ -1652,6 +1652,9 @@ class Weights(Container, cfdm.Container):
         )
 
         for key, aux in auxiliary_coordinates_1d.items():
+            #            if aux.ctype not in "XYZ":
+            #                continue
+
             aux_axis = f.get_data_axes(key)[0]
 
             ugrid = f.domain_topology(default=None, filter_by_axis=(aux_axis,))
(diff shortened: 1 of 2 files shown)


cf/weights.py
```diff
@@ -1652,6 +1652,9 @@ def _geometry_ugrid_cells(cls, f, domain_axis, cell_type, auto=False):
         )
 
         for key, aux in auxiliary_coordinates_1d.items():
+            if aux.ctype not in "XYZ":
```
sadielbartholomew, line 1655: The `ctype` can emerge as `None`, e.g. from `type(cf.example_field(1).construct("auxiliarycoordinate2").ctype)`, which leads to two errors emerging in the test suite (`test_weights.py`) of:

```python
...
ERROR: test_weights_polygon_area_geometry (__main__.WeightsTest)
...
...
...
  File "/home/slb93/git-repos/cf-python/cf/weights.py", line 1655, in _geometry_ugrid_cells
    if aux.ctype not in "XYZ":
TypeError: 'in <string>' requires string as left operand, not NoneType

```

To avoid the need for type checking, I'd suggest this:

```suggestion
            if str(aux.ctype) not in "XYZ":
```

but any other way to handle the `None` cases is fine.

sadielbartholomew (changes requested):
One amendment required, but that should be all. Let me know what you think.

cf/weights.py
davidhassell: Yes indeed. I'm not sure why I didn't see this, I _thought_ I ran the tests ...

PR #730, opened 2024-02-29
For DSGs, don't attempt to aggregate non-feature axes

### LINKED ISSUE #723: Allow DSG tractories with identical `trajectory_id` values to be aggregated.
Currently you can't aggregate with `cf.aggregate` DSG trajectory fields with identical `cf_role=trajectory_id` coordinates, because of the general rule that when no 1-d coordinates for a given axis are different between fields, no aggregation can occur on that axis.

However, this is too restrictive for DSG trajectories, for which we may want to aggregate multiple realizations of the same trajectory path.

davidhassell:
Fixes #723, along with #724

Additional logic that didn't make it into #724


cf/aggregate.py
```diff
@@ -3505,10 +3514,15 @@ def _create_hash_and_first_values(
     # identity.
     canonical_direction = {}
 
+    aggregating_axes = list(axes)
```
sadielbartholomew, line 3517: If we've already specified this input should be a `list`, we don' t really need to do this conversion, though I appreciate this might be for safety in case of another sequence input. I'd personally remove it, but it is up to you:

```suggestion
    aggregating_axes = axes
```

sadielbartholomew (approved):
One minor comment, but all good so please merge once you've considered it (not a blocker by any means).

cf/aggregate.py
davidhassell: Good point. Contrary to what I said I thought was the case, it is much cleaner to just replace the `axes` input parameter with `aggregating_axes`. Done. 4f023d6063fedecfb824258a663df7cfac05a588

PR #738, opened 2024-03-13
Recognise LBPROC 131072 when reading UM files

### LINKED ISSUE #737: Recognise LBPROC 131072 (Mean over an ensemble of parallel runs) when reading UM files
Currently (v3.16.1) when LBPROC includes 131072 it is ignored by `cf.read`, whereas it should result in the creation of a `realization: mean` cell method.

davidhassell:
Fixes #737


sadielbartholomew (approved):
All good as far as I can tell. I'd ask for a corresponding test, but I suspect this one might be difficult to test on, so please merge if happy.

PR #742, opened 2024-03-20
Improve `cf.Field.collapse` performance by lazily computing reduced axis coordinates

### LINKED ISSUE #741: Improve `cf.Field.collapse` performance by lazily computing reduced axis coordinates
Currently (v3.16.1), collapsed coordinates are crated non-lazily. E.g in the following the collapsed size 1 coordinate values, and their bounds (latitude, longitude and time) are all computed non-lazily. This can be slow if the original coordinates are on disk, and _very_ slow if they are on disk on a remote server.

```python
>>> print(f)
Cell methods    : area: mean

>>> print(f.collapse('mean'))
Data            : specific_humidity(latitude(1), longitude(1)) 1
Cell methods    : area: mean latitude(1): longitude(1): mean
Dimension coords: latitude(1) = [0.0] degrees_north
                : longitude(1) = [180.0] degrees_east
```
It would be good compute these values from cached elements, if present, or else do it lazily so that the computation only occurs if the values are ever inspected.

davidhassell:
Fixes #741

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -9,6 +9,8 @@ version NEXT
   to regrid the vertical axis in logarithmic coordinates to
   `cf.Field.regrids` and `cf.Field.regridc`
   (https://github.com/NCAS-CMS/cf-python/issues/715)
+* Improve `cf.Field.collapse` performance by lazily computing reduced
+  axis coordinates (https://github.com/NCAS-CMS/cf-python/issues/741)
 * Fix misleading error message when it is not possible to create area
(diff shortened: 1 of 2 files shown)


sadielbartholomew (approved):
Seems logically sound and motivated. I have ran `test_collapse` (as well as the full test suite) both on this branch and on `main` and there is a significant speed improvement in each case, though I appreciate the main gains will be for more memory-intensive collapses.

All good, please merge.

PR #745, opened 2024-03-22
Improve `cf.Field.__getitem__` performance by not re-calculating axis cyclicity

### LINKED ISSUE #744: Improve `cf.Field.__getitem__` performance by not re-calculating axis cyclicity
Currently (v3.16.1), the cyclicty of all axes with dimension coordinates is checked during a Field subspace. This likely entails getting the first and last values if the coordinates, which might be very slow if the values are on (a remote) disk, or need computing.

However, the cyclicity is already known on the subspaced field's `Data` object (`cf.Data._getitem__` updates it), so we can just take it from there and disable the expensive automatic cyclicity check.

davidhassell:
Fixes #744

### DIFF
--- a/cf/field.py
+++ b/cf/field.py
@@ -444,12 +444,12 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
         # Record which axes were cyclic before the subspace
         org_cyclic = [data_axes.index(axis) for axis in new.cyclic()]
 
-        # Se the subspaced data
+        # Set the subspaced data
         new.set_data(new_data, axes=data_axes, copy=False)
 
         # Update axis cylcicity. Note that this can only entail
         # setting an originally cyclic axis to be non-cyclic. Doing
-        # this now it enables us to disable the (possibly very slow)
+        # this now enables us to disable the (possibly very slow)
         # automatic check for cyclicity on the 'set_construct' calls
         # below.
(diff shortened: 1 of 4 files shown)


cf/field.py
```diff
@@ -441,6 +441,23 @@ def __getitem__(self, indices):
...
+        # Update axis cylcicity. Note that this can only entail
+        # setting an originally cyclic axis to be non-cyclic. Doing
+        # this now it enables us to disable the (possibly very slow)
+        # automatic check for cyclicity on the 'set_construct' calls
+        # below.
+        if org_cyclic:
+            new_cyclic = new_data.cyclic()
+            for i in org_cyclic:
+                if i not in new_cyclic:
+                    new.cyclic(i, iscyclic=False)
```
sadielbartholomew, line 459: Could list comprehension it (though it might need some line-wrapping for PEP8/black, hard to tell line-length within the suggestion feature)?

```suggestion
            [new.cyclic(i, iscyclic=False) for i in org_cyclic if i not in new_cyclic]
                    
```

sadielbartholomew (approved):
All good, and well-motivated. I've raised some minor comments but please merge once you've considered those.

cf/field.py
davidhassell: Much nicer!

PR #747, opened 2024-03-23
Improve the performance of reading and accessing the data of PP and UM fields files

### LINKED ISSUE #746: Improve the performance of reading and accessing the data of PP and UM fields files
Aspects of accessing PP and UM fields file data has sometimes been very slow, for a quite a while. I had previously always assumed that this was a `cf.aggregation` issue, which it very much sometimes was! ... but I think aggregation now performs pretty well. 

@theabro kindly raised a case of reading a CF field from a 16 GB PP file. the CF Field itself comprised 2040 (= 24 x 85) 2-d PP fields:
```python
>>> print(f)
<CF Field: id%UM_m01s50i500_vn1300(time(24), atmosphere_hybrid_height_coordinate(85), latitude(144), longitude(192))>
```
Accessing the full data array  with `a = f.array` is taking **~11,000 seconds** - far too long!

Investigations showed that the reason for this was that the whole PP file was being parsed (i.e. all headers read and processed) for every 2-d PP field that contributes to the array, i.e. 2040 times in this case.

Stopping this parsing reduces the time taken to get the full array, on the same machine, to **~2 seconds** (!). The entire 16 GB can read from disk in ~3.5 minutes.

The size of the file per se is not the cause of the problem, rather the large amount of individual lookup headers in the file: 162,888 in this case. For small my test cases with fewer than 5 PP fields, the slow down is invisible :(

Long overdue PR to follow.

davidhassell:
Fixes #746

### DIFF
--- a/cf/data/array/umarray.py
+++ b/cf/data/array/umarray.py
@@ -8,7 +8,7 @@ from ...functions import (
     load_stash2standard_name,
     parse_indices,
 )
-from ...umread_lib.umfile import File
+from ...umread_lib.umfile import File, Rec
 from .abstract import Array
 from .mixin import FileArrayMixin
 
@@ -272,14 +272,12 @@ class UMArray(FileArrayMixin, cfdm.data.mixin.FileArrayMixin, Array):
                 The record container.
 
         """
-        # TODOCFA: This method doesn't require data_offset and disk_length,
-        # so plays nicely with CFA. Is it fast enough that we can
-        # use this method always?
+        return Rec.from_file_and_offsets(f, header_offset)
         for v in f.vars:
             for r in v.recs:
                 if r.hdr_offset == header_offset:
+                    print (r.__dict__)
                     return r
-
     def _set_units(self, int_hdr):
         """The units and calendar properties.
 
@@ -683,4 +681,5 @@ class UMArray(FileArrayMixin, cfdm.data.mixin.FileArrayMixin, Array):
             byte_ordering=self.get_byte_ordering(),
             word_size=self.get_word_size(),
             fmt=self.get_fmt(),
+            parse=False,
         )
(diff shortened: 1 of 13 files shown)


cf/read_write/um/umread.py
```diff
@@ -3567,6 +3590,7 @@ def file_open(self, filename):
             byte_ordering=g.get("byte_ordering"),
             word_size=g.get("word_size"),
             fmt=g.get("fmt"),
+            parse=parse,
```
sadielbartholomew, line 3593: I think it would be useful if, in this method or otherwise (but just above this call seems like a good place to do it), we add a `logging` call to highlight whether `parse` is enabled or disabled. Maybe even at warning level to warn when we are not parsing the contents. If you agree, please can you add in a brief line?

sadielbartholomew (approved):
Excellent stuff! I have checked timings with some hefty `.pp` files and array access is indeed much quicker. I've raised some minor comments about the code changes, but I'm happy overall so please merge once you feel ready.

cf/read_write/um/umread.py
sadielbartholomew: Sorry (as per discussion today) this is based upon a misunderstanding of mine - the parsing is not a user configuration choice, but purely under-the-hood. So ignore me here, marking as resolved.

cf/data/array/umarray.py
```diff
@@ -272,13 +272,22 @@ def _get_rec(self, f, header_offset):
...
+        return Rec.from_file_and_offsets(f, header_offset)
+
+        # ------------------------------------------------------------
+        # Leave the following commented code here for debugging
+        # purposes. Replacing the above line with this code moves the
+        # calculation of the data offset and disk length from pure
+        # Python to the C library, at the expense of completely
+        # parsing the file. Note: If you do replace the above line
+        # with the commented code, then you *must* also set
+        # 'parse=True' in the `open` method.
```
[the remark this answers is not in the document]
davidhassell: You've trained me well :)

PR #752, opened 2024-03-26
UNTESTED: do "make clean" in umlib build

alaniwi:
Removing the `.o` files for umlib will save a little bit of disk space.
I have not tested this patch, so you may want to change the base branch for merge to something other than main so that you can test it.


davidhassell:
Now that the reading of PP and UM datasets has moved to `umfive` (#961), we can close this, thanks :)

PR #753, opened 2024-03-27
Open and half-open intervals for `wi` queries

### LINKED ISSUE #740: Support (half)-open intervals for `wi` query endpoints
It would be useful (as identified today for work towards the VISION project, context summary below) to be able to query on, not just a *closed* interval for a 'within a range' (`wi`) query, i.e. inclusive of the given endpoints,  as is the only option at present, but an *open* (exclusive of the endpoints) and a *half-open* (inclusive on one side, for either side) condition.

To demonstrate, the behaviour of `wi` is: 

```python
>>> q = cf.wi(10, 20)
>>> q.evaluate(15)
True
>>> q.evaluate(10)
True
>>> q.evaluate(20)
True
```

but depending on context someone might want the evaluation results here to be `(T, F, F)`, `(T, T, F)`, or `(T, F, T)` instead. And equivalently for a without (`wo`) query.

This could be achieved with a compound query but it would be nicer to provide an interface to do this in one single query.

FYI the use case we had for our VISION project work, this would be especially handy for cases where querying is being done piecewise across a contiguous sequence, because with a closed interval there is *double counting* of the endpoints inside it.

### Implementation proposal

As discussed today, the simplest and tentatively agreed way to implement this is via a pair of new keywords which enable the start and end of the interval, the lower and upper bound, to become `open` instead of `closed`.  This would look to the user like so, where I am proposing the names `open_lower` and `open_upper` (but perhaps `open0` and `open1` could work also to match the `value0` and `value1` argument names) for those kwargs:

```python
>>> q = cf.wi(10, 20)
>>> q.evaluate(15)
True
>>> q.evaluate(10)
True
>>> q.evaluate(10, open_lower=True)
False
>>> q.evaluate(20)
True
>>> q.evaluate(20, open_upper=True)
False
```

~~These kwargs would also be added to `wo` for the equivalent (half-)open queries.~~ *(see below)*

#### Update: we've decided not to apply these to "wo" queries as well as "wi" queries, because in that context it would be quite confusing due to the inversion of the statement relative to "wi" - see https://github.com/NCAS-CMS/cf-python/pull/753#issue-2211307716.

sadielbartholomew:
Close  #740, where we have come to realise that it might not be a good idea to support the feature for `wo` queries since there is the potential for a lot of confusion, namely I started to document and implement the following behaviour:

> open_lower: `bool`, optional
>              If True, open the interval at the lower
>              bound so that value0 is excluded from the
>              range and therefore included in the set of
>              values outside of the range captured by "wo".
>              By default the interval is closed so that
>              value0 is included in the range and not
>              captured by "wo".

and @davidhassell and I decided it might be too confusing when applied to "outside a range" as opposed to "inside a range", where it is quite intuitive. So I will update the Issue description accordingly.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -24,6 +24,8 @@ version NEXT
 * Fix bug in `cf.read` when reading UM files that caused LBPROC value
   131072 (Mean over an ensemble of parallel runs) to be ignored
   (https://github.com/NCAS-CMS/cf-python/issues/737)
+* New keyword parameters to `cf.wi`: ``open_lower`` and ``open_upper``
+  (https://github.com/NCAS-CMS/cf-python/issues/740)
 
 ----
 
--- a/cf/query.py
+++ b/cf/query.py
@@ -207,7 +207,8 @@ class Query:
         exact=True,
         rtol=None,
         atol=None,
-        **kwargs,
+        open_lower=False,
+        open_upper=False,
     ):
         """**Initialisation**
 
@@ -250,6 +251,18 @@ class Query:
 
                 .. versionadded:: 3.15.2
 
+            open_lower: number, optional
+                Only applicable to the ``'wi'`` operator.
+                TODO
+
+                .. versionadded:: NEXTVERSION
+
+            open_upper: number, optional
+                Only applicable to the ``'wi'`` operator.
+                TODO
+
+                .. versionadded:: NEXTVERSION
+
             exact: deprecated at version 3.0.0.
                 Use `re.compile` objects in *value* instead.
 
@@ -290,13 +303,15 @@ class Query:
             self._rtol = rtol
             self._atol = atol
 
-        self._open_bounds = (False, False)
-        if kwargs:
-            if "open_bounds" in kwargs:
-                self._open_bounds = kwargs["open_bounds"]
-            else:
+        if open_lower or open_upper:
+            if operator != "wi":
                 raise ValueError(
-                    f"Unrecognised kwargs to Query object: {kwargs}")
+                    "Can only set the 'open_lower' and 'open_upper' "
+                    "parameters for the 'wi' operator"
+                )
+
+            self._open_lower = open_lower
+            self._open_upper = open_upper
 
     def __dask_tokenize__(self):
         """Return a hashable value fully representative of the object.
@@ -468,12 +483,10 @@ class Query:
         # for a closed interval, the default. For a (half-)open interval need
         # square bracket(s) -> parenthesis(/es), so unpack to adjust the repr.
         repr_value = str(self._value)
-        if True in self._open_bounds:  # that is, at least one side is open
-            open_lower, open_upper = self._open_bounds
-            if open_lower:
-                repr_value = "(" + repr_value[1:]
-            if open_upper:
-                repr_value = repr_value[:-1] + ")"
+        if self.open_lower:
+            repr_value = "(" + repr_value[1:]
+        if self.open_upper:
+            repr_value = repr_value[:-1] + ")"
 
         if not compound:
             out = f"{attr}({operator} {repr_value}"
@@ -618,6 +631,20 @@ class Query:
                 return units0
 
         raise AttributeError(f"{self!r} has indeterminate units")
+
+    @property
+    def open_lower(self):
+        """TODO
+
+        """
+        return getattr(self, "_open_lower", False)
+
+    @property
+    def open_upper(self):
+        """TODO
+
+        """
+        return getattr(self, "_open_upper", False)
 
     @property
     def rtol(self):
@@ -928,13 +955,12 @@ class Query:
             if _wi is not None:
                 return _wi(value)
 
-            open_lower, open_upper = self._open_bounds
-            if open_lower:
+            if self.open_lower:
                 lower_bound = x > value[0]
             else:
                 lower_bound = x >= value[0]
 
-            if open_upper:
+            if self.open_upper:
                 upper_bound = x < value[1]
             else:
                 upper_bound = x <= value[1]
@@ -1753,7 +1779,7 @@ def wi(
     """
     return Query(
         "wi", [value0, value1], units=units, attr=attr,
-        open_bounds=(open_lower, open_upper),
+        open_lower=open_lower, open_upper=open_upper,
     )
 
 
--- a/cf/query.py
+++ b/cf/query.py
@@ -251,15 +251,21 @@ class Query:
 
                 .. versionadded:: 3.15.2
 
-            open_lower: number, optional
+            open_lower: `bool`, optional
                 Only applicable to the ``'wi'`` operator.
-                TODO
+                If True, open the interval at the lower
+                bound so that value0 is excluded from the
+                range. By default the interval is closed
+                so that value0 is included.
 
                 .. versionadded:: NEXTVERSION
 
-            open_upper: number, optional
+            open_upper: `bool`, optional
                 Only applicable to the ``'wi'`` operator.
-                TODO
+                If True, open the interval at the upper
+                bound so that value1 is excluded from the
+                range. By default the interval is closed
+                so that value1 is included.
 
                 .. versionadded:: NEXTVERSION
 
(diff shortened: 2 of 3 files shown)


sadielbartholomew:
Sorry David I forgot to tag you as requested reviewer last week. All good to review (I resolved one trivial merge conflict in the Changelog for that last merge-based commit).

cf/query.py
```diff
@@ -207,6 +207,7 @@ def __init__(
         exact=True,
         rtol=None,
         atol=None,
+        **kwargs,
```
davidhassell, line 210: I think that it's better to have two new keyword arguments Boolean `open_lower` and `open_lower` (or perhaps one new argument `open` that is a tuple of 2 bools?) than to use unspecified `**kwargs`. It's important that different methods share the same API - whatever we use for `cf.wi` should be the same for `cf.Query`.

cf/query.py
```diff
@@ -905,7 +914,16 @@ def _evaluate(self, x, parent_attr):
...
+            open_lower, open_upper = self._open_bounds
+            if open_lower:
+                lower_bound = (x > value[0])
+            else:
+                lower_bound = (x >= value[0])
+            if open_upper:
+                upper_bound = (x < value[1])
+            else:
+                upper_bound = (x <= value[1])
+            return lower_bound & upper_bound
```
davidhassell, line 926: Just some house style stuff:
```suggestion
            if open_lower:
                lower_bound = x > value[0]
            else:
                lower_bound = x >= value[0]

            if open_upper:
                upper_bound = x < value[1]
            else:
                upper_bound = x <= value[1]

            return lower_bound & upper_bound
```

cf/query.py
```diff
@@ -1671,9 +1709,20 @@ def wi(value0, value1, units=None, attr=None):
     True
     >>> q.evaluate(4)
     False
+    >>> q.evaluate(5)
+    True
+    >>> q.evaluate(5, open_lower=True)
```
davidhassell, line 1714: This example doesn't look right - it's `cf.wi` that has the `open_lower` keyword, not `evaluate`.

cf/query.py
```diff
@@ -1671,9 +1709,20 @@ def wi(value0, value1, units=None, attr=None):
...
+    False
+    >>> q.evaluate(7)
+    True
+    >>> q.evaluate(7, open_upper=True)
```
davidhassell, line 1718: This example doesn't look right - it's `cf.wi` that has the `open_lower` keyword, not `evaluate`.

cf/query.py
```diff
@@ -1671,9 +1709,20 @@ def wi(value0, value1, units=None, attr=None):
...
+    False
 
     """
-    return Query("wi", [value0, value1], units=units, attr=attr)
+    return Query(
+        "wi", [value0, value1], units=units, attr=attr,
+        open_bounds=(open_lower, open_upper),
```
davidhassell, line 1724: See above comments on API

cf/query.py
```diff
@@ -1629,9 +1647,17 @@ def isclose(value, units=None, attr=None, rtol=None, atol=None):
     )
 
 
-def wi(value0, value1, units=None, attr=None):
+def wi(
+        value0, value1, open_lower=False, open_upper=False,
```
davidhassell, line 1651: See above comments on API

Changelog.rst
```diff
@@ -24,6 +24,8 @@ version 3.16.2
 * Fix bug in `cf.read` when reading UM files that caused LBPROC value
   131072 (Mean over an ensemble of parallel runs) to be ignored
   (https://github.com/NCAS-CMS/cf-python/issues/737)
+* New keyword parameters to `cf.wi`: ``open_lower`` and ``open_upper``
```
davidhassell, line 27: See `query.py` comments on API

cf/query.py
sadielbartholomew: Fair enough. Happy to change accordingly. Will wait until I get to the office so we can talk about the API in-person, though, before making any changes.

cf/query.py
sadielbartholomew: OK, good spot. I will have another look.

sadielbartholomew:
Thanks David for your useful feedback and offline discussions on this from last week. Updating my local branch now and I'll push everything up when ready for the re-review.

Changelog.rst
sadielbartholomew: Not sure what there is to be done here, given your feedback as registered (on the PR and in person). Please let me know if you had anything specific in mind to tweak or add, here.

cf/test/test_Query.py
```diff
@@ -47,6 +50,13 @@ def test_Query(self):
         str(u)
         u.dump(display=False)
 
+        # For "wi", check repr. provides correct notation for open/closed-ness
+        # of the interval captured.
+        self.assertEqual(repr(q), "<CF Query: (wi [2, 5])>")
+        self.assertEqual(repr(v), "<CF Query: (wi (2, 5])>")
+        self.assertEqual(repr(w), "<CF Query: (wi [2, 5))>")
+        self.assertEqual(repr(x), "<CF Query: (wi (2, 5))>")
```
sadielbartholomew, line 58: When updating the PR I was thinking I'd consider it an improvement if we add a space after the stated interval, to make this more readable and highlight the bracket/parenthesis notation, e.g. instead of `<CF Query: (wi [2, 5])>` we would have `<CF Query: (wi [2, 5] )>` (or maybe `<CF Query: ( wi [2, 5] )>` for symmetry?) and so on.

Not sure if you agree? Even if you do, I don't mind leaving it out of this PR though, since the release is imminent and that is a new commit I'd need to update the PR with.

sadielbartholomew:
I have just pushed my new commits which I believe address all of your feedback, thanks @davidhassell. All ready to (rock and) roll with a re-review.

cf/query.py
```diff
@@ -315,6 +345,8 @@ def __dask_tokenize__(self):
         operator = self._operator
         if operator == "isclose":
             value += (self.rtol, self.atol)
+        if operator == "wi":
```
davidhassell, line 348: Just a bit of house style - we generally have a  blank line after clauses:
```suggestion

        if operator == "wi":
```

cf/query.py
```diff
@@ -452,8 +484,20 @@ def __str__(self):
...
+
+        # For "wi" queries only, open intervals are supported. For "wi" _value
+        # is a list of two values, with representation from string list form
+        # of '[a, b]' which corresponds to the standard mathematical notation
+        # for a closed interval, the default. But an open endpoint is indicated
+        # by a parenthesis, so adjust repr. to convert square bracket(s).
+        repr_value = str(self._value)
+        if self.open_lower:
+            repr_value = "(" + repr_value[1:]
+        if self.open_upper:
```
davidhassell, line 496:
```suggestion

        if self.open_upper:
```

davidhassell, line 496:
```suggestion

        if self.open_upper:
```

cf/query.py
```diff
@@ -1643,6 +1733,18 @@ def wi(value0, value1, units=None, attr=None):
         value1:
              The upper bound of the range.
 
+        open_lower: `bool`, optional
+             If True, open the interval at the lower
+             bound so that value0 is excluded from the
+             range. By default the interval is closed
+             so that value0 is included.
```
davidhassell, line 1740:
```suggestion
             so that value0 is included.
             
             .. versionadded:: NEXTVERSION
```

cf/query.py
```diff
@@ -1643,6 +1733,18 @@ def wi(value0, value1, units=None, attr=None):
...
+
+        open_upper: `bool`, optional
+             If True, open the interval at the upper
+             bound so that value1 is excluded from the
+             range. By default the interval is closed
+             so that value1 is included.
```
davidhassell, line 1746:
```suggestion
             so that value1 is included.
             
             .. versionadded:: NEXTVERSION
```

Changelog.rst
davidhassell: All good.

cf/query.py
```diff
@@ -645,7 +711,7 @@ def value(self):
 
     def addattr(self, attr):
         """Return a `Query` object with a new left hand side operand
-        attribute to be used during evaluation. TODO.
+        attribute to be used during evaluation.
```
davidhassell, line 714: I think that this TODO was to probably to make this fit on one line (as PEP8 suggests). How about:
```
        """Redefine the query to be on an object's attribute.
```

davidhassell (approved):
Thanks, Sadie. This looks great, and there's some excellent testing :). I've made some minor style and docstring suggestions. Please merge when when you have considered them.

cf/query.py
sadielbartholomew: I see. I just thought it would show up in the docs and seem ugly so I blitzed it.

Great, that is always good practice. Nice suggestion, I will include it in a new commit to push up.

sadielbartholomew: Done in 030e5cf5df48a5bca0cb240d139206f405fc329c. Marking as resolved.

PR #763, opened 2024-04-22
Override `del_construct` to update cyclic axes set when due

### LINKED ISSUE #758: Post-`regrids` `Field.cyclic()` retains removed domain axes
On the current `main` branch, a Field after a `regrids` can return with `Field.cyclic()` a (cyclic) domain axes which no longer exists on the Field because it has been removed via the regridding operation. I believe this is the underlying reason that we see the issue of #756, though I think adding in a tweak to the logic there as suggested in #757 makes it more robust, so think it would be good to add regardless and will leave that PR open for consideration.

We may want to prioritise a fix for this to put in before our imminent release, because it is probably a issue with the new feature of "Added ... the option to regrid the vertical axis in logarithmic coordinates to cf.Field.regrids and cf.Field.regridc " (since it shows up in a vertical regrid, as per the details below).

### Example case and behaviour

I've spent a lot of time trying to get a minimal (or at least, pared down) working example of this to share here, but I am struggling to find something so have nearly given up on a small reproducer, in favour of moving onto getting a fix since I feel like I have isolated the problem enough to do that now.

But, for detail and context, the issue has emerged as part of the breaking of our VISION end-to-end script recently as noticed during https://github.com/NCAS-CMS/cf-python/issues/756 (see the report there for the underlying error emerging). So I'll share relevant details of that here now which can hopefully shine sufficient light on the problem, since I am unsure whether I'll have time before going on leave after tomorrow to get the reduced example up.

Before the regridding we have a field `a` with `print(a, a.constructs(), a.cyclic())` being (I added spaces between lines for clarity):

```python
Field: id%UM_m01s51i010_vn1105 (ncvar%UM_m01s51i010_vn1105)
-----------------------------------------------------------
Data            : id%UM_m01s51i010_vn1105(time(5), air_pressure(19), latitude(144), longitude(192)) 1
Cell methods    : time(5): point
Dimension coords: time(5) = [2017-07-03 11:00:00, ..., 2017-07-03 15:00:00] gregorian
                : air_pressure(19) = [1000.0, ..., 100.0] hPa
                : latitude(144) = [-89.375, ..., 89.375] degrees_north
                : longitude(192) = [0.9375, ..., 359.0625] degrees_east

Constructs:
{'cellmethod0': <CF CellMethod: domainaxis0: point>,
 'dimensioncoordinate0': <CF DimensionCoordinate: time(5) days since 2017-1-1 gregorian>,
 'dimensioncoordinate1': <CF DimensionCoordinate: air_pressure(19) hPa>,
 'dimensioncoordinate2': <CF DimensionCoordinate: latitude(144) degrees_north>,
 'dimensioncoordinate3': <CF DimensionCoordinate: longitude(192) degrees_east>,
 'domainaxis0': <CF DomainAxis: size(5)>,
 'domainaxis1': <CF DomainAxis: size(19)>,
 'domainaxis2': <CF DomainAxis: size(144)>,
 'domainaxis3': <CF DomainAxis: size(192)>}

{'domainaxis3'}
```

and with a field `b` of:

```python
Field: mole_fraction_of_ozone_in_air (ncvar%O3_TECO)
----------------------------------------------------
Data            : mole_fraction_of_ozone_in_air(ncdim%obs(11160)) ppb
Auxiliary coords: time(ncdim%obs(11160)) = [2017-07-03 11:15:07, ..., 2017-07-03 14:21:06] standard
                : altitude(ncdim%obs(11160)) = [2577.927001953125, ..., 151.16905212402344] m
                : air_pressure(ncdim%obs(11160)) = [751.6758422851562, ..., 1006.53076171875] hPa
                : latitude(ncdim%obs(11160)) = [52.56147766113281, ..., 52.0729866027832] degree_north
                : longitude(ncdim%obs(11160)) = [0.3171832859516144, ..., -0.6249311566352844] degree_east
                : cf_role=trajectory_id(cf_role=trajectory_id(1)) = [STANCO]
```

after an operation of:
```python
c = a.regrids(b, method="linear", z="air_pressure", ln_z=True,)
```

we get, for `print(c, c.constructs(), c.cyclic())`:

```python

AFTER Field: id%UM_m01s51i010_vn1105 (ncvar%UM_m01s51i010_vn1105)
-----------------------------------------------------------
Data            : id%UM_m01s51i010_vn1105(time(5), ncdim%obs(11160)) 1
Cell methods    : time(5): point
Dimension coords: time(5) = [2017-07-03 11:00:00, ..., 2017-07-03 15:00:00] gregorian
Auxiliary coords: time(ncdim%obs(11160)) = [2017-07-03 11:15:07, ..., 2017-07-03 14:21:06] standard
                : altitude(ncdim%obs(11160)) = [2577.927001953125, ..., 151.16905212402344] m
                : air_pressure(ncdim%obs(11160)) = [751.6758422851562, ..., 1006.53076171875] hPa
                : latitude(ncdim%obs(11160)) = [52.56147766113281, ..., 52.0729866027832] degree_north
                : longitude(ncdim%obs(11160)) = [0.3171832859516144, ..., -0.6249311566352844] degree_east

Constructs:
{'auxiliarycoordinate0': <CF AuxiliaryCoordinate: time(11160) days since 1900-01-01 00:00:00 standard>,
 'auxiliarycoordinate1': <CF AuxiliaryCoordinate: altitude(11160) m>,
 'auxiliarycoordinate2': <CF AuxiliaryCoordinate: air_pressure(11160) hPa>,
 'auxiliarycoordinate3': <CF AuxiliaryCoordinate: latitude(11160) degree_north>,
 'auxiliarycoordinate4': <CF AuxiliaryCoordinate: longitude(11160) degree_east>,
 'cellmethod0': <CF CellMethod: domainaxis0: point>,
 'dimensioncoordinate0': <CF DimensionCoordinate: time(5) days since 2017-1-1 gregorian>,
 'domainaxis0': <CF DomainAxis: size(5)>,
 'domainaxis4': <CF DomainAxis: size(11160)>}

{'domainaxis3'}
````

**Notice how `'domainaxis3'` is no longer a construct on the Field `c`, but it remains listed in `c.cyclic()`, when it should have been removed.**

### PR DESCRIPTION
**Pull request by sadielbartholomew**
Fix #758 and #756 (the former being the underlying issue and the latter a specific bug as a downstream effect) by overriding the `del_construct` inherited from cfdm to account for and update the store of `cyclic` axes. Previously we used the inherited method as-is so cyclic axes weren't updated and could get out of sync, notably contain axes which no longer existed due to deletion.

Note:
* The test `test_Domain_del_construct` added here will currently fail halfway-through due to #762, as noted in the test. I am fixing that next.
* I will add a note of two to the code in anticipation of some comments in review, for discussion.

### DIFF
--- a/cf/mixin/fielddomain.py
+++ b/cf/mixin/fielddomain.py
@@ -2418,7 +2418,7 @@ class FieldDomain:
         removed even if it is referenced by coordinate reference
         construct.
 
-        .. versionadded:: 3.16.2
+        .. versionadded:: NEXTVERSION
 
         .. seealso:: `get_construct`, `constructs`, `has_construct`,
                      `set_construct`
@@ -2445,7 +2445,7 @@ class FieldDomain:
 
             {{filter_kwargs: optional}}
 
-                .. versionadded:: (cfdm) 1.8.9.0
+                .. versionadded:: NEXTVERSION
 
         :Returns:
 
--- a/cf/mixin/fielddomain.py
+++ b/cf/mixin/fielddomain.py
@@ -2484,10 +2484,9 @@ class FieldDomain:
         # have the concept of cyclic axes, so have to update the
         # register of cyclic axes when we delete a construct in cf.
 
-        # Get the relevant key first because it will be lost upon deletion.
+        # Get the relevant key first because it will be lost upon deletion
         key = self.construct_key(*identity, default=None, **filter_kwargs)
-        # Copy since should never change value of _cyclic attribute in-place
-        cyclic_axes = self._cyclic.copy()
+        cyclic_axes = self._cyclic
 
         deld_construct = super().del_construct(
             *identity, default=None, **filter_kwargs
@@ -2495,21 +2494,20 @@ class FieldDomain:
         if deld_construct is None:
             if default is None:
                 return
-                
-            return self._default(default, "Can't find unique construct to remove")
+
+            return self._default(
+                default, "Can't find unique construct to remove"
+            )
 
         # If the construct deleted was a cyclic axes, remove it from the set
-        # of stored cyclic axes, to update that appropriately. Do this
-        # afterwards because the deletion might not be successful and don't
-        # want to update the cyclic() set unless we know the deletion occurred.
+        # of stored cyclic axes, to sync that. This is safe now, since given
+        # the block above we can be sure the deletion was successful.
         if key in cyclic_axes:
-            # The below is to test that the construct was successfully deleted
-            if isinstance(default, Exception) or (
-                not isinstance(default, Exception)
-                and deld_construct != default
-            ):
-                cyclic_axes.remove(key)
-                self._cyclic = cyclic_axes
+            # Never change value of _cyclic attribute in-place. Only copy now
+            # when the copy is known to be required.
+            cyclic_axes = cyclic_axes.copy()
+            cyclic_axes.remove(key)
+            self._cyclic = cyclic_axes
 
         return deld_construct
 
--- a/cf/mixin/fielddomain.py
+++ b/cf/mixin/fielddomain.py
@@ -2490,8 +2490,13 @@ class FieldDomain:
         cyclic_axes = self._cyclic.copy()
 
         deld_construct = super().del_construct(
-            *identity, default=default, **filter_kwargs
+            *identity, default=None, **filter_kwargs
         )
+        if deld_construct is None:
+            if default is None:
+                return
+                
+            return self._default(default, "Can't find unique construct to remove")
 
         # If the construct deleted was a cyclic axes, remove it from the set
         # of stored cyclic axes, to update that appropriately. Do this
--- a/cf/mixin/fielddomain.py
+++ b/cf/mixin/fielddomain.py
@@ -2509,6 +2509,8 @@ class FieldDomain:
                     f"({key}), so it was removed from the cyclic() axes set."
                 )
 
+        return deld_construct
+
     def set_coordinate_reference(
         self, coordinate_reference, key=None, parent=None, strict=True
     ):
--- a/cf/mixin/fielddomain.py
+++ b/cf/mixin/fielddomain.py
@@ -2490,7 +2490,8 @@ class FieldDomain:
         cyclic_axes = self._cyclic.copy()
 
         deld_construct = super().del_construct(
-            *identity, default=default, **filter_kwargs)
+            *identity, default=default, **filter_kwargs
+        )
 
         # If the construct deleted was a cyclic axes, remove it from the set
         # of stored cyclic axes, to update that appropriately. Do this
@@ -2499,13 +2500,13 @@ class FieldDomain:
         if key in cyclic_axes:
             # The below is to test that the construct was successfully deleted
             if isinstance(default, Exception) or (
-                    not isinstance(default, Exception)
-                    and deld_construct != default
+                not isinstance(default, Exception)
+                and deld_construct != default
             ):
                 cyclic_axes.remove(key)
                 self._cyclic = cyclic_axes
                 logger.info(
-                "Deleted a construct that corresponds to a cyclic axis "
+                    "Deleted a construct that corresponds to a cyclic axis "
                     f"({key}), so it was removed from the cyclic() axes set."
                 )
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by sadielbartholomew** [comment]
Just going to add one commit to fix a few style/formatting issues detected by the CI linting.

File: cf/mixin/fielddomain.py [hunk 1]
```diff
@@ -2408,6 +2408,110 @@ def set_construct(
...
+        # If the construct deleted was a cyclic axes, remove it from the set
+        # of stored cyclic axes, to update that appropriately. Do this
+        # afterwards because the deletion might not be successful and don't
+        # want to update the cyclic() set unless we know the deletion occurred.
+        if key in cyclic_axes:
+            # The below is to test that the construct was successfully deleted
+            if isinstance(default, Exception) or (
+                not isinstance(default, Exception)
+                and deld_construct != default
+            ):
```
sadielbartholomew, line 2505: FYI (reviewing), I considered using `has_construct` to test whether the deletion occurred successfully, but thought it would only test the construct was not there which may have the potential to hide future bugs, whereas this logic will test that the `super().del_construct()` call directly above did actually manage to delete a construct that was previously there. I do accept it isn't ideal because it is a bit of a clunky way to check. It left me wishing the signature return included a Boolean to confirm whether or not the deletion happened, explicitly, but I think that would complicate the signature and output too much.

Not sure what you think, David - is this best or should I use `has_construct` or some other means, to check whether the `del_construct` operation above worked?

sadielbartholomew (commented):
@davidhassell one note to consider for your review. Thanks.

cf/mixin/fielddomain.py
davidhassell: Hi Sadie, we usually deal with this by setting the default to `None` and then handling that easily testable case (see above)

cf/mixin/fielddomain.py
```diff
@@ -2408,6 +2408,110 @@ def set_construct(
...
+        # register of cyclic axes when we delete a construct in cf.
+
+        # Get the relevant key first because it will be lost upon deletion.
+        key = self.construct_key(*identity, default=None, **filter_kwargs)
+        # Copy since should never change value of _cyclic attribute in-place
+        cyclic_axes = self._cyclic.copy()
+
+        deld_construct = super().del_construct(
+            *identity, default=default, **filter_kwargs
+        )
```
davidhassell, line 2494: See comment below on handling this case:
```suggestion
        deld_construct = super().del_construct(
            *identity, default=None, **filter_kwargs
        )
        if deld_construct is None:
            if default is None:
                return
                
            return self._default(default, "Can't find unique construct to remove")
```
This safer and faster, in general.

davidhassell (changes requested):
Hi Sadie - look good - I've made some suggestions on handling the construct-doesn't exist case.

cf/mixin/fielddomain.py
```diff
@@ -2408,6 +2408,110 @@ def set_construct(
...
+            if isinstance(default, Exception) or (
+                not isinstance(default, Exception)
+                and deld_construct != default
+            ):
+                cyclic_axes.remove(key)
+                self._cyclic = cyclic_axes
+                logger.info(
+                    "Deleted a construct that corresponds to a cyclic axis "
+                    f"({key}), so it was removed from the cyclic() axes set."
+                )
```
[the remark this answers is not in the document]
sadielbartholomew: Ultimately, no. I do think it would be nice as verbose information to at least add "Deleted a construct that corresponds to a cyclic axis." i.e. the first line, but potentially as a `logger.detail` or even `debug` level if you think it is a bit much to tell the user on `info` level?

cf/mixin/fielddomain.py
sadielbartholomew: Aha, thanks David. Good to know we have a better way to do it than those two approaches I thought of (in https://github.com/NCAS-CMS/cf-python/pull/763#discussion_r1574530753), both of which I wasn't keen on!

cf/mixin/fielddomain.py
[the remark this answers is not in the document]
davidhassell: Hi Sadie - I'm afraid that I think it's too much at any level :) To me this is just basic functionality that we shouldn't make slower with logging calls.

sadielbartholomew:
All ready for your re-review, @davidhassell. Thanks.

davidhassell (approved):
Look great. Thanks!

sadielbartholomew:
Thanks David. In that case, merging!

Issue #766, opened 2024-04-24
Addition of return_esmpy_regrid_operator to regrids and regridc?


mattjbr123:
Hiya,
I've noticed that between v3.13.1 and v3.14.0, when the regridding code was refactored as part of the dask implentation (#438), the behaviour of the `return_operator `kwarg to regrids and regridc changed. In v3.13.1 the `cf.RegridOperator` it returned could be used to [access the `ESMF.Regrid` object via the regrid attribute](https://github.com/NCAS-CMS/cf-python/blob/cc4d7dfa985e800f5e6df68db7b4b5980216e2d6/cf/regrid/regridoperator.py#L136). In v3.14.0 and beyond this is no longer the case, but a `return_esmpy_regrid_operator` kwarg has been added to the [new regrid worker function](https://github.com/NCAS-CMS/cf-python/blob/6967d544c83ff247b2f6e97497c13f1ddb7159d3/cf/regrid/regrid.py#L81) which allows access to the `ESMF.Regrid` object. However this is not a kwarg available in [regrids ](https://github.com/NCAS-CMS/cf-python/blob/6967d544c83ff247b2f6e97497c13f1ddb7159d3/cf/field.py#L13858)and [regridc](https://github.com/NCAS-CMS/cf-python/blob/6967d544c83ff247b2f6e97497c13f1ddb7159d3/cf/field.py#L14135), which call `regrid`, so I would have to call `regrid `directly to access the `ESMF.Regrid` object, which is probably not recommended/ideal etc. 

To my relatively untrained eye it looks like there should not be a problem adding the `return_esmpy_regrid_operator `kwarg to `regrids `and `regridc `and passing it to `regrid`, and am happy to put together a very simple PR to add this in if so, but you might be able to advise better!

More details on the issue this is causing here: https://github.com/unifhy-org/unifhy/issues/98

davidhassell:
Hi Matt,

Thanks for this, and sorry that we confounded things for you! I'm sure that we can put this back, and we'd welcome a  PR - thank you.

If you're _really, really_ quick, it _just might (no promises)_ get in to `3.16.2`, due to be released later this week.

All the best,
David

PR #767, opened 2024-04-24
Add return_esmpy_regrid_operator kwarg to regrids and regridc per #766

mattjbr123:
For #766
Added return_esmpy_regrid_operator kwargs to field/regrids and field/regridc, so that users wanting to make use of the functionality offered by it can do so via regrids and regridc, without having to call the regrid worker function directly.

This is possibly my 2nd ever PR, so if I've done something stupid, please forgive me... xD


mattjbr123:
Humm, I think I messed something up by copying files around outside of the repo, I'll close this and try again...

PR #768, opened 2024-04-24
Added return_esmpy_regrid_operator to regrids and regridc per #766

mattjbr123:
That looks much better. #766 

Added return_esmpy_regrid_operator kwargs to field/regrids and field/regridc, so that users wanting to make use of the functionality offered by it can do so via regrids and regridc, without having to call the regrid worker function directly.

This is possibly my 2nd ever PR, so if I've done something stupid, please forgive me... xD
For example, see #767

### DIFF
--- a/cf/field.py
+++ b/cf/field.py
@@ -13491,6 +13491,7 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
         inplace=False,
         i=False,
         _compute_field_mass=None,
+        return_esmpy_regrid_operator=False,
     ):
         """Regrid the field to a new latitude and longitude grid.
 
@@ -13727,6 +13728,10 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
                 .. versionadded:: 3.16.0
 
             {{inplace: `bool`, optional}}
+            
+            {{return_esmpy_regrid_operator: `bool`, optional}}
+
+                .. versionadded:: 3.16.2            
 
             axis_order: sequence, optional
                 Deprecated at version 3.14.0.
@@ -13744,7 +13749,8 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
             `Field` or `None` or `RegridOperator`
                 The regridded field construct; or `None` if the
                 operation was in-place; or the regridding operator if
-                *return_operator* is True.
+                *return_operator* is True; or the esmpy.Regrid operator
+                object if *return_esmpy_regrid_operator* is True.
 
         **Examples**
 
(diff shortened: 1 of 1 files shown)


cf/field.py
```diff
@@ -13744,7 +13749,8 @@ def regrids(
             `Field` or `None` or `RegridOperator`
                 The regridded field construct; or `None` if the
                 operation was in-place; or the regridding operator if
-                *return_operator* is True.
+                *return_operator* is True; or the esmpy.Regrid operator
```
davidhassell, line 13752:
```suggestion
                *return_operator* is True; or the `esmpy.Regrid` operator
```

cf/field.py
```diff
@@ -14033,7 +14045,8 @@ def regridc(
...
```
davidhassell, line 14048:
```suggestion
                *return_operator* is True; or the `esmpy.Regrid` operator
```

cf/field.py
```diff
@@ -13491,6 +13491,7 @@ def regrids(
         inplace=False,
         i=False,
         _compute_field_mass=None,
+        return_esmpy_regrid_operator=False,
```
davidhassell, line 13494: Could this be moved to just before the `inplace` keyword (just some secret house style)

cf/field.py
```diff
@@ -13845,6 +13852,7 @@ def regridc(
...
```
davidhassell, line 13855: Could this be moved to just before the  the `inplace` keyword (just some secret house style)

davidhassell (changes requested):
Hi Matt - looks good - thanks for the quick turn around!

I've made a few style suggestions.

Also we need `{{return_esmpy_regrid_operator: bool, optional}}` to be fleshed out in `cf/docstring/docstring.py`, after [this line](https://github.com/NCAS-CMS/cf-python/blob/main/cf/docstring/docstring.py#L653) would be a good spot - hopefully the formatting is clear enough ...

And we need a new test method in `cf/test/test_regrid.py` - just one that sets `return_esmpy_regrid_operator=True` in both `regrids` and `regridc`, and checks the type (`self.assertIsInstance`) of the returned object should suffice.

There's still time :)

cf/field.py
mattjbr123: Resolved in commit https://github.com/NCAS-CMS/cf-python/pull/768/commits/40b00233a50df02a3a728cbbb59c7526e31e8c9f

cf/field.py
mattjbr123: Resolved in commit https://github.com/NCAS-CMS/cf-python/pull/768/commits/40b00233a50df02a3a728cbbb59c7526e31e8c9f

mattjbr123:
How's this?
https://github.com/NCAS-CMS/cf-python/pull/768/commits/f9bb52f23503c24dc0b350a575d8e69a66912b70

davidhassell (approved):
Nicely done, Matt - thanks. I took it upon myself to do a bit of tidying, and shall merge it. You made it in to 3.16.2!

PR #769, opened 2024-04-26
Override `cfdm.Field.get_domain` to store axes cyclicity

### LINKED ISSUE #762: `cyclic`: different sets for `Field.domain` & corresponding `Field`
During work towards a fix for #758 I have noticed that the set returned from `cyclic()` for a Domain is, at least in some cases such as the example below (namely example field no. 2), not the same as the set for the corresponding `Field` i.e. `<a Field>.domain.cyclic()` and `<a Field>.cyclic()` are not equal in terms of size as well as named keys. However, they should be the same, at least with respect to the number of axes present regardless of keyed names, because axes of a Field's domain are, by contextual definition, the same as the axes of that Field.

#### Concrete example

```python
>>> f = cf.example_field(2)
>>> f.cyclic()
{'domainaxis2'}
>>> d = f.domain
>>> d.cyclic()
set()
```

sadielbartholomew:
Fix #762. The reason for the underlying issue was that the `Field.domain` attribute called `get_domain` which was inherited from `cfdm`, which doesn't have the concept of cyclic axes. So quite similar in terms of cause to #763 (maybe we should check there aren't further cases of cyclic axes not being accounted for, in case other bugs might exist?).

### DIFF
--- a/cf/field.py
+++ b/cf/field.py
@@ -2691,6 +2691,30 @@ class Field(mixin.FieldDomain, mixin.PropertiesData, cfdm.Field):
             for c in self.constructs.filter_by_data(todict=True).values():
                 c.cfa_update_file_substitutions(substitutions)
 
+    def get_domain(self):
+        """Return the domain.
+
+        .. versionadded:: NEXTVERSION
+
+        .. seealso:: `domain`
+
+        :Returns:
+
+            `Domain`
+                 The domain.
+
+        **Examples**
+
+        >>> d = f.get_domain()
+
+        """
+        domain = super().get_domain()
+
+        # Set axis cyclicity for the domain
+        domain._cyclic = self._cyclic
+
+        return domain
+
     def radius(self, default=None):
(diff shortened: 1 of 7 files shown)


cf/field.py
```diff
@@ -10455,9 +10479,9 @@ def convolution_filter(
                     new_bounds[0 : length - lower_offset, 1:] = old_bounds[
                         lower_offset:length, 1:
                     ]
-                    new_bounds[
-                        length - lower_offset : length, 1:
-                    ] = old_bounds[length - 1, 1:]
+                    new_bounds[length - lower_offset : length, 1:] = (
```
sadielbartholomew, line 10482: This was just from linting the files I touched in this PR. For the release we can re-lint the whole codebase since it seems we are a bit behind on linting (`black` edits >10 files under `cf/`).

sadielbartholomew:
Merging after offline acceptance from @davidhassell.

PR #778, opened 2024-05-30
Include the UM version as a field property when reading UM files

### LINKED ISSUE #777: Request UM version to be passed through pp conversion
Given the UM version is now often in the PP headers, it would be desirable for that to appear as an attribute on the fields produced during cf.read of this (ancient) data.

davidhassell:
Fixes #777 

Changes to:

- cf/weights.py
- cf/field.py
- cf/test/test_Data.py

are not relevant to this PR, but are just some linting that crept in from somewhere else.


cf/read_write/um/umread.py
```diff
@@ -1040,6 +1040,17 @@ def __init__(
             cf_properties["stash_code"] = str(stash)
             cf_properties["submodel"] = str(submodel)
 
+            # Convert the UM version to a string and provide it as a
+            # CF property. E.g. 405 -> '4.5', 606.3 -> '6.6.3', 1002
+            # -> '10.2'
+            a, b = divmod(int(self.um_version), 100)
```
sadielbartholomew, line 1046: I was wondering why, for the fraction component, you didn't just accept the raw numeric and then seeing this I realise it is to do with the floating point conversion:

```python
>>> divmod(606.3, 100)
(6.0, 6.2999999999999545)
```

Might be worth a one-line comment, since I was curious enough to look into that?

sadielbartholomew (approved):
One small code-related suggestion, but overall this is all good. Please merge when ready.

PR #780, opened 2024-06-04
Return a `FieldList` with `example_fields()`

### LINKED ISSUE #725: `example_fields()` should return a `Fieldlist` not a regular list
Unlike `cfdm` which has no concept of a `Fieldlist`, cf-python should return fields contained in a `FieldList` object rather than a regular Python list, which it returns at the moment:

```python
>>> type(cfdm.example_fields())  # expect a list
<class 'list'>
>>> type(cf.example_fields())  # expect a FieldList
<class 'list'>
```

sadielbartholomew:
Quick one to fix #725.

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -1,3 +1,15 @@
+
+version NEXTRELEASE
+-------------------
+
+**2024-??-??**
+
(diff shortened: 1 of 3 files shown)


davidhassell (approved):
Perfect - thanks.

PR #783, opened 2024-06-11
Correct finding coordrefs in `cf.aggregate`

davidhassell:
A bit tricky to test, as I can't see a case where the bug isn't benign: If two fields have different coordinate references, then they won't  get aggregated anyway, regardless of whether or not any coordinate references are attached to dimension coordinates (which is the "bug")

### DIFF
--- a/cf/aggregate.py
+++ b/cf/aggregate.py
@@ -511,7 +511,7 @@ class _Meta:
                         "cf_role": None,
                         "hasdata": dim_coord.has_data(),
                         "hasbounds": hasbounds,
-                        "coordrefs": self.find_coordrefs(axis),
+                        "coordrefs": self.find_coordrefs(dim_coord_key),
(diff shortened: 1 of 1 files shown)


sadielbartholomew:
Hi David, were you wanting a review on this - not sure since it's open not as a draft and seems ready but I've not been assigned, so thought I'd check?

sadielbartholomew (approved):
All good, though we might want to add a change log entry?

davidhassell:
Got any ideas of what it could say (I haven't :))

sadielbartholomew:
> Got any ideas of what it could say (I haven't :))

You make a good point, I do not :laughing: Therefore, leaving this out from the changelog only makes sense - as you imply, the bug this fixes could only manifest for users if they are doing something really quite strange, so is unlikely.

So, all good - please merge.

PR #812, opened 2024-09-03
New method: `cf.Field.filled`

### LINKED ISSUE #811: New method: `cf.Field.filled`
It would be useful to provide the `filled` method to `Field` objects - it is currently only defined on `cf.Data` (https://ncas-cms.github.io/cf-python/method/cf.Data.filled.html).

So, currently, to fill a field's missing values no in-place you have to do something like
```python
>>> g = f.copy()
>>> g.data.filled(0, inplace=True)
```
but it would be much nicer to be able to do, with the same result:
```python
>>> f.filled(0, inplace=True)
```

davidhassell:
Fixes #811

### DIFF
--- a/cf/mixin/propertiesdata.py
+++ b/cf/mixin/propertiesdata.py
@@ -3469,7 +3469,7 @@ class PropertiesData(Properties):
                 The fill value. By default the fill returned by
                 `fill_value` is used, or if this is not set then
                 the netCDF default fill value for the data type is
-                used (as defined by `netCDF.fillvals`).
+                used (as defined by `cf.default_netCDF_fillvals`).
 
             {{inplace: `bool`, optional}}
 
--- a/cf/mixin/propertiesdata.py
+++ b/cf/mixin/propertiesdata.py
@@ -3467,7 +3467,7 @@ class PropertiesData(Properties):
 
             fill_value: scalar, optional
                 The fill value. By default the fill returned by
-                `get_fill_value` is used, or if this is not set then
+                `fill_value` is used, or if this is not set then
                 the netCDF default fill value for the data type is
                 used (as defined by `netCDF.fillvals`).
 
--- a/cf/mixin/propertiesdata.py
+++ b/cf/mixin/propertiesdata.py
@@ -3458,6 +3458,36 @@ class PropertiesData(Properties):
         return set()
 
     @_inplace_enabled(default=False)
+    def filled(self, fill_value=None, inplace=False):
+        """Replace masked elements with a fill value.
+
+        .. versionadded:: NEXTVERSION
+
+        :Parameters:
+
+            fill_value: scalar, optional
+                The fill value. By default the fill returned by
+                `get_fill_value` is used, or if this is not set then
+                the netCDF default fill value for the data type is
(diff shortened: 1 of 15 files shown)


cf/mixin/propertiesdata.py
```diff
@@ -3457,6 +3457,36 @@ def file_locations(self):
...
+    def filled(self, fill_value=None, inplace=False):
+        """Replace masked elements with a fill value.
+
+        .. versionadded:: NEXTVERSION
+
+        :Parameters:
+
+            fill_value: scalar, optional
+                The fill value. By default the fill returned by
+                `get_fill_value` is used, or if this is not set then
```
sadielbartholomew, line 3470:
```suggestion
                `fill_value` is used, or if this is not set then
```

since this method isn't defined in the codebase so I guess from the below you mean this:

```pytb
>>> f.get_fill_value()
AttributeError: 'Field' object has no attribute 'get_fill_value'. Did you mean: 'fill_value'?
>>> print(f.fill_value())
None
```

cf/mixin/propertiesdatabounds.py
```diff
@@ -2079,6 +2079,41 @@ def file_locations(self):
...
+    def filled(self, fill_value=None, bounds=True, inplace=False):
+        """Replace masked elements with a fill value.
+
+        .. versionadded:: NEXTVERSION
+
+        :Parameters:
+
+            fill_value: scalar, optional
+                The fill value. By default the fill returned by
+                `get_fill_value` is used, or if this is not set then
```
sadielbartholomew, line 2092:
```suggestion
                `fill_value` is used, or if this is not set then
```

(As per above comment.)

cf/mixin/propertiesdatabounds.py
```diff
@@ -2079,6 +2079,41 @@ def file_locations(self):
...
+                the netCDF default fill value for the data type is
+                used (as defined by `netCDF.fillvals`).
```
sadielbartholomew, line 2094: As per the above comment:

```suggestion
                used (as defined by `cf.default_netCDF_fillvals`).
```

cf/mixin/propertiesdata.py
```diff
@@ -3457,6 +3457,36 @@ def file_locations(self):
...
+                the netCDF default fill value for the data type is
+                used (as defined by `netCDF.fillvals`).
```
sadielbartholomew, line 3472: Not sure what specifically this refers to (a `netCDF` class here or in cfdm?), from a quick poke around we could use either:

* our `cf.default_netCDF_fillvals()` method; or
* `netCDF4.default_fillvals` as the under-the-hood origin;
* 
either of which might be more appropriate - certainly they would be more searchable from the docs to the end user.

I suggest the first is best but up to you...

```suggestion
                used (as defined by `cf.default_netCDF_fillvals`).
```

sadielbartholomew (approved):
I've raised a pair of minor comments regarding the docstring parameter descriptions, but otherwise this is all good so please merge when you are ready. Thanks.

davidhassell:
Hi Sadie - good catches on those method names - they were throwbacks to some `cfdm` text that I blindly copied!

PR #820, opened 2024-10-09
Add new recipe (18) by summer student: correlation calculation

sadielbartholomew:
The final recipe created by Natalia, which again I have updated a bit to produce only one output plot, of best quality, as per the other recipes with a STDOUT report of the calculation output, plus to tidy and add further comment detail e.g. to describe the data sources. All things Natalia could have done herself had she had more time than the two weeks placement!

### Expected output

final-recipe-18


docs/source/recipes/plot_18_recipe.py
```diff
@@ -0,0 +1,140 @@
...
+# the snow cover extent, so snow cover is the dependent variable. The snow
+# cover data is the
+# 'Snow Cover Extent 2017-present (raster 500 m), Europe, daily – version 1'
+# sourced from the Copernicus Land Monitoring Service which is described at:
+# https://land.copernicus.eu/en/products/snow/snow-cover-extent-europe-v1-0-500m
+# and the elevation data is the 'NOAA NGDC GLOBE topo: elevation data' dataset
+# which can be sourced from the IRI Data Library, or details found, at:
+# http://iridl.ldeo.columbia.edu/SOURCES/.NOAA/.NGDC/.GLOBE/.topo/index.html.
+orog = cf.read("~/recipes/1km_elevation.nc")[0]
+snow = cf.read("~/recipes/snowcover")[0]
```
sadielbartholomew, line 33: Both of these are new datasets we need to use, so I have uploaded them to the 'recipes' Google Drive folder in readiness.

sadielbartholomew:
Given how busy Ankit is with his PhD we have agreed I can merge this without second review. At release time I will enact the plan I outlined above:

> let me suggest that I add all five recipes via PRs like this one, with the Python script and addition to the recipe_list.txt, then I do a final PR where I add all of the built material from those five to the repo finally, show they show up on the site

So, merging, and there will eventually be a follow-on PR to get all the new recipes built and included.

PR #821, opened 2024-10-10
Add new recipe (19) as started by summer student: per-season trends

sadielbartholomew:
A recipe I constructed based on the beginnings of a recipe produced by our second summer student, George. I decided it would be good to go on from his basis to show how simple it is for cf-python to calculate seasonal trends and to plot them with cf-plot, which is a bit different to anything shown in the existing recipes, as far as I have seen.

### Expected output

global_avg_sst_plot_donePR


docs/source/recipes/plot_19_recipe.py
```diff
@@ -0,0 +1,111 @@
...
+# %%
+# 1. Import cf-python, cf-plot and other required packages:
+import cfplot as cfp
+import matplotlib.pyplot as plt
+
+import cf
+
+# %%
+# 2. Read the dataset in extract the SST Field from the FieldList:
+f = cf.read("~/recipes_break/ERA5_monthly_averaged_SST.nc")
```
sadielbartholomew, line 21: Forgot to update this, let me see if I can update it via a suggestion to myself to avoid branch switching etc. :smile: :

```suggestion
f = cf.read("~/recipes/ERA5_monthly_averaged_SST.nc")
```

PR #822, opened 2024-10-11
Add new recipe (20) started by summer student: current divergence

sadielbartholomew:
Final recipe made or initiated from work by the summer students. This is the second one from a basis of work by George. I have updated ti to produce a plot of the vectors on top of the divergence.

### Expected output

irish-sea-currents-divergence-2006-01-15-23:30:00-finalPR


sadielbartholomew:
Given how busy Ankit is with his PhD we have agreed I can merge this without review. At release time I will enact the plan I outlined above:

> let me suggest that I add all five recipes via PRs like this one, with the Python script and addition to the recipe_list.txt, then I do a final PR where I add all of the built material from those five to the repo finally, show they show up on the site

So, merging, and there will eventually be a follow-on PR to get all the new recipes built and included.

PR #825, opened 2024-10-30
New function: `cf.cell_overlaps`

### LINKED ISSUE #824: New function: `cf.cell_overlaps`
The current `cf.cellwi(start, end)` only matches cells that lie _completely_ with the closed range `[start, end]`. 

Sometimes we would like to match cells that merely _overlap_ the range.

I propose a new function `cf.cell_overlaps(start, end)` that does just this. It will be a constructor function for  `<CF Query: [upper_bounds(ge start) & lower_bounds(le end)]>`

(Compare this with `cf.cellwi`, which is a constructor function for  `<CF Query: [lower_bounds(ge start) & upper_bounds(le end)]>`)

The docstrings for the various `Query` constructor functions could also be improved to more accurately reflect what they do!

davidhassell:
Fixes #824

### DIFF
--- a/cf/query.py
+++ b/cf/query.py
@@ -1581,7 +1581,7 @@ def eq(value, units=None, attr=None, exact=True):
 
 
 def ne(value, units=None, attr=None, exact=True):
-    """A condition for "not equalto a value".
+    """A condition for "not equal to a value".
 
     .. seealso:: `cf.contains`, `cf.eq`, `cf.ge`, `cf.gt`, `cf.le`,
                  `cf.lt`, `cf.set`, `cf.wi`, `cf.wo`, `cf.isclose`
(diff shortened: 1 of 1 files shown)


sadielbartholomew (commented):
Good idea for new methods, should be useful. Implementation and new testing and docs are otherwise all good but I have questioned the inconsistent underscore in the name and I have one general suggestion (plus spotted a few typos):

* Since this new `cell_overlaps`, as well as `cellwi` and `cellwo` which are `wi`/`wo` queries applied to a cell, it would be nice to include the new-ish `open_lower=False, open_upper=False` keywords we added recently to the general `wi`/`wo` queries so they can be made open or half-open.
* For the case of `cell_overlaps` the open-/closed-ness could do with being clarified a bit since it's a bit less direct/clear as to the role of interval openness. You state as with the other relevant methods 'The range is closed, meaning that is inclusive of its end points.' so I guess that means if the endpoint of the cell 'overlapped' with an endpoint of the range, then in the closed/inclusive default case it would be `True`, e.g. for cell `[5, 10]` and range `[10, 15]` it is deemed to overlap (True)? It would be good to add either a brief sentence or a new docstring example to clarify that.

Happy to approve once you'd considered those. Thanks.

PR #833, opened 2024-11-05
Allow 'nearest_dtos' 2-d regridding to work with discrete sampling geometry source grids

### LINKED ISSUE #832: Allow 'nearest_dtos' 2-d regridding to work with discrete sampling geometry source grids
It would be great to allow 'nearest_dtos' 2-d regridding to work with discrete sampling geometry source grids.

davidhassell:
Fixes #832

### DIFF
--- a/cf/data/dask_regrid.py
+++ b/cf/data/dask_regrid.py
@@ -507,7 +507,9 @@ def _regrid(
             # Note: It is much more efficient to access
             #       'weights.indptr', 'weights.indices', and
             #       'weights.data' directly, rather than iterating
-            #       over rows of 'weights' and using 'weights.getrow'.
+            #       over rows of 'weights' and using
+            #       'weights.getrow'. Also, 'np.count_nonzero' is much
+            #       faster than 'np.any' and 'np.all'.
             count_nonzero = np.count_nonzero
             indptr = weights.indptr.tolist()
             indices = weights.indices
@@ -547,7 +549,9 @@ def _regrid(
             # Note: It is much more efficient to access
             #       'weights.indptr', 'weights.indices', and
             #       'weights.data' directly, rather than iterating
-            #       over rows of 'weights' and using 'weights.getrow'.
+            #       over rows of 'weights' and using
+            #       'weights.getrow'. Also, 'np.count_nonzero' is much
+            #       faster than 'np.any' and 'np.all'.
             count_nonzero = np.count_nonzero
             where = np.where
             indptr = weights.indptr.tolist()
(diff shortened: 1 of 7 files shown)


cf/test/test_regrid_featureType.py
```diff
@@ -169,6 +163,78 @@ def test_Field_regrid_grid_to_featureType_3d(self):
...
+            ],
+        )
+
+        for src_masked in (False, True):
+            y = y0.copy()
+            if src_masked:
+                src = src.copy()
+                src[6:8] = cf.masked
+                # This following element should be smaller, because it
+                # now only has two source cells conrtibuting to it,
```
sadielbartholomew, line 206:
```suggestion
                # now only has two source cells contributing to it,
```

cf/data/dask_regrid.py
```diff
@@ -562,12 +567,45 @@ def _regrid(
 
             del indptr, pos_data
 
+        elif method == "nearest_dtos":
+            # 3) Nearest neighbour dtos method:
+            #
+            # Set to 0 any weight that corresponds to a masked source
```
sadielbartholomew, line 573: To check my understanding of this via the maths, am I correct in deducing that the ultimate dot product deals with the fact that, if we set to zero for masked source points (which is the only sensible option IMO so I agree is the way) then we are effectively using less points as sources, i.e. instead of shape (J, I) weights matrix as described here:

>         weights: `scipy.sparse.spmatrix`
>            The sparse weights matrix that defines the regridding
>            operation. Might be modified (not in-place) to account for
>            missing data in *a*. Must have shape ``(J, I)``, where
>            ``J`` is the number of destination grid cells and ``I`` is
>            the number of source grid cells. 

we now really have (J, I - M) where M is the number of masked source points? So I can convince myself the reduction in total source points is accounted for in the calculation. Reminding myself about dot products I think it will, but it would be good to get confirmation. Thanks.

sadielbartholomew (approved):
One question mostly for my understanding, but overall this is great and will be really useful for VISION work. Please merge once you've fixed one typo. Thanks.

cf/data/dask_regrid.py
davidhassell: Hi Sadie. Missing-data source points are effectively ignored by setting their corresponding weights to zero. Those zero-valued weights still need to appear in the `(J, I)` shaped matrix of weights, otherwise the dot product wont work, because the missing elements still exist in the source data (which by this stage has shape `(I,)`.

sadielbartholomew: OK I see, thanks for clarifying. All good!

davidhassell:
Great. Merging now. Thanks.

PR #837, opened 2024-12-17
Removed multiple warnings of unescaped backslash characters in the...

tylov:
..."manual"-output string, replacing them with \\\\
2. Added raw-text prefix 'r' to regular expression pattern (as it does not use formatting prefix 'f').


davidhassell (approved):
Hi @tylov - thanks for catching these! We are imminently going to be releasing 3.17.0 = perhaps this week. Before I merge the changes, would you like to add you name to https://ncas-cms.github.io/cf-python/contributing.html#contributors, which you can do at `cf/docs/source/contributing.rst`?

davidhassell:
Thanks, Tyge. Merging.

PR #858, opened 2025-03-31
Fix bug that caused `Data._axes` to be incorrect after a call to `cf.Field.collapse`

### LINKED ISSUE #857: `Data._axes` is not updated after a `cf.Field.collapse`
`Data._axes` is not updated after a `cf.Field.collapse`. If `squeeze=False` (the default) then this is not a problem, but if `squeeze=True` then the removed collapsed axes need stripping from returned `Data` object `_axes` attributes.

```python
>>> # Current
>>> import cf
>>> f = cf.example_field(0)
>>> g = f.collapse('X: mean', squeeze=True)
>>> g.data._axes
('dim0', 'dim1')  # WRONG
```

```python
>>> # Should be ...
>>> import cf
>>> f = cf.example_field(0)
>>> g = f.collapse('X: mean', squeeze=True)
>>> g.data._axes
('dim0',)  # RIGHT
```

davidhassell:
Fixes #857

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -13,6 +13,9 @@ version 3.17.0
   ``key`` (https://github.com/NCAS-CMS/cf-python/issues/802)
 * New keyword parameter to `cf.histogram`: ``density``
   (https://github.com/NCAS-CMS/cf-python/issues/794)
+* Fix bug that caused `Data._axes` to be incorrect after a call to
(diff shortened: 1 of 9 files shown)


sadielbartholomew (approved):
Perfect - isolates and fixes the bug. Please merge.

PR #860, opened 2025-04-01
 Fix bug that caused wrong directions from `cf.DimensionCoordinate.direction`

### LINKED ISSUE #859: Error in behaviour of `cf.DimensionCoordinate.direction`
The #856 it was [noted that](https://github.com/NCAS-CMS/cf-python/issues/856#issuecomment-2765985361) the collapsed time coordinates came out incorrectly as monotonically _decreasing_ when `cf.djf` was used in the MRE, whilst using `cf.mam` led to  the correct monotonically _increasing_ time coordinate sequence.

It turns this is because the time coordinates in the original field do not have bounds, and in the `cf.djf` grouped collapse case, one of the grouped collapses has just one month in it (December 2022), with bounds _that equal the coordinate_ value. In this case, `cfdm.DimensionCoordinate._infer_direction` tells us that the direction is decreasing due to the use of the strict inequalities `<`  (https://github.com/NCAS-CMS/cf-python/blob/v3.16.3/cf/dimensioncoordinate.py#L130), when non-strict ones (`<=`) are intended. The decreasing information feeds its way into the sorting of the collapsed groups via `reverse=coord.decreasing` (https://github.com/NCAS-CMS/cf-python/blob/v3.16.3/cf/field.py#L8747).

Very short PR to follow that fixes this by simply changing the appropriate `<` to `<=`.

davidhassell:
Fixes #859

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -16,6 +16,9 @@ version 3.17.0
 * Fix bug that caused `Data._axes` to be incorrect after a call to
   `cf.Field.collapse`
   (https://github.com/NCAS-CMS/cf-python/issues/857)
+* Fix bug that caused wrong directions from
+  `cf.DimensionCoordinate.direction`
+  (https://github.com/NCAS-CMS/cf-python/issues/859)
 * Changed dependency: ``Python>=3.9.0``
 * Changed dependency: ``numpy>=2.0.0``
 * Changed dependency: ``cfdm>=1.12.0.0, <1.12.1.0``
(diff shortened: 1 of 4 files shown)


sadielbartholomew (commented):
Do you have time to write a quick test for this within the next hour or so? I can forgive not having one given we are due to do a release ASAP, but ideally one can be added that would catch the original bug.

sadielbartholomew (approved):
Nice job working out what was going wrong. Thanks for the quick fix and adding a test. Perfect, please merge.

PR #861, opened 2025-04-01
Recipes tweaks for Sphinx Gallery build and rendering

sadielbartholomew:
Was intending to push these directly as commits, but since we've had some last-minute bug fixes I will use a PR to keep it self-contained. This PR is mostly tweaks to the numbered blocks in the new recipes, to get all `gopen()` and `gclose()` wrapped code into one block, else plots don't seem to emerge through the Sphinx Gallery build (in future I'll look into ways to fix this so we can keep the blocks separated out, which is clearer, but for the new release there is no time to investigate). Otherwise there is one update to a deprecated keyword and a new option for the `release_docs` script to facilitate recipe build debugging.

No review required as infrastructural and non-controversial and I am due to do the release ASAP.

### DIFF
--- a/docs/source/recipes/plot_04_recipe.py
+++ b/docs/source/recipes/plot_04_recipe.py
@@ -13,7 +13,7 @@ import cf
 # %%
 # 2. Read the field constructs:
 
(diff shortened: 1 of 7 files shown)


sadielbartholomew:
I will lint all as part of the release process, so ignoring the pre-commit job fail.

PR #862, opened 2025-04-04
Add "isort:skip_file" directives to recipes

davidhassell:
`pre-commit` wants to reorder the imports in some of the recipe files, but due to the esmpy/shapely class we don't want this to happen - so we can add some "isort:skip_file" directives to the relevant Python recipe files.

See https://pycqa.github.io/isort/ for details.


sadielbartholomew (approved):
Good spot, thanks for adding. Please merge.

PR #875, opened 2025-06-09
New `to_units` methods to allow changing units in a chain

### LINKED ISSUE #874: New `to_units` methods to allow changing units in a chain
@znichollscr suggested:

Currently, to convert the units you have to do `thing.Units = cf.Units("Tg")` and then you can carry on. Given how helpful chaining is, just adding a simple "to_unit" API would be great, so you could do `thing.to_unit("Tg").next_operation().follow_up()`.

PR to follow.

davidhassell:
Fixes #874

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -5653,11 +5653,19 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
     def override_units(self, units, inplace=False, i=False):
         """Override the data array units.
 
-        Not to be confused with setting the `Units` attribute to units
-        which are equivalent to the original units. This is different
-        because in this case the new units need not be equivalent to the
-        original ones and the data array elements will not be changed to
-        reflect the new units.
+        The new units need not be equivalent to the original ones, and
+        the data array elements will not be changed to reflect the new
+        units. Therefore, this method should only be used when it is
+        known that the data array values are correct but the units
+        have incorrectly encoded.
+
+        Not to be confused with changing to equivalent units with the
+        `to_units` method or the `Units`, `units`, or `calendar`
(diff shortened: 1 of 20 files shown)


cf/data/data.py
```diff
@@ -7365,6 +7380,57 @@ def to_memory(self):
             "Consider using 'Data.persist' instead."
         )
 
+    @_inplace_enabled(default=False)
+    def to_units(self, units, inplace=False):
+        """Change the data array units.
+
+        Changing the units will causes the data values to be changed
```
sadielbartholomew, line 7387:
```suggestion
        Changing the units causes the data values to be changed
```

cf/mixin/propertiesdatabounds.py
```diff
@@ -2229,16 +2238,19 @@ def override_calendar(self, calendar, inplace=False, i=False):
...
+        units. Therefore, this method should only be used when it is
+        known that the data array values are correct but the units
+        have incorrectly encoded.
 
-        Not to be confused with setting the `units` or `Units` attribute
-        to units which are equivalent to the original units.
+        Not to be confused with changing to equivalent units with the
+        `to_units` method or the `Units`, `units`, or `calendar`
+        attributes. These approaches also convert the data to have the
+        new units.
```
sadielbartholomew, line 2250: I suggest being a bit more explicit with this sentence, else the crux of it could easily be missed. Something like:

```suggestion
        attributes. These approaches also convert the data values to conform
        with the new units.
```

sadielbartholomew (approved):
One cut-and-pasted typo, one suggestion and one question, but otherwise all good - does as promised and is well tested. Please merge when ready.

PR #877, opened 2025-07-09
Docs: fix DOI badge at source & patch in built docs (to avoid rebuild)

sadielbartholomew:
On the docs homepage, through the introduction page source included also (via 'include' directive) in the index, some raw HTML was included in the RST for the latest version which caused some unrendered HTML text to show up directly under the heading 'Functionality':
badge_issue

I've updated the RST to include it in RST format so it should render to HTML when built but to avoid a full docs re-build I have patched the affected pages (index.html and introduction.html), with the fix which should now come from source as well.

FYI @davidhassell. Will merge without review since it is simple/self-contained, but I will check that it renders OK in the updated docs once the pages update via GitHub Pages and if not push a commit to ensure it is rendering and linking as it should.


sadielbartholomew:
Confirmed to work - docs pages already updated so that the badge now shows correctly.

Issue #890, opened 2025-08-15
arctan2 fails to account for units


ThatDesert:
When calling `cf.Data.arctan2(y, x)`, the function will ignore any units that `y` and `x` have and just treat them as unitless.

This means, for instance, that if `y` is in metres and `x` is in kilometres, no conversion will take place to divide the values of `y` by 1000 before calculating `da.arctan(y, x)`, resulting in incorrect values being returned due to incorrect unit scale in the same dimension.

If `x` and `y` have units, `cf.Data.arctan2()` should attempt to conform the units if they have the same dimension and raise an error if they do not.

If either `x`, `y`, or neither have units, then `cf.Data.arctan2()` should treat them both as unitless.

``` python
import cf

y = cf.Data([1], units=cf.Units('km'))
x = cf.Data([1], units=cf.Units('m'))

rad = cf.Data.arctan2(y, x)
print(rad.array)
```
``[0.7853981633974483]``

**vs.**

``` python
import cf

y = cf.Data([1000], units=cf.Units('m'))
x = cf.Data([1], units=cf.Units('m'))

rad = cf.Data.arctan2(y, x)
print(rad.array)
```
``[1.5697963271282298]``

sadielbartholomew:
Good spot, thanks Ollie.

PR #891, opened 2025-08-15
`cf.Data.arctan2()`: Fixes NCAS-CMS/cf-python#890 by trying to conform valid units first.

ThatDesert:
The function `cf.Data.arctan2` now first checks if `x2` has `cf`-style units (of the same type as the `cf.Units` class).

- If `x1` unitless and `x2` has `cf` units:
  - Attempts to conform `x1` to `x2`'s units, but `x1` remains unitless
  - `da.arctan2()` treats them both as unitless
- If `x1` has `cf` units and `x2` unitless:
  - Does not attempt to conform `x1` to `x2`'s units
  - `da.arctan2()` treats them both as unitless
- If `x1` and `x2` both unitless
  - Does not attempt to conform `x1` to `x2`'s units
  - `da.arctan2()` treats them both as unitless
- If 'x1' and 'x2' both have `cf` units
  - Attempts to conform `x1` to `x2`'s units
    - If `x1`'s units are of a different dimension than `x2`'s, an exception is raised
  - `da.arctan2()` operates on `x2` and the converted `x1`

### DIFF
--- a/cf/data/data.py
+++ b/cf/data/data.py
@@ -4407,7 +4407,9 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
             x2: array_like
                 X coordinates. *x1* and *x2* must be broadcastable to
                 a common shape (which becomes the shape of the
-                output).
+                output). If both *x1* and *x2* have units, they must
+                be in the same dimension (can be conformed), else
+                they will be treated as unitless.
 
         :Returns:
 
@@ -4434,6 +4436,10 @@ class Data(DataClassDeprecationsMixin, Container, cfdm.Data):
         [90.0 -90.0]
 
         """
+
+        if isinstance(getattr(x2, "Units", None), Units):
+            x1 = conform_units(x1, x2.Units)
+
         try:
             y = x1.to_dask_array()
         except AttributeError:


cf/data/data.py
```diff
@@ -4407,7 +4407,9 @@ def arctan2(cls, x1, x2):
             x2: array_like
                 X coordinates. *x1* and *x2* must be broadcastable to
                 a common shape (which becomes the shape of the
-                output).
+                output). If both *x1* and *x2* have units, they must
+                be in the same dimension (can be conformed), else
+                they will be treated as unitless.
```
sadielbartholomew, line 4412: A good use of dimensional as in 'dimensional analysis', but in this context (when considering multidimensional data arrays as is central to cf-python), 'dimension' already has a specific meaning, namely the [(number of) array axes](https://numpy.org/doc/stable/glossary.html#term-dimension), so best avoid using it in the first context.

We tend to use the word 'equivalent' to mean units that can be conformed i.e. converted by scaling values e.g. 'kg' and 'g' as opposed to 'kg' and 's' or 'apples' (see https://ncas-cms.github.io/cfunits/cfunits.Units.html if you want to learn more), so I suggest:

```suggestion
                be equal or equivalent, in which case they
                will be conformed.
```

(Also no need to cover what happens if not so I suggest dropping that end to the sentence.)

sadielbartholomew (commented):
Thanks for this Ollie, really good stuff - I've requested a pair of minor tweaks but they are mostly to make the changes of the PR consistent with (e.g. terminology) in our codebase and 'house style' so nothing which could have been foreseen to do.

Please either accept those suggestions or commit something similar to address the feedback and then I can merge this. As we discussed on your placement, a test would usually be added but since we ran out of time I can get the test added after this PR is merged (or guide you through how to add it if preferred).

PR #893, opened 2025-09-22
Allow regridding for very large grids

### LINKED ISSUE #878: Allow regridding for very large grids
Regridding to and/or from a very large grid (in terms of number of grid points) can fail due to running out of memory.

This occurs during the call to `esmpy.Regrid` which we use to calculate the regridding weights (https://github.com/NCAS-CMS/cf-python/blob/v3.18.0/cf/regrid/regrid.py#L2434). The problem isn't (so much) the size of the final weights matrix (although that can be quite large), more that `esmpy.Regrid` uses up a lot of memory for temporary arrays.

A real use case using 'linear' regridding has destination grid size 181440000, resulting in a sparse matrix representation of the weights taking up $$8.79\ GiB \equiv 4 \times 181440000 \times 3.25 \times 4 / 2^{30}\ GiB$$ (have a poke round inside the guts of a `scipy` CSR sparse array to see where that number comes from). The peak memory memory required by `esmpy` to create this is many times that! 

A work around for this is to calculate weights separately for non-overlapping partitions of the destination (not the source) grid, and then combining the weights from each of these partitions to create the final weights  to partition the final weights matrix.

We can add this as an option to the regrid API:

```python
>>> # Calculate the weights using 10 separate partitions
>>> f.regrids(g, 'linear', dst_grid_partitions=10)
>>> # Calculate the weights using the maximum possible number of partitions
>>> f.regrids(g, 'linear', dst_grid_partitions='maximum')
```

Why not do this always? Because it is often not required, and there can be a large speed performance hit for doing this. (Therefore it could be recommended that when using `dst_grid_partitions` you also save the resulting weights to a file for future use.)

----

To see how this works, consider a nearest neighbour regridding from a size 6 source grid to a size 5 destination grid. The `5 x 6` weights matrix for this could look like:

```
             I = 6
        + - - - - - - 
        | 1 0 0 0 0 0 
        | 0 0 0 0 1 0
  J = 5 | 0 0 1 0 0 0
        | 0 0 0 1 0 0
        | 0 0 0 0 0 1
```
This is clearly just a concatenation of the following `3 x 6` and `2 x 6` weights matrices:
```
             I = 6
        + - - - - - - 
        | 1 0 0 0 0 0 
 J0 = 3 | 0 0 0 0 1 0
        | 0 0 1 0 0 0
        + - - - - - - 
 J1 = 2 | 0 0 0 1 0 0
        | 0 0 0 0 0 1
```
The key is that two calls to `esmpy.Regrid` (one for the `J0` partition, one of the `J1` partition) create the correct rows of this matrix for any subset of the destination grid, provided:
1. the source grid is left un-partitioned, 
2. the regridded result at destination grid cell _j_ depends only on grid cell _j_, and not any other destination grid cells. If this were the case, as happens for `conservative_2d` and `nearest_dtos` methods, then some weights would be incorrect, as when they are calculated they "don't know" about destination grid cells in any of the other partitions.

davidhassell:
Fixes #878


cf/docstring/docstring.py
```diff
@@ -564,6 +569,43 @@
...
+            memory usage and minimising the calculation time. If the
+            string ``'maximum'`` is given then the largest possible
+            number of partitions of the destination grid will be used,
+            minimising memory usage and maximising the calculation
+            time. A positive integer specifies the exact number of
+            partitions, capped by the maximum allowed, allowing the
+            balance between memory usage and calculation time to be
+            adjusted.
+
+            The actual number of destination grid partitions; and each
```
sadielbartholomew:
```suggestion
            The actual number of destination grid partitions and each
```

unless the intention was to break the sentence up with further semi-colons?

cf/regrid/regrid.py
```diff
@@ -2400,8 +2564,30 @@ def create_esmpy_weights(
...
             col = None
 
+    if esmpy_regrid_operator is not None:
+        # If we're returning the esmpy regrid operator, we need to
+        # create it by computing the weights.
+        compute_weights = True
+
     from_file = True
-    if compute_weights or esmpy_regrid_operator is not None:
+    if compute_weights:  # or esmpy_regrid_operator is not None:
```
sadielbartholomew: Is the comment here meant as a note, or is this logic commented-out as left over from development or similar?

cf/regrid/regrid.py
```diff
@@ -3132,3 +3464,91 @@ def set_grid_type(grid):
...
+        or grid.name == "source"
+        or grid.dummy_size_2_dimension
+    ):
+        # One partition
+        if return_n:
+            return 1
+
+        return (None,)
+
+    from math import ceil
```
sadielbartholomew: Why not use `numpy.ceil`? SLB TODO

cf/regrid/regrid.py
```diff
@@ -3132,3 +3464,91 @@ def set_grid_type(grid):
...
     elif grid.is_locstream:
         grid.type = f"DSG {grid.featureType}"
+
+
+def partitions(grid, grid_partitions, return_n=False):
+    """Partitions of the grid.
+
+    Each partition is defined as an index to cell coordinates. Only
+    the last (left-most) dimension in esmpy order is partitioned. Note
+    that the coordinates storded in `grid.coords` are in esmpy order.
```
sadielbartholomew:
```suggestion
    that the coordinates stored in `grid.coords` are in esmpy order.
```

cf/test/test_functions.py
```diff
@@ -38,7 +38,6 @@ def test_keyword_deprecation(self):
 
     def test_aliases(self):
         self.assertEqual(cf.log_level(), cf.LOG_LEVEL())
-        self.assertEqual(cf.free_memory(), cf.FREE_MEMORY())
```
sadielbartholomew: I don't see why you got rid of this -  was it intentional?

sadielbartholomew (approved):
Very nice! And very well tested. I have only minor feedback, as raised in-line plus the below.

One general comment is that, ideally, we can clarify the maximum cap that gets applied - for example, a user might set `dst_grid_partitions=7` and it be capped to 5, say, then they could either be confused as to why 5 was used (if they happened to view with high verbosity i.e. debug mode), or (perhaps worse), be blissfully unaware the value applied was not that which they set. A particularly bad case would be when the maximum allowed is 1, such that setting any value including "maximum" would result still in 1 - which looks like something isn't working in the logic but really it is due to the cap. This isn't just a theoretical case, since I saw it when playing around interactively to explore the debug outputs, e.g:

```python
>>> src, dst = cf.example_fields(1, 0)
>>> z = src.regridc(dst, method='conservative', axes=['Y'], dst_grid_partitions="maximum", verbose=-1)
Source Grid:
Grid(name='source', coord_sys='Cartesian', type='structured grid', axis_keys=['domainaxis1'], axis_indices=[1], axes=['domainaxis1'], n_regrid_axes=1, dimensionality=1, shape=(10,), esmpy_shape=(10,), coords=[<CF DimensionCoordinate: grid_latitude(10) degrees>, array([0.])], bounds=[<CF Bounds: grid_latitude(10, 2) degrees>, array([[-1.,  1.]])], cyclic=False, method='conservative', dummy_size_2_dimension=False, is_grid=True, is_mesh=False, is_locstream=False, mesh_location=None, domain_topology=None, featureType=None, new_axis_keys=[], z=None, ln_z=False, z_index=None)

Destination Grid:
Grid(name='destination', coord_sys='Cartesian', type='structured grid', axis_keys=['domainaxis0'], axis_indices=[0], axes=['domainaxis0'], n_regrid_axes=1, dimensionality=1, shape=(5,), esmpy_shape=(5,), coords=[<CF DimensionCoordinate: latitude(5) degrees_north>, array([0.])], bounds=[<CF Bounds: latitude(5, 2) degrees_north>, array([[-1.,  1.]])], cyclic=False, method='conservative', dummy_size_2_dimension=False, is_grid=True, is_mesh=False, is_locstream=False, mesh_location=None, domain_topology=None, featureType=None, new_axis_keys=[], z=None, ln_z=False, z_index=None)

Calculating weights ...

Number of destination grid partitions: 1
Free memory before weights creation: 8.845745086669922 GiB
Maximum RSS before weights creation: 0.2023167908191681 GiB
```

where the `Number of destination grid partitions: 1` had me confused until I clocked why.

Does that make sense?

My preference in this regard would be to note the cap in the debug log (something like a line `Maximum cap on destination grid partitions: N` above the `Number of destination grid partitions: M` line) but more importantly raise a log 'warning' when the user sets a value which exceeds the cap, to inform them that it wasn't possible given the cap to use that value.

(Also you can set the `dst_grid_partitions` keyword to 0 or negative integers and they will be silently ignored such that the default  value 1 is applied - hopefully users will realise not to do that but might be good to raise a `ValueError` on nonsensical values.)

davidhassell:
There is a possible (probable!) issue with conservative 1-d Cartesian regridding - for this, a dummy size 1 2nd dimension is introduced, because ESMF likes that. However, that dummy dimension is the trailing poisition, so gets the partitioning treatment ... but it's size 1! 

(Note to self: for _linear_ 1-d Cartesian regridding this already  is handled correctly, because `grid.dummy_size_2_dimension=True`)

There is an easy fix for this, but not on the timescales of this release, so I'll open a new issue for it.

PR #904, opened 2025-10-29
Reference training materials (& remove reference to 'the the'!)

sadielbartholomew:
Quick one to add a note of the training material in the introduction/summary info for cf-python, in the various locations it appears: README, setup.py, docs intro (and the index page via inclusion of the introduction content there) etc. Whilst I was finalising the training materials for the ISC I realised it would be great to advertise those - though happy to tweak the sentence which links to them if you think.

Whilst preparing the PR I noticed some duplicated words in the docs - and upon further investigation the snippet (and band 😆) 'the the' in particular appears very frequently (I wonder how that happened?) - so I've blitzed those via some `find` and `sed` commands.

Includes commiting of re-built HTML for any of the pages featuring the typo (excluding the API ones where we cab wait for the typo fix until next release's docs build) - otherwise only tweaks sources.

### DIFF
--- a/docs/analysis.html
+++ b/docs/analysis.html
@@ -6,13 +6,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" /><meta name="viewport" content="width=device-width, initial-scale=1" />
 
     <title>Analysis &#8212; Documentation</title>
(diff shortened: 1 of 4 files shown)


davidhassell (approved):
Great - so many repeated words :)

PR #905, opened 2025-11-01
Reduce the time taken to import `cf`

### LINKED ISSUE #902: Reduce the time taken to import `cf`
Importing `cf` is quite slow ... on my local computer it takes 2 seconds.

There seem to be two main culprits for the slow import:
- Doc string rewriting
  - At import time, every docstring (of which there are currently 5569) is inspected for doc string substitutions, replacing any that are found
- Importing external modules that themselves have a slow import
  - The main problems here are `dask`, `scipy`, `s3fs`, `zarr`, `h5netcdf`, `uritools`, `netCDF4`

These can be improved on:
- Doc string rewriting
  - Handled in `cfdm` : https://github.com/NCAS-CMS/cfdm/issues/361
- Importing external modules that themselves have a slow import
  - Move the imports to run time, rather than import time. Also import a lot of this from `cfdm`. Many will not ever get imported, and when they do, the time is usually negligible compared to the operation being run.

#### Results

By applying these changes, my local import time reduces to 0.25 seconds (from 2 seconds - a factor of 6 speed-up). 

PR to follow

davidhassell:
Fixes #902

Should be considered after https://github.com/NCAS-CMS/cfdm/pull/362 has been dealt with

### DIFF
--- a/docs/source/recipes/plot_08_recipe.py
+++ b/docs/source/recipes/plot_08_recipe.py
@@ -9,10 +9,11 @@ In this recipe, we will analyse and plot temperature trends from the HadCRUT.5.0
 # 1. Import cf-python, cf-plot, numpy and scipy.stats:
 
 import cfplot as cfp
+import cf
+
 import numpy as np
 import scipy.stats as stats
 
-import cf
 
 # %%
 # 2. Three functions are defined:
--- a/docs/source/recipes/plot_12_recipe.py
+++ b/docs/source/recipes/plot_12_recipe.py
@@ -13,8 +13,8 @@ retrievals.
 # %%
 # 1. Import cf-python, cf-plot and matplotlib.pyplot:
 
-import cfplot as cfp
 import matplotlib.pyplot as plt
+import cfplot as cfp
 
 import cf
 
--- a/docs/source/recipes/plot_13_recipe.py
+++ b/docs/source/recipes/plot_13_recipe.py
@@ -18,10 +18,12 @@ period of six months or more.
 # in next steps.
 
 import cartopy.crs as ccrs
-import cfplot as cfp
 import matplotlib.patches as mpatches
 
+import cfplot as cfp
+
 import cf
+
 
 # %%
 # 2. Read and select the SST by index and look at its contents:
--- a/docs/source/recipes/plot_17_recipe.py
+++ b/docs/source/recipes/plot_17_recipe.py
@@ -11,8 +11,8 @@ misrepresentation, or lack of accessibility, a careful choice must be made.
 # %%
 # 1. Import cf-python and cf-plot:
 
-import cfplot as cfp
 import matplotlib.pyplot as plt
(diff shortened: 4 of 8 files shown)


cf/functions.py
```diff
@@ -3208,6 +3225,8 @@ def environment(display=True, paths=True):
     cf: 3.18.0
 
     """
+    #    regridding =  _get_module_info("esmpy", alternative_name="ESMF", try_except=True)
+
```
sadielbartholomew: Left over during dev?

```suggestion
```

sadielbartholomew (approved):
Hmm, the code changes all look fine and sensible for improving import time to some extent, but as for timings on my local system, I am not seeing much speed-up on this branch relative to `main` in terms of import time, at least judging it by the useful test command you [shared in one of the original issues](https://github.com/NCAS-CMS/cfdm/issues/361#issuecomment-3442929502).

There is a bit, which is of course good, but nothing like the factor 6 you saw when you tested as per your comment on the Issue. For example, on first run in a fresh terminal tab, with the `cfdm` branch kept on [the one just merged with its own import timing improvements](https://github.com/NCAS-CMS/cfdm/pull/362):

```console
$ git checkout main # after
...
import time:      1297 |     606844 | cf
$ git checkout david/import-speed-up # after
...
import time:      1160 |     602408 | cf
```

so remains ~0.6 s, only 1% faster.

I wonder if you are quoting the factor 6 improvement using timings with this branch in combination with the cfdm sister one NCAS-CMS/cfdm#362, relative to the cfdm `main` (before that was merged) - i.e. the timing improvement is mostly down to  the cfdm-side changes? I do notice considerable speed-up on the cf-python side with the cfdm branch set to that relative to `main` before: 0.6 s relative to 1.3 s. But we can't attribute that to this PR.

So, unless I'm missing something this cf-side PR doesn't seem to produce that significant of a reduction in import time, but it does produce a small one which is good enough to be merge-worthy - so I am approving (but would be interesting to know more context to your own timings). Please merge after checking the small number of minor feedback items raised, plus this more general comment:
*  I would prefer it if we don't change the import order in the recipes, since those were carefully placed to avoid the possibility for seg faulting due to bad matplotlib and esmpy lower-level code interactions (see https://github.com/NCAS-CMS/cf-plot/issues/24) - however I will review that issue in terms of whether it can be avoided given the state of dependencies on our and those libraries now. I imagine the changes to anything in `docs/source/recipes` was due to linting and notably `isort`. Would you mind reverting those?

(Again it is a shame we can't use 'PEP 810 – Explicit lazy imports' now to avoid the movement of imports as per my notes on the cfdm PR. But should be good for the next several years until Python 3.15 is the standard minimum. 😆)

davidhassell:
On my laptop, I get 1.3 seconds for the original, 0.75 seconds with improved cfdm, and 0.25 seconds with imprved cfdm _and_ improved cf-python

PR #914, opened 2026-01-07
Control setting and display of cached data elements

### LINKED ISSUE #913: Control setting and display of cached data elements
When the array being passed to `cf.Data.__init__` is not a Dask array and is in memory, why don't we grab the first (, second) and last values for the cache, as it's much quicker to do it at this time than hauling them out of the instantiated Dask array at a later stage.

At the same time, we can provide a function `cf.display_data` that allows the user to not display first (, second) and last values during `dump`, `str` or `repr`, unless they're in the cache.

This means that you can still inspect the data even when computing the display values takes forever and a day.

Most of the heavy lifting for this will be provided by `cfdm` via https://github.com/NCAS-CMS/cfdm/issues/363.

PR to follow.

davidhassell:
Fixes #913

### DIFF
--- a/cf/read_write/um/umread.py
+++ b/cf/read_write/um/umread.py
@@ -3575,7 +3575,7 @@ class UMRead(cfdm.read_write.IORead):
                 # Return now if there are valid file types
                 return []
 
-        f = self.file_open(filename, parse=True)
+        f = self.dataset_open(filename, parse=True)
 
         info = is_log_level_info(logger)
 
@@ -3598,7 +3598,7 @@ class UMRead(cfdm.read_write.IORead):
             for var in f.vars
         ]
 
-        self.file_close()
+        self.dataset_close()
 
         return [field for x in um for field in x.fields if field]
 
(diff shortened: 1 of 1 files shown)


Changelog.rst
```diff
@@ -1,10 +1,22 @@
...
+* New function to control the creation of cached elements during data
+  display: `cf.display_data`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* New methods: `cf.Data.get_cached_elements`
+  `cf.Data.cache_elements`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* Set cached elements during `cf.Data.__init__`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* Removed the `cf.constants.CONSTANTS` dictionary, replacing it
+  `cf.ConstantAccess.constants`
```
sadielbartholomew:
```suggestion
  with `cf.ConstantAccess.constants`
```

Issue #895, opened 2025-09-24
Write Zarr datasets


davidhassell:
It would be very nice to be able to write [Zarr](https://zarr-specs.readthedocs.io) datasets with `cf.write`

Only Zarr v3 will be supported.

The writing of group hierarchies will be supported.

See also #863 (Read Zarr datasets) and #894 (Read grouped Zarr datasets), and https://github.com/NCAS-CMS/cfdm/issues/354

PR #915, opened 2026-01-08
Zarr write + Zarr groups

### LINKED ISSUE #894: Read grouped Zarr datasets
It would be very nice to be able to read, with `cf.read`, [Zarr](https://zarr-specs.readthedocs.io) datasets that contain a group hierarchy.

#863 implemented the reading of Zarr datasets, but only those with no group structure (i.e. only the root group).

Zarr v2 (xarray-style) and Zarr v3 will be supported.

This will almost entirely implemented in `cfdm` (https://github.com/NCAS-CMS/cfdm/issues/355), and inherited from there.

davidhassell:
Fixes #894 and #895

### DIFF
--- a/Changelog.rst
+++ b/Changelog.rst
@@ -10,6 +10,20 @@ Version NEXTVERSION
 * Reduce the time taken to import `cf`
   (https://github.com/NCAS-CMS/cf-python/issues/902)
 * New optional dependency: ``zarr>=3.1.3``
+* New function to control the creation of cached elements during data
+  display: `cf.display_data`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* New methods: `cf.Data.get_cached_elements`,
+  `cf.Data.cache_elements`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* Set cached elements during `cf.Data.__init__`
+  (https://github.com/NCAS-CMS/cf-python/issues/913)
+* Removed the `cf.constants.CONSTANTS` dictionary, replacing it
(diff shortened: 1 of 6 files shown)


Changelog.rst
```diff
@@ -1,10 +1,16 @@
 Version NEXTVERSION
--------------------
+--------------
 
-**2025-12-??**
+**2026-01-??**
 
+* Write Zarr v3 datasets with `cf.write`, and allow the reading of
+  grouped Zarr v2 and v3 datasets with `cf.read`
```
sadielbartholomew:
```suggestion
* Write Zarr v3 datasets with `cf.write`
```

since the second part is duplicating the information in the bullet below (and I think it is nice to have them as separate bullets given the separate linked Issues).

docs/source/tutorial.rst
```diff
@@ -5345,8 +5348,8 @@ By default the output file will be for CF-|version|.
    
 The `cf.write` function has optional parameters to
 
-* set the output netCDF format (all netCDF3 and netCDF4 formats are
-  possible);
+* set the output netCDF format (all netCDF3 and netCDF4 formats, as
```
sadielbartholomew:
```suggestion
* set the output format (all netCDF3 and netCDF4 formats, as
```

seems better since Zarr isn't really netCDF.

davidhassell:
Thanks for the git command. I did `git checkout 9328936f04628183b8f6eda1a06006c50e5a54e9 -- docs/source/recipes/
` which fixed 7 of the dodgy changes. I need to have a look at the others.

PR #918, opened 2026-01-15
Update hooks & workflows esp. to skip recipes with `isort`

sadielbartholomew:
Some updates to workflows and hooks:
* fix a (fairly) recent one-line change which broke the workflow `run-test-suite.yml` due to Actions strictness over valid YAML formatting;
* update all pre-commit hooks, since they were quite behind the latest ones by now (we haven't shown the hooks and workflows much love lately);
* configure to skip running `isort` on the docs source directory holding the recipes, since we want to keep the specific off-rule import order we apply there.

Feel free to review @davidhassell though as it is pure infrastructure it might not need it. As for whether the workflows will run and pass now - I am not sure since the GitHub Status page is saying Actions is presently degraded. But I will check when it reports as being back to normal that everything works as hoped, else put in a follow-on commit(s) until it does.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
   # Use specific format-enforcing pre-commit hooks from the core library
   # with the default configuration (see pre-commit.com for documentation)
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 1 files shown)


davidhassell:
Happy to not review this :)

PR #919, opened 2026-01-16
Decouple recipes gallery from main docs build

### LINKED ISSUE #916: Separate out recipes build from main docs build
The recipes in the docs are really useful and important, however:
* they are - naturally given they work with large (often multiple GB in size) datasets - slow to build and prone to errors including ones such as lower-level seg faults which are opaque without resorting to going in with a debugger, and ones relating to issues with the Sphinx build;
* building them every time we build the main docs is unnecessary since we only need to re-build with new recipes or API changes (though we want to test them all as working each release, that doesn't require building them into the docs again, we can just run the scripts and eventually do image comparison on the plots).

For this reason, it would make life, especially release time, much easier if we can build the recipes separately to the build of the main cf-python docs. We foresee at least two good ways to make the separation. Overall, the ideal solution would allow us to build the recipes separately but have them still seamlessly appearing the main rendered docs site as they are now - which can be done via organising both sets of built HTML and links from one to the other.

At present we are tending towards preferring Option 2 below, since keeping them in the repo keeps the whole of cf-python and its current docs contained in one place with no need to move them and manage another repo. But each has pros and cons.

### Options

1. Move recipes to a new separate repo, with its own documentation infrastructure (in Sphinx, largely identical to the cf-python repo one in config. and theme etc.).
2. Move recipes to a separate directory within the main cf-python repo with its own docs build, i.e. so a separate `make html` command applies.
3. Another option? Happy to hear other suggestions.

sadielbartholomew:
Close #916. We chose to keep the recipes in this repo, but have them in a separate directory to the rest of the docs (the 'main docs' henceforward) with a separate make build where configuration is inherited from the `conf.py` from the main docs, with a little extra logic to set up `sphinx-gallery` and point to the appropriate static resources.

Merging now since this is required for the upcoming release for which I have commits locally ready to push shortly.

Note the `release_docs` bash script can likely be tidied a bit further, but in the interests of not delaying the release I will leave it in the present form which is decent enough.

### DIFF
--- a/.github/workflows/run-test-suite.yml
+++ b/.github/workflows/run-test-suite.yml
@@ -146,8 +146,11 @@ jobs:
       # Also, it means code coverage is only generated if the test suite is
       # passing at least for that job, avoiding useless coverage reports.
       uses: codecov/codecov-action@v3
+      # Warning: GitHub Actions expressions require single-quoted strings.
+      # Double quotes will break workflow validation (even though YAML allows
+      # them). So don't change the quoting below.
       if: |
-        matrix.os == "ubuntu-latest" && matrix.python-version == "3.10"
+        matrix.os == 'ubuntu-latest' && matrix.python-version == '3.10'
       with:
         file: |
           ${{ github.workspace }}/main/cf/test/cf_coverage_reports/coverage.xml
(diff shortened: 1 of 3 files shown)


sadielbartholomew:
Linting failure is due to picking up the old recipes directory location - now updated. Test job failures are due to using the new release of cfdm without updates I have locally - so I will get those committed next. Merging.

PR #940, opened 2026-03-26
Add all cf-plot dependencies (i.e. Cartopy) to `cf.environment`

### LINKED ISSUE #908: cf-plot (& dependencies) inclusion in `cf.environment()`
cf-python has an [`environment` function](https://ncas-cms.github.io/cf-python/function/cf.environment.html) which reports the versions, and optionally paths to, itself and its dependencies, especially useful for user support where we ask users to share the output they see from that so we can quickly learn about the exact versions they are using for anything relevant. ~It would be good to have a cf-plot equivalent so that we can use it in cf-plot itself (users may not be using cf-python explicitly, so have to `import cf` to access it in code importing only cf-plot) and notably get information about extra cf-plot dependencies such as Cartopy etc.~

~I propose:~
~1. We create a cf-plot equivalent, `cfp.environment()`, which wraps cf-python to report on anything cf-python needs, since cf-python is a cf-plot dependency, but adds any further dependencies cf-plot has to the list. This separates concerns nicely.
2. `cf.environment` actually reports the version of cf-plot and some of its dependencies (notably matplotlib) despite these not being relevant to cf-python itself, so I also propose we move anything relating to cf-plot such as cf-plot and matplotlib out of that and into `cfp.environment()`.~

[Proposal changed - see next comment bold text. Ultimately, we want to provide a way to get crucial version and dependency information for both cf-python separately and cf-python plus cf-plot.]

sadielbartholomew:
Close #908. For the benefit of user support etc. it would be ideal to have [all cf-plot dependencies](https://github.com/NCAS-CMS/cf-plot/blob/26d90b65ca0c4b37c373fa8c2d499ac792e35f0e/setup.py#L32-L37) reported by `cf.environment` and `cartopy` is the only one that is not already contained there.(though if any get added as new cf-plot deps I will need to ensure they also get included here, in future.

Notes:
* We no longer include example outputs of `environment` calls so it seems only the underlying function code needed updating.
* Test passes still but I thought that was a good reason to update the test to be more thorough, so that is also included.
* I don't think it is worth updating the changelog to note that one new llibrary appears in the output of `environment` (too minor), but let me know if you think otherwise.

### DIFF
--- a/cf/test/test_functions.py
+++ b/cf/test/test_functions.py
@@ -288,24 +288,79 @@ class functionTest(unittest.TestCase):
         e = cf.environment(display=False)
         ep = cf.environment(display=False, paths=False)
 
+        # Basic structure
         self.assertIsInstance(e, list)
         self.assertIsInstance(ep, list)
+        self.assertEqual(len(e), len(ep))
+        self.assertTrue(all(isinstance(s, str) for s in e))
+        self.assertTrue(all(isinstance(s, str) for s in ep))
 
-        components = ["Platform: ", "netCDF4: ", "numpy: ", "cftime: "]
-        for component in components:
-            self.assertTrue(any(s.startswith(component) for s in e))
-            self.assertTrue(any(s.startswith(component) for s in ep))
-        for component in [
-            f"cf: {cf.__version__} {os.path.abspath(cf.__file__)}",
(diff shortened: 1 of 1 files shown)


davidhassell (approved):
Perfect - thanks! Merge at will.

PR #950, opened 2026-05-29
New keyword parameter to `cf.Field.regrids` and `cf.Field.regridc`:   max_masked

### LINKED ISSUE #949: Enable linear regridding to ignore masked source cells and still produce unmasked destination cells
To date (`cf<=v3.20.0`), linear regridding  (just linear, not other methods) has always produced a masked destination grid cell if _any_ of the contributing source grid cells are masked. This approach is not always desirable, as it can produce too-large mismatches between the source grid mask and the mask of the regridded data on the destination grid.

We should have an option to specify how many masked source grid cells can be ignored for each destination grid cell, allowing a numerical value to be calculated. When this occurs, the weights of the remaining non-masked source cells need to be adjusted so that they still sum to `1.0`.

I propose a non-negative integer-valued `max_masked` keyword parameter to `cf.Field.regrids` and `cf.Field.regridc`. This defines the maximum number of masked source grid cells which will be ignored for each destination grid cell. If the number of masked source grid cells exceeds this number then the destination grid cell will always be masked.

The default value is `0`, which provides the current behaviour - i.e. any masked source cell leads to a masked destination cell.  

```python
>>> new = f.regrids(g, method='linear', max_masked=0)   # Equivalent to f.regrids(g, method='linear')
>>> new = f.regrids(g, method='linear', max_masked=2)
```

davidhassell:
Fixes #949 

All the excitement happens in `cf/data/dask_regrid.py`. The rest is just passing the new `max_masked` parameter about, docs and tests :)


cf/docstring/docstring.py
```diff
@@ -394,14 +394,37 @@
...
                 Destination grid cell ``j`` will only be masked if a)
                 it is masked in the destination grid definition; or b)
                 the sum of ``w_ji`` for all non-masked source grid
                 cells ``i`` is strictly less than *min_weight*.""",
+    # max_masked
+    "{{max_masked: `int`, optional}}": """max_masked: `int`, optional
+                For linear regridding only. Ignored for all other
+                regridding methods.
+
+                The maximum allow number of masked source cells which
```
sadielbartholomew:
```suggestion
                The maximum number of masked source cells which
```

cf/regrid/regrid.py
```diff
@@ -261,6 +262,57 @@ def regrid(
...
+            strictly less than *min_weight*.
+
+        max_masked: `int`, optional
+            For linear regridding only. Ignored for all other
+            regridding methods.
+
+            The maximum allow number of masked source cells which are
+            allowed to be ignored when calculating a non-masked
+            destination cell. When masked source cells are ignored,
+            the weights w_ij of non-masked source cells i are adjusted
```
sadielbartholomew:
```suggestion
            the weights w_ji of non-masked source cells i are adjusted
```

just to be consistent, otherwise mathsy types might read into the index swap (unless it was indeed intended?).

cf/data/dask_regrid.py
```diff
@@ -556,11 +596,42 @@ def _regrid(
...
+                w = data[i0:i1]
+                if where(w[mask] >= min_weight)[0].size > max_masked:
+                    # There are more masked src cells than allowed
                     dst_mask[j] = True
+                else:
+                    # The number of masked src cells does not exceed
+                    # the minimum masked-cells threshold
+                    non_masked_indices = where((~mask) & (w >= min_weight))[0]
+                    if non_masked_indices.size == 1:
+                        # There areis exactly one non-masked src cell
```
sadielbartholomew:
```suggestion
                        # There is exactly one non-masked src cell
```

cf/data/dask_regrid.py
```diff
@@ -421,6 +449,16 @@ def _regrid(
 
             See `regrid` for details.
 
+        max_masked: `int`, optional
+            For linear regridding only. Ignored for all other
+            regridding methods. The maximum allow number of masked
```
sadielbartholomew:
```suggestion
            regridding methods. The maximum number of masked
```

cf/data/dask_regrid.py
```diff
@@ -13,164 +13,185 @@ def regrid(
...
+                Destination grid cell j will only be masked if a) it is
+                masked in the destination grid definition; or b) the sum
+                of ``w_ji`` for all non-masked source grid cells i is
+                strictly less than *min_weight*.
+
+            max_masked: `int`, optional
+                For linear regridding only. Ignored for all other
+                regridding methods.
+
+                The maximum allow number of masked source cells which are
```
sadielbartholomew:
```suggestion
                The maximum number of masked source cells which are
```

cf/data/dask_regrid.py
```diff
@@ -13,164 +13,185 @@ def regrid(
...
+                The final value of Vd_j is the sum of w_ji * Vs_i for all
+                source grid cells i. Note that it is typical that for a
+                given j most w_ji will be zero, reflecting the fact only a
+                few source grid cells intersect a particular destination
+                grid cell. I.e. *weights* is usually a very sparse matrix.
+
+                If the destination grid has masked cells, either because
+                it spans areas outside of the source grid, or by selection
+                (such as ocean cells for land-only data), then the
+                corresponding rows in the weights matrix must be be
```
sadielbartholomew:
```suggestion
                corresponding rows in the weights matrix must be
```

(Was from the original docstring, not introduced in this PR, but I noticed it so might as well fix here.)

sadielbartholomew (approved):
Aside from some typos it is perfect, so please review those and then merge at will.

That said, I wanted to comment regarding API consistency - though it relates more to the issue at hand and your proposed way to handle it (i.e. #949) rather than the implementation of it here. Namely I notice we have the keyword  'mtol' for `Field.collapse` and the various convenience methods for collapses (mean, integral etc.) and that aligns very closely with the new `max_masked` keyword introduced, [given its description](https://ncas-cms.github.io/cf-python/method/cf.Field.collapse.html#cf.Field.collapse):

> Set the fraction of input data elements which is allowed to contain missing data when contributing to an individual output data element. Where this fraction exceeds mtol, missing data is returned. The default is 1, meaning that a missing datum in the output array occurs when its contributing input array elements are all missing data. A value of 0 means that a missing datum in the output array occurs whenever any of its contributing input array elements are missing data. Any intermediate value is permitted.

They have different names but (by my understanding) effectively control the same thing (input masking to output masking) - bar the slight complication of the role of any weighting used for the regrid case? But one is specified by number,  the other by fraction.

So in future we could consider choosing one over the other for both collapses and regridding (and any other applicable methods).

davidhassell:
Hi Sadie - thanks for the review.

I'd like to think about the pertinent API question you raise before merging. `mtol` specified as a fraction in `[0,1]` has the advantage that for linear regridding from unstructured source grids, there may be different numbers of potentially contributing cells for each destination grid cell. So being able to specify a fraction (`0.75`, say) means you can play nicely with triangles, squares, pentagons, etc.

This would involve  small tweak to the code (to ascertain the number of source grid cells in play, and to convert the fraction to a number of masked points in each case), but that's no more or less fiddly than what we've already got! 

I'll take a look ....

davidhassell:
Hi Sadie - PR updated for `mtol` ... 698b18d218241e318b34ee2e391dd6e99362d056

cf/data/dask_regrid.py
```diff
@@ -13,164 +13,200 @@ def regrid(
...
+                to 1.
+
+                Define ``w_ji`` as the multiplicative weight that
+                defines how much of ``Vs_i`` (the value in source grid
+                cell ``i``) contributes to ``Vd_j`` (the value in
+                destination grid cell ``j``).
+
+                A destination grid cell j will be masked if *mtol*
+                multiplied by the total number of source cells i for
+                which ``w_ji >= min_weight`` is greater then the
```
sadielbartholomew:
```suggestion
                which ``w_ji >= min_weight`` is greater than the
```

cf/data/dask_regrid.py
```diff
@@ -13,164 +13,200 @@ def regrid(
...
+                By default *mtol* is ``0``, meaning that destination
+                grid cell j will be masked if any source cell i for
+                which ``w_ji >= min_weight`` is masked.
+
+                For instance, for a rectilinear source grid for which
+                up to 4 source grid cells contribute to each
+                destination grid cell, if *mtol* is in the range
+                ``[0.5, 0.75)`` then a destination grid cell will in
+                general only be be masked if three or more of its
+                source grid cells are masked.
```
sadielbartholomew, line 202: +1, very useful to have an example as its a bit difficult to envision the concept in practice otherwise.

cf/docstring/docstring.py
```diff
@@ -393,15 +393,48 @@
...
+                to 1.
+
+                Define ``w_ji`` as the multiplicative weight that
+                defines how much of ``Vs_i`` (the value in source grid
+                cell ``i``) contributes to ``Vd_j`` (the value in
+                destination grid cell ``j``).
+
+                A destination grid cell j will be masked if *mtol*
+                multiplied by the total number of source cells i for
+                which ``w_ji >= min_weight`` is greater then the
```
sadielbartholomew:
```suggestion
                which ``w_ji >= min_weight`` is greater than the
```

cf/docstring/docstring.py
```diff
@@ -393,15 +393,48 @@
...
+
+                By default *mtol* is ``0``, meaning that destination
+                grid cell j will be masked if any source cell i for
+                which ``w_ji >= min_weight`` is masked.
+
+                For instance, for a rectilinear source grid for which
+                up to 4 source grid cells contribute to each
+                destination grid cell, if *mtol* is in the range
+                ``[0.5, 0.75)`` then a destination grid cell will in
+                general only be be masked if three or more of its
```
sadielbartholomew:
```suggestion
                general only be masked if three or more of its
```

cf/data/dask_regrid.py
```diff
@@ -13,164 +13,200 @@ def regrid(
...
+
+                By default *mtol* is ``0``, meaning that destination
+                grid cell j will be masked if any source cell i for
+                which ``w_ji >= min_weight`` is masked.
+
+                For instance, for a rectilinear source grid for which
+                up to 4 source grid cells contribute to each
+                destination grid cell, if *mtol* is in the range
+                ``[0.5, 0.75)`` then a destination grid cell will in
+                general only be be masked if three or more of its
```
sadielbartholomew:
```suggestion
                general only be masked if three or more of its
```

cf/regrid/regrid.py
```diff
@@ -261,6 +262,70 @@ def regrid(
...
+            adjusted so that they sum to 1.
+
+            Define ``w_ji`` as the multiplicative weight that defines
+            how much of ``Vs_i`` (the value in source grid cell ``i``)
+            contributes to ``Vd_j`` (the value in destination grid
+            cell ``j``).
+
+            A destination grid cell j will be masked if *mtol*
+            multiplied by the total number of source cells i for which
+            ``w_ji >= min_weight`` is greater then the number of those
```
sadielbartholomew:
```suggestion
            ``w_ji >= min_weight`` is greater than the number of those
```

cf/regrid/regrid.py
```diff
@@ -261,6 +262,70 @@ def regrid(
...
+            source grid cells which are masked.
+
+            By default *mtol* is ``0``, meaning that destination grid
+            cell j will be masked if any source cell i for which
+            ``w_ji >= min_weight`` is masked.
+
+            For instance, for a rectilinear source grid for which up
+            to 4 source grid cells contribute to each destination grid
+            cell, if *mtol* is in the range ``[0.5, 0.75)`` then a
+            destination grid cell will in general only be be masked if
```
sadielbartholomew:
```suggestion
            destination grid cell will in general only be masked if
```

sadielbartholomew (approved):
Thanks David for updating the approach - sorry it was extra work but like you indicated, it provides a more flexible approach than the `max_masked` original kwarg idea and it means we have a consistent kwarg 'mtol' across multiple supported methods. So, much nicer overall.

I reviewed the new commit and then did a sanity check on the whole PR branch. Overall is all good. Some more typos to correct (just two really, copied and pasted across) and a changelog conflict to resolve, but other than that all ready to merge so please go ahead.

davidhassell:
Thanks, Sadie!

